PR #5, opened 2023-07-10
feature (15441) - Allow JWT reissue in CORS

simon622:
Add the reissuance header to the CORS filter. see https://hivemq.kanbanize.com/ctrl_board/57/cards/15441/details/


simon622:
Formality for tracking purposes (actually pushed fix to master)

PR #17, opened 2023-07-19
feat(15459): Adding first-use username/password from configuration endpoint

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/15459/details/

The PR adds the `/api/v1/frontend/configuration` endpoint prior to the login form in order to receive, if defined, a pre-populated username and password. 

A welcome message is also added to the UI when contained in the endpoint's payload.

The post-form message ("Contact your system administrator for access credentials") is not displayed if the welcome message is given

### Before
screenshot-localhost_3000-2023 07 19-10_41_29

### After
screenshot-localhost_3000-2023 07 19-10_41_06

### Out-of-scope
Note that there is still a use case to decide on, whenever the user changes the prefilled values and triggers a validation error. Both welcome and error messages are displayed at the same time

screenshot-localhost_3000-2023 07 19-10_45_10

### DIFF
--- a/.github/workflows/frontend-cli.yml
+++ b/.github/workflows/frontend-cli.yml
@@ -5,11 +5,12 @@ name: Frontend - React Testing Pyramid
 
 on:
   push:
-    branches: [master, alpha]
+    branches:
+      - "**"        # matches every branch
     paths:
       - hivemq-edge/src/frontend/**
   pull_request:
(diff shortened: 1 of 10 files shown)


simon622 (approved):
Confimed first use details prefilled and messaging in place.

PR #29, opened 2023-08-10
feat(16123): Ensure image URLs are served from API server

simon622:
Whilst working in developer mode, it important that images render when running Frontend on a different developer server. For this reason, when specified, the developer flag will enable logoUrls to be fully qualified.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/api/utils/ApiUtils.java
+++ b/hivemq-edge/src/main/java/com/hivemq/api/utils/ApiUtils.java
@@ -15,8 +15,14 @@
  */
 package com.hivemq.api.utils;
 
+import com.google.common.base.Preconditions;
+import com.hivemq.api.config.ApiListener;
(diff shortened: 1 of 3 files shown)


hivemq-edge/src/main/java/com/hivemq/api/utils/ApiUtils.java
```diff
@@ -31,4 +38,41 @@ public static boolean hasDefaultUser(List<UsernamePasswordRoles> users){
...
+        List<ApiListener> listeners = apiConfigurationService.getListeners();
+        if(listeners == null || listeners.isEmpty()){
+            return null;
+        }
+
+        String protocol = null;
+        int port = 80;
+        String host = null;
+
+        for(ApiListener listener : listeners){
```
sfrehse, line 53: Reads very implicit to break at the end. :-)

sfrehse (approved):

PR #30, opened 2023-08-10
feat(16109): Add PING event to usage data. Include session tracking.

simon622:
Add a PING event to the tracking system. Fix version flaw. Add session token to tracking data and encapsultate the population of this data.


sfrehse (commented):
Please put some hints directly in the doc what we are actually sending.

PR #32, opened 2023-08-15
feat (15617) - Ensure modules are loaded from workspace in development mode

simon622:
It is currently difficult to develop modules in a full runtime, due to the nature of the module loader only considering the built artefacts. Allow the build/ contributes workspace directories to be considered ahead of the lib contributed .jar files.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/edge/modules/ModuleLoader.java
+++ b/hivemq-edge/src/main/java/com/hivemq/edge/modules/ModuleLoader.java
@@ -16,6 +16,7 @@
 package com.hivemq.edge.modules;
 
 import com.hivemq.configuration.info.SystemInformation;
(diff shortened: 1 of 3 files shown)


hivemq-edge/src/main/java/com/hivemq/edge/modules/ModuleLoader.java
```diff
@@ -86,6 +103,7 @@ private void discoverWorkspaceModule(final File dir, final ClassLoader parentCla
     }
 
     private void loadFromModulesDirectory(final ClassLoader parentClassloader){
+        log.debug("Loading modules from HiveMQ Home.");
```
sfrehse, line 106: Is it worth to print the directory?

sfrehse (approved):

PR #36, opened 2023-08-18
fix (16220) - Backed off (terminal) error jobs are not removed from pool

simon622:
Added a callback hook, so that jobs who directly manage the connections through a sync API can callback and close underlying instances aligning it to the backoff job pool terminal event code path. For example when Job A exceeds its threshold and Job B shares the connection, Job A will be removed from the pool and will close the underlying resource that JobB shares - Job B will then need to be safe against bad connection handles. NOTE: The failure of Job A should NOT remove Job B by design, since Job A may have failed for a reason unrelated to the underlying connection resource, and thus Job B may still have a valid handle to work with (this becomes adapter dependant).

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/params/impl/ProtocolAdapterPollingInputImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/params/impl/ProtocolAdapterPollingInputImpl.java
@@ -58,4 +58,9 @@ public abstract class ProtocolAdapterPollingInputImpl implements ProtocolAdapter
     public int getMaxErrorsBeforeRemoval() {
         return maxErrorsBeforeRemoval;
     }
+
+    @Override
+    public void close() {
+
+    }
 }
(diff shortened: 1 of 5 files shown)


hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/params/impl/ProtocolAdapterPollingInputImpl.java
```diff
@@ -58,4 +62,14 @@ public TimeUnit getUnit() {
     public int getMaxErrorsBeforeRemoval() {
         return maxErrorsBeforeRemoval;
     }
+
+    @Override
+    public void close() {
```
sfrehse, line 67: Is the function being called? I couldn't find it.

simon622: Yup - it's called by the monitoring job on shutdown. FWIW - I don't like the pattern of the runnable being the Input object - we should refactor this at some point. The small issue is the interfaces need to be appropriate for both async and sync patterns - since we don't know what the underlying connection API will support.

`if(future.cancel(false)){
                    pollingJob.getInput().close();
                }`

PR #40, opened 2023-08-21
fix(16308): Allows Edge frontend to initiate backend requests when offline 

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/16308/details/

This PR fixes the problem of API requests not being fired when their is no connectivity.

The main problem was with the configuration of `ReactQuery`, see documentation (https://tanstack.com/query/v4/docs/react/guides/network-mode). By default, the `Network Mode` is set to `online`, in which Queries and Mutations will not fire unless you have network connection, setting the query into a `paused` state.

The PR switches the `Network Mode` to `always`, in which Queries and Mutations are always sent and ReactQuery basically ignore online/offline status.  

The PR also introduces a `timeout` limit to all `Axios` requests (by default set to 0, implying a permanent loading).

### DIFF
--- a/hivemq-edge/src/frontend/src/api/hooks/useHttpClient/useHttpClient.ts
+++ b/hivemq-edge/src/frontend/src/api/hooks/useHttpClient/useHttpClient.ts
@@ -29,6 +29,12 @@ export const useHttpClient = () => {
   function createInstance() {
     // Make sure to clear the interceptors, since axiosInstance is global
     axiosInstance.interceptors.response.clear()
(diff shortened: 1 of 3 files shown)


simon622 (approved):
I think we should expose this configuration somewhere within our own constant file so users don't have to search for it.

vanch3d:
> I think we should expose this configuration somewhere within our own constant file so users don't have to search for it.

Done for the current bug but we will need to design a strategy for the config, to devide
- what goes to a central configuration file
- where is that configuration done

PR #43, opened 2023-08-22
feat(16258): Show which protocol the user is creating/editing and adapter for

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/16258/details/

The PR adds the name and the icon of the protocol being edited (for both creation or modification), in the header of the side panel.

It allows the user to immediatly see which protocol the form is associated with.

### Before
screenshot-localhost_3000-2023 08 22-11_41_42

### After
screenshot-localhost_3000-2023 08 22-11_41_29


simon622:
The icons are still small - but we can look to standardise these with some custom sizes moving forward.

simon622 (approved):
The icons are still small - but we can look to standardise these with some custom sizes moving forward.

PR #44, opened 2023-08-22
feat(16263): Add type icon to adapter instance list

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/16263/details/

The PR changes the content of the `type` column in the list of adapters, replacing the internal type by the proper name and icon of the protocol.

The PR also fixes accessibility of the protocol icon (adding a `alt` property)


simon622 (approved):
Looks good with full adapter string and icon

PR #47, opened 2023-08-22
fix (16316) - Set and agent value in user properties

simon622:
As a HiveMQ, we want to identify edge installations connected to our brokers, either enterprise installed on-premise or managed cloud. 

==

Example from development env:
client-agent: “HiveMQ-Edge; Development Snapshot”


sfrehse (approved):
❤️

PR #48, opened 2023-08-22
fix(16318): ID attribute is disabled if adapter modified rather than created

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/16318/details/

The PR add, through the `UISchema`, a `disabled` state to the `id` property if the adapter is modified rather than created

Notes:
- The `UISchema` is based on property names, like `id`. There it is expected that the unique identifier (or name) of an adapter is always attributed to a `id` property 
- The fix propsed in the PR needs to be tested for nested properties that might be named `id` 

### Before
screenshot-localhost_3000-2023 08 22-16_36_35

### After
screenshot-localhost_3000-2023 08 22-16_36_20


simon622 (approved):
Create, Edit & Delete all work nicely as does existing validation

PR #70, opened 2023-09-01
fix(16511) - Ensure pings are NOT sent during a non-tracked runtime

simon622:
Pings are sent directly by the HttpService. This means they were not considering the runtime configuration which was functionally considered in the layer above. We now pass an indication as to whether to start the usage tracking service through the API via contructor - which determines if the tracking engine is started.


sfrehse (approved):
LGTM

PR #71, opened 2023-09-04
fix(16486): Fix a11y contrast violation for connection status

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/16486/details/

THis PR changes the `colorScheme` definition for the custom status, ensuring compliance with WCAG contrast rules

The colors for the different states remain the same but are properly defined on a scale, ensuring combined application of font and background colors. 

The variant of the status component has been changed to make it lighter.

Aslo, the variant of the version badge (in adapters) was changed to differentiate from the status

### After


sfrehse (commented):
I understand the changes but I don't get why it solves the problem :-)

vanch3d:
> I understand the changes but I don't get why it solves the problem :-)

My original theme for the status used a single colour for the type, e.g. 
```js
connected: {
        500: '#38A169',
      },
```

But internally, `ChakraUI` uses a range of values, with increasing hue from`50` to `900`, e.g. 
```js
    blue: {
      50: '#ffffff',
      100: '#ccd8e2',
      200: '#99b2c5',
      300: '#668ba8',
      400: '#33658b',
      500: '#003e6e',
      600: '#003258',
      700: '#002542',
      800: '#00192c',
      900: '#000c16',
    },
```

and then, it will combine text colour (e.g. `black.700`) with background colour (e.g. `blue.200`) to get a proper contrast. 
With my single colour, it couldn't adapt. 
The change defines the colours of the status with a whole range and I just reuse the default schemes from the base then: 
```js
      connected: baseTheme.colors.green,
      disconnected: baseTheme.colors.gray,
      connecting: baseTheme.colors.orange,
      disconnecting: baseTheme.colors.orange,
      error: baseTheme.colors.red,
```

PR #81, opened 2023-09-06
fix(16608): add optimistic update of status to the creation of adapters

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/166608/details/

An `ERROR` could be seen on the status of a new adapter, because its updated from the pooled Status request which, immediately after sending the create request, doesn't have a record for the new adapter. It gets fixed with the next call of Status

The PR utilises `ReactQuery` optimistic update to add a temporary `CONNECTING` item in the cache for the status. 

The PR also fix the ley for the request, which will also force an update of the `GET adapters` when creating a new one


sfrehse (approved):
LGTM a very good UX improvement

PR #89, opened 2023-09-08
feat(16640) - Add gatekeeping to adapter discovery and allow adapters to determine if they wish to add support

simon622:
Current discovery is an advanced adapter feature, which only some adapters support. In anticipation of UI and API calling these, we need to provide an indication as to wwhether the underlying type supports it. e.g. HTTP will never support discovery.

New method added to the type metadata, new abstraction of discoverValues in place for default implementations and strong gatekeeping around the API method. (Tests in IT).


hivemq-edge/src/main/java/com/hivemq/api/model/adapters/ProtocolAdapter.java
```diff
@@ -66,6 +66,10 @@ public class ProtocolAdapter {
     @Schema(description = "Is the adapter installed?")
     private final @NotNull Boolean installed;
 
+    @JsonProperty("supportsDiscovery")
```
sfrehse, line 69: Do we want to provide a list of capabilities of an adapter, e.g., writeable, readable, discoverable.

This would lead to an array of string rather than a boolean which would be things more extendable.

simon622: Very good point. Let's use capability constants. I will add.

sfrehse (approved):
Awesome :-) LGTM

PR #98, opened 2023-09-11
Fix(94): Connection Status are now updated on the workspace

vanch3d:
Fixes https://github.com/hivemq/hivemq-edge/issues/94

The PR corrects one missed feature from the workspace, the ability to regularly check for and update the connection status of bridges and adapters.

Both API requests are handled together, updating the list of nodes when necessary. 

Note that, by nature of the existing API endpoints, we need to send 2 requests for every update.

### Out of scope
- The ability to start and stop adapters or bridges will be done in a subsequent ticket

### DIFF
--- a/hivemq-edge/src/frontend/src/modules/EdgeVisualisation/utils/status-utils.spec.ts
+++ b/hivemq-edge/src/frontend/src/modules/EdgeVisualisation/utils/status-utils.spec.ts
@@ -1,0 +1,103 @@
+import { expect } from 'vitest'
+import { Node, NodeProps } from 'reactflow'
+
+import { Adapter, ConnectionStatus } from '@/api/__generated__'
+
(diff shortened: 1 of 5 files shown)


hivemq-edge/src/frontend/src/modules/EdgeVisualisation/utils/status-utils.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+          bridgeRuntimeInformation: expect.objectContaining({
+            connectionStatus: expect.objectContaining({ status: ConnectionStatus.status.CONNECTING }),
+          }),
+        }),
+      }),
+    ]),
+  },
+]
+
+describe('updateNodeStatus', () => {
```
sfrehse, line 97: nice one

sfrehse (approved):
LGTM

Issue #107, opened 2023-09-14
Auto-detections of Adapter's topics in the frontend


vanch3d:
For usage in the workspace an other components, we need to be able to extract the topics defined in every adapter, regardless of their internal structure

While a generic mapping could be delivered as part of the consolidation of the backend's adapter definition (see ?????), this is a job that in fact can be easily implemented in the frontend

The current definition of all known adapters is using a custom `format` attribute : 

```
/// opc-ua
'mqtt-topic': {
  type: 'string',
  title: 'Destination MQTT topic',
  description: 'The MQTT topic to publish to',
  format: 'mqtt-topic',
},
/// modbus
destination: {
  type: 'string',
  title: 'Destination Topic',
  description: 'The topic to publish data on',
  format: 'mqtt-topic',
},
```

PR #110, opened 2023-09-14
feat(107): Extraction of topics from generic adapters

vanch3d:
Fixes https://github.com/hivemq/hivemq-edge/issues/107

This PR implements routines for identifying and extracting topics from any adapters, regardless of its internals (i.e. `JSONSchema`).

The process is based on the fact that, for all the known adapters, topics are clearly identified by a custom `format` for the property. 

This is true regardless of the name and location of the `JSONSchema` property holding this information, 

The routine is based on three simple steps:
- flatten the `JSONSchema` into a list of path-defined properties
- clean the paths to isolate `mqtt-topic` and multiple items 
- recursively traverse the form data associated with the active adapter to isolate the value of the `mqtt-topic` properties

When applied to the Workspace configuration, it eliminates the need for hardcoded types or external knowledge of the adapters, since it is in fact internally exposed. 

### Before
screenshot-localhost_3000-2023 09 14-22_10_59

### After
screenshot-localhost_3000-2023 09 14-22_10_03

### DIFF
--- a/hivemq-edge/src/frontend/src/__test-utils__/adapters/http.ts
+++ b/hivemq-edge/src/frontend/src/__test-utils__/adapters/http.ts
@@ -1,0 +1,152 @@
+import { ProtocolAdapter } from '@/api/__generated__'
+
+export const MOCK_PROTOCOL_HTTP: ProtocolAdapter = {
+  id: 'http',
+  protocol: 'HTTP(s) over TCP',
+  name: 'HTTP(s) to MQTT Protocol Adapter',
+  description:
+    'Connects HiveMQ Edge to arbitrary web endpoint URLs via HTTP(s), consuming structured JSON or plain data.',
+  url: 'https://github.com/hivemq/hivemq-edge/wiki/Protocol-adapters#http',
+  version: 'Development Snapshot (BETA)',
+  logoUrl: 'http://localhost:8080/images/http-icon.png',
+  author: 'HiveMQ',
+  installed: true,
+  // capabilities: ['READ'],
+  category: {
+    name: 'CONNECTIVITY',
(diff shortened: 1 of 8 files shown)


hivemq-edge/src/frontend/src/__test-utils__/adapters/http.ts
```diff
@@ -0,0 +1,152 @@
+import { ProtocolAdapter } from '@/api/__generated__'
```
sfrehse, line 1: Are these files auto-generated?

hivemq-edge/src/frontend/src/modules/EdgeVisualisation/utils/topics-utils.ts
```diff
@@ -42,3 +45,67 @@ export const getBridgeTopics = (bridge: Bridge): { local: TopicFilter[]; remote:
...
+  }
+  return result
+}
+
+export const getTopicPaths = (configSchema: RJSFSchema) => {
+  const flattenSchema = flattenObject(configSchema)
+  return (
+    Object.entries(flattenSchema)
+      // Only interested in topics, internally defined by the string format `format: 'mqtt-topic'`
+      .filter(([k, v]) => k.endsWith('format') && v === 'mqtt-topic')
```
sfrehse, line 67: Does it make sense to extract `mqtt-topic` as a constant?

hivemq-edge/src/frontend/src/__test-utils__/adapters/http.ts
vanch3d: Yes, using a manual converter. They are not needed anymore. Removing them

PR #116, opened 2023-09-18
feat(108) - Define 2 status fields for adapters & bridges

simon622:
We need to be able to express 2 status' for adapters and bridges. We now have runtimeStatus and connectionStatus.


simon622:
Merged - note the IT need merging in the test project. //TODO simon

Issue #113, opened 2023-09-15
Improve UX flow when network connection errors


vanch3d:
Incidentally, the frontend, thanks to `React Query` caching, is behaving pretty good when network is down but there are way too many operations that should not be available and way not enough feedback to the user notifying about the  downtime

_Originally posted by @vanch3d in https://github.com/hivemq/hivemq-edge/issues/95#issuecomment-1721178110_

PR #117, opened 2023-09-20
feat(113): add error and loading state to the main pages

vanch3d:
Fixes https://github.com/hivemq/hivemq-edge/issues/113

This PR adds the missing handlers for loading and error states to the main pages of the application, when data are being fetched 

While data is being loaded, feedback is being provided either by a loading spinner (e.g. login age) or by "skeleton" when visualising data. 

Errors are indicated by inline alerts replacing the expected content.

Error and loading state are alos preventing further interactions to the page (e.g. error while loaing the list of adapters prevent editing or creating a new adapter). 

Note that all the data requests are cached by `React-Query`. Covering loading and error in the main routing pages ensures that the expected data have been properly fetched by the time hooks are used further down the DOM tree. 

### Namespace


simon622 (approved):
Looks good - I did note that the workspace view did not display any error messages. All other views had a nice skeleton and subsequent error message.

Issue #124, opened 2023-09-20
Start & Stop Bridges


simon622:
We should allow bridges to be independently started and stopped via API and UI. A new action on the overflow menu to START | STOP | RESTART the connection.

PR #135, opened 2023-09-27
feat(123): Start, stop or restart bridges and adapters

### LINKED ISSUE #123: Start & Stop Adapters
Now we have introduced isolated states for runtime and connection status, it gives us far more control to be able to start / stop individual instances. Depending on the implementation of the adapter, some adapters will NOT try to reestablish a connection on an invalid set of details after a number of backoff retries (they will be transitioned from connection ERROR back to DISCONNECTED - but still running) where they will stay until they are edited. We should allow more control on these, but allowing a START | STOP | RESTART on each instance in the control actions.

We may also wish to consider the fundamental control/state cycle of each adapter. Suggest we detail some rheudamentary state transition diagrams for this work unit.

vanch3d:
Fixes #123 and #124

The PR refactors the handling of the connection chamges for both bridges and adapters. 

- For the bridges, it adds a "connection controller" on each card with three buttons (start or stop, restart), start or stop being displayed depending on the current status of the bridge. 

- For the adapters, it add the same "connection controller" with three menu items, supporting the same login as for the bridges.

- The "connection controller" is also available in the workspace (for adapters only, bridges will be updated later). 

The  "connection controller" handled the operation of connection change by sending the request and delegating any update of the rendering to the `GET status` rtequest. It however exploit a "minimum time delay" (returned by successful requests) to ensure that none of the status change CTAs are needlessly enabled for further change.

The PR also fixes the rendering of the status badge, to reflect the dual composition of `runtime` and `connection` status of a device: `OFF` when the device has been turned off, `CONNECTED`, `DISCONNECTED`, `STATELESS` or `ERROR` otherwise.

### Bridges 
screenshot-localhost_3000-2023 09 27-16_40_36

### Adapters
screenshot-localhost_3000-2023 09 27-16_42_24

screenshot-localhost_3000-2023 09 27-16_39_51

### Workspace
  
screenshot-localhost_3000-2023 09 27-16_41_23


simon622 (commented):
Looks good Nicolas - couple of points;

1. When you're in a stopped (Off) state - it's possible to Start OR Restart an entity - really only START is valid when in the stopped state.
2. It wasn't 100% clear to me that the icon used for Start was indeed a start button. We use the traditional Stop and Restart icon, and then a non-traditional Start icon (I would have expected a play button - triangle rotated 90 degrees clockwise). Perhaps we can all input into this one, as iconography is always subjective.

simon622:
All looks good and works well!

simon622 (approved):
Look great, thanks!

Issue #138, opened 2023-09-29
Add custom topic editing to the adapters


vanch3d:
**User Story**

As a HiveMQ Edge user, I want to recognise that I'm editing a topic for an adapter and how that topic might relate to previous configuration

**Value** 

- Enable users with a consistent approach to visualising and editing topics, segments and topic tree

**Definition of Done**
This ticket is part of an initiative consolidating the visualisation, selection and editing of topics across the Edge. See (#138, #139 and #140 )
 
- The current input component is refactored to be used across Bridge and Adapters
- The adapter form is generated using a custom `UISchema` associated with the format `mqtt-topic`, as extracted from their `JSONSchema` configuration
- The new input component is an open text with auto-completion 
- Options of the auto-completion select component are provided by all the topics defined through the app's topic tree
- Rendering of the options should reflect the design choices of the topic tree selectors
- Rendering of the selected topics should reflect the design choices of both topics and topic tree

![Image](https://github.com/hivemq/hivemq-edge/assets/2743481/aba3e030-ffe8-4cd7-bd8b-55099c85bc49)

### To do

- [x] `JSONSchemaForm`: create a custom `BaseInputTemplate` to deal with `mqtt-format`
- [ ] `CreatableSelect`: create a custom auto-completion selector for the topics
   - support single and multiple items 
   - custom  rendering for options
- [x] `ReactQuery`: refactor the extraction of topics from the data sources (bridges and adapters)
   - support publish-only filtering
- [ ] `JSONSchemaForm`: create a custom validation rule to validate topics (e.g. rejecting wildcards, ...)

PR #146, opened 2023-10-04
feat(138): Custom editing field for topics in the adapter form

vanch3d:
See #138

This PR is the first step in an initiative to make topic editing consistent across Hive Edge frontend. It replaces the standard open-text field by a more complex auto-complete select input widget. 

The widget is associated with the `mqtt-topic` format of the `JSONSchema` content of the adapter and will apply to any property associated with this string format, regardless of their name.

### Desgin

The widget itself is a reusable auto-complete select component that offers the following features:

- the left-hand side of the field is showing the common icon we use across the app for `topics`
- the field (the `mqtt-topic` format in fact) is associated with a custom validation, preventing the use of restricted characters for publishing subscription (all adapters are publish-only) 
- the input iself is supported by a list of matching topics that can be used to immediatly complete the typing. Typing any string in the input will only show the options matching the text (any substring)
- If the typed string is not in the list of existing options, the widget prompts the user to add the new item (to both the list of options and the input)
- the right-hand side of the widget includes a `clear` icon (to clear the typed text from the input) and a `dropdown` icon (to open the list of options) 
- the list of options is created by all the topics used across the other (publishing) entities and comes under two flavours: the topics themselves (e.g. `leaves`, for direct reuse) or the segment stub to the topics (e.g. `branches`, for creating sa new topic out of an existing structure)

Notes: 
- At this moment, there is no definitive scenario for the choice of one option flavour or the other (leaves or branches), and no easy design to offer  that choice to the user. The current implementation uses `leaves` 

### Out-of-scope
- Pressing `ESC` while focused on the widget doesn't close the form, as every other fields in the side panel does. This behaviour needs to be made consistent.

- Once created (by typing or selecting an option), the text in the input cannot be modified. Editing it again will start from scratch. This is not the expected behaviour; the user should be able to select an existing topic (leaf or branch) and conitnue editing to create a new topic. Overriding the behaviour will be the target of the dedicated ticket.

- The auto-completion of a topic, using a `segment` by segment approach cannot be achieve without the previous issue. Once fixed, creating a dynamic option list to show strings for the next segment only, will be the target of a dedicated ticket 

### Before
screenshot-localhost_3000-2023 10 04-10_38_10

### After
screenshot-localhost_3000-2023 10 04-10_41_07

screenshot-localhost_3000-2023 10 04-10_43_22 (1)

screenshot-localhost_3000-2023 10 04-10_48_31


simon622 (changes requested):
Bug discovered during first use:

1. Start API server
2. Login to Refreshed webapp
3. Immediately click the Workspace link
4. App crashes with following screen
Screenshot 2023-10-10 at 16 58 08

simon622 (approved):
Confirmed bug no longer present and the topic field works as expect. nice.

PR #151, opened 2023-10-06
fix(17040) - Add local and remote subscription varients and add excludes to local

simon622:
The excludes field residing on the BridgeSubscription entity is misleading in the UI, it is mandating that the excludes is a required field (where it is not).

We should move excludes field to its own extending type


simon622 (commented):
All looks good to the changes you made NVL

Issue #139, opened 2023-09-29
Add custom topic editing to the bridge


vanch3d:
**User Story**

As a HiveMQ Edge user, I want to recognise that I'm editing a topic for the bridge and how that topic might relate to previous configuration

**Value** 

- Enable users with a consistent approach to visualising and editing topic, segments and topic tree

**Definition of Done**
This ticket is part of an initiative consolidating the visualisation, selection and editing of topics across the Edge. See (#138, #139 and #140)
 
- The current input component is refactored to be used across Bridge and Adapters
- The new input component is an open text with auto-completion 
- Options of the auto-completion select component are provided by all the topics defined through the app's topic tree
- Rendering of the options should reflect the design choices of the topic tree selectors
- Rendering of the selected topics should reflect the design choices of both topics and topic tree

![Image](https://github.com/hivemq/hivemq-edge/assets/2743481/28d5f114-4f2e-4757-98d0-a910e4f16b14)

PR #152, opened 2023-10-09
feat(139): Custom editing field for topics in the Bridge form

vanch3d:
See #139 

This PR is the second step in an initiative to make topic editing consistent across Hive Edge frontend. It replaces the standard auto-complete field by a more complex select input widget, similar to the work done for adapters. 

The PR also fixes the rendering of the topic, especially for non-segmented and long string

### Design 

The widget is an extension of the single topic editor used for the adapters, this time supporting multiple topics

- the input iself is supported by a list of matching topics that can be used to immediatly complete the typing. Typing any string in the input will only show the options matching the text (any substring)
- If the typed string is not in the list of existing options, the widget prompts the user to add the new item (to both the list of options and the input)
- The list of options is collected from all the topics used in configuration of both adapters and bridges, with special wildcards accepted
- the right-hand side of the widget includes a clear icon (to clear the typed text from the input) 
- when editing is completed (`enter`, `tab` or clicking on the `add the topic` option), a new item is added to the field. The item has the `topic` icon and a clear button amd visualised exactly as in the `workspace`, as a "badge".

### Out-of-scope
- Selected topics cannot be changed once added, except by removing them completely from the field and recreating them 

- As with the adapter's signle topic, editing a topci always start from scratch. This is not the expected behaviour; the user should be able to select an existing topic (leaf or branch) and conitnue editing to create a new topic. Overriding the behaviour will be the target of the dedicated ticket.

- The auto-completion of a topic, using a segment by segment approach cannot be achieve without the previous issue. Once fixed, creating a dynamic option list to show strings for the next segment only, will be the target of a dedicated ticket

- The `destination` property of a bridge's subscription requires special UX to support users when editing custom filters and placeholders (like `bridge.name`). It will be done in a different ticket

### Before
screenshot-localhost_3000-2023 10 09-16_23_42

screenshot-localhost_3000-2023 10 10-09_31_05

### After
screenshot-localhost_3000-2023 10 09-16_22_45

screenshot-localhost_3000-2023 10 10-09_31_26


simon622 (approved):
Looks great and functions as expected. BTW - think the having the fields appear underneath each other is a better choice as its less cramped!

PR #156, opened 2023-10-13
feat(17143) - Event API and UX

simon622:
Per the ticket, new API available, new open-api specification generated (version .8) with a Mock of the event stream.


simon622:
Merge for sanity - master will tests sorting

PR #160, opened 2023-10-18
Add github dependency submission Job

pjgg:
**Motivation**

According to GitHub [documentation](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/using-the-dependency-submission-api#about-the-dependency-submission-api), we can use the REST Submission API to submit dependencies for a project. All submitted dependencies will receive Dependabot alerts and Dependabot security updates for any known vulnerabilities.

This task could be accomplished through a [pre-made](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/using-the-dependency-submission-api#using-pre-made-actions) GitHub action.

https://github.com/marketplace/actions/gradle-dependency-submission

### DIFF
--- a/.github/workflows/gradle-dependency-submission.yml
+++ b/.github/workflows/gradle-dependency-submission.yml
@@ -1,0 +1,39 @@
+name: Gradle Dependency Submission
+on:
+  push:
+    branches:
+      - develop
+
+jobs:
+  build:
+    name: Dependencies
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


simon622:
I think this is in the wrong category - seems like a new feature rather than a `SC` fix. Moved to correct category.

schaebo:
@cla-bot check

DC2-DanielKrueger:
We went for renovate.

PR #161, opened 2023-10-18
Renamed datagov-classes to something more meaningful

codepitbull:
**Motivation**

There were several classes/packages with datagov in the name.

**Changes**
Adjust package and classnames to use HiveMQEdge where appropriate or remove the prefix completely.


schaebo:
@cla-bot check

PR #164, opened 2023-10-18
Improve scalability of ProtocolAdapterPollingServiceImpl

hlohse:
Previously, every job run attempted to grab the locks of
executingJobs and semaphore multiple times, concurrently
to other jobs and the watchdog.

In scenarios of high contention due to many connections
and/or a high polling rate, this could lead to high lock
contention and subsequent, costly context switches.

Therefore, this commit removes replaces said locks with:
* A per-job AtomicBoolean to mark it as (not) running.
  This removes the need to check the previous, synchronized
  executingJobs, completely removing multiple lock attempts.
* An AtomicInteger counter of jobs currently running.
  Any job which is the "first" running one will synchronize
  on it to notify the potentially waiting watchdog.
  This significantly reduces the number of job threads which
  might attempt to do this concurrently.

Additionally, replace calls to System.currentTimeMillis with .nanoTime because the former
* is not guaranteed to be strictly monotonically increasing, and
* tends to have a significantly higher overhead


schaebo:
@cla-bot check

PR #165, opened 2023-10-18
Slightly reduce memory consumption per Modbus subscription

hlohse:
EnunMap is more space-efficient than HashMap for Enum keys.


schaebo:
@cla-bot check

PR #166, opened 2023-10-18
Avoid eventual consistency of the HTTP client

hlohse:
The volatile keyword was missing which could have led to
onSamplerInvoked not recognizing for some time that the
adapter already has a client.


schaebo:
@cla-bot check

simon622 (approved):
Confirmed

PR #167, opened 2023-10-20
Refactor monorepo gradle files in order to have a single dependencies version

pjgg:
**Motivation**

In order to avoid issues with transitive dependencies and reduce the number of vulnerabilities (CVEs), it's convenient to have a single version for third-party dependencies, all managed from a central point ([Gradle catalog](https://docs.gradle.org/current/userguide/platforms.html))

This PR is refactoring the current Gradle files to reuse the same dependency declarations and also creates common dependency bundles.

Note: the Gradle refactor was started in the Athens hackathon but I could not finish on time, It took me ..."one" extra day (I finished today (one day later), in the airport). 

IntelliJ will require gradle 8 or gradle 7 with this plugin: https://plugins.jetbrains.com/plugin/18949-gradle-libs-error-suppressor

My recommendation it´s to move on to Gradle 8 . No problem at all on Github Action or the terminal, is just a Intelij issue

### DIFF
--- a/hivemq-edge/build.gradle.kts
+++ b/hivemq-edge/build.gradle.kts
@@ -7,21 +7,21 @@ plugins {
     `java-library`
     `maven-publish`
     signing
-    id("com.github.johnrengelman.shadow")
-    id("com.github.sgtsilvio.gradle.utf8")
-    id("com.github.sgtsilvio.gradle.metadata")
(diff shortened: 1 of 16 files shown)


schaebo:
@cla-bot check

schaebo:
@cla-bot check

DC2-DanielKrueger:
Was already done.

PR #168, opened 2023-10-25
feat(17144): Implement the main UX for the Event Log

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/17144/details/

This PR introduces the `Event Log` into Edge and the basic UX for exporing and exploiting the events. 

The `Event Log` is available from the main navigation menu and gives access to a paginated table of the most recent events in Edge. A limited set of the `Event` properties are available in the table's column (sortable):

-  date created
- severity
- source
- message 

An `Event` can be individually expanded in a right-hand side panel, giving access to the whole properties, in particular, the payload if any. 

The `source` of the `Event` might refer to an entity in Edge (such as an adapter or a bridge). In this case, the `source` is rendered as a link to this entity in context. 

### Out-of-scope
- there are problems with navigation to the editing panel for bridge and adapter, to be fixed in another ticker (see https://hivemq.kanbanize.com/ctrl_board/57/cards/17265/details/) 

- Filtering of columns will be implemented as a feature for the paginated table component (see https://hivemq.kanbanize.com/ctrl_board/57/cards/17145/details/) 

- highlighting the new `Events` since the "last" visit of the user will be implemented in a separate ticket 

### Design
screenshot-localhost_3000-2023 10 25-09_57_54

screenshot-localhost_3000-2023 10 25-09_58_14

screenshot-localhost_3000-2023 10 25-09_58_47

screenshot-localhost_3000-2023 10 25-09_58_59


simon622 (commented):
Feedback about issues on call, this PR does not require a merge since I rebase my PR with this branch, so we can bother work on same branch "feat/17143/event-api"

PR #169, opened 2023-10-25
feat/athens-feature/data-visualisation

vanch3d:
This feature PR is a very quick-and-dirty example on how we can utilise the workspace to allow users to define, in context, a dynamic report containing selected graphs for selected metrics.

The process is intended to run in three steps:
- define a `dataset` from the available metrics of a system's entity (bridge, adapter or edge itself). At the moment, this step is the selection of a single metric (e.g. ` com.hivemq.edge.protocol-adapters.simulation.test.read.publish.success.count ` ) but it could be extended to add multiple data sources (e.g. different metrics in the same graph)

- define a `rendering chart`, using the dataset to populate the data of the visualisation. We are using `Nivo` (https://nivo.rocks/) and are supporting two basic graphs: a Line Chart (ResponsiveLine) and a Bar Chart (ResponsiveBar). `Nivo` charts have a large range of options (formatting, rendering, layout) that could be exposed as a customisation panel to the user

- create a `report` associated with one of the elements of the system by adding charts to the side panel. Access is through the "connector" between the element and Edge, materialising the notion of observability on the data (or the metrics at least) being exchanged between the system. At the moment, it's a simple grid with limited `colspan` but wider layout customisation could be implemented 

### Out-of-scope
- the metrics don't have any persistence of their `datasets`. They are emitted one timestamped data point at a time. Closing the side panel or reloading the app will wipe out the graph. Some local state will need to be implemented for usability

- there is no persistence of the reports created. A simple `localStorage` option have not been implemented

- no extra chart types except the line and the bar charts. Fancy stuff to come when we have time 

- no custom theme to match the `Nivo` graphs style with the HiveMQ design system. The theme mechanism exists but has not been investigated 

- no test! 

- no dedicated accessibility support (but good internal elements from `Nivo`)

### Design
screenshot-localhost_3000-2023 10 25-12_22_29

screenshot-localhost_3000-2023 10 25-12_38_14


schaebo:
@cla-bot check

vanch3d:
This POC has been refactored and integrated into the "metrics" initiative, see https://github.com/hivemq/hivemq-edge/pull/215.
It can be closed

PR #174, opened 2023-10-31
feat(17145): Add optional filtering to the paginated tables

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/17145/details/

This PR adds filters to selected columns of a paginated table. On a valid column, the table renders an input field that can be used to type a value to apply as a filter to the column: 
- only the rows showing the substring will be shown in the table
- a `datalist` will show the matching values that are available for select
-  All the other filterable columns act as "facets": their datalist only shows the matching values after filtering on other columns
- Each filter contains a "Clear" button that will remove the value from the filter and re-populate the table

The PR also fixes the accessibility of the column sorting, see https://hivemq.kanbanize.com/ctrl_board/57/cards/16518/details/:
- by wrapping the sortable column name in a button (for active area)
- by adding the proper `aria-sort` label to the table's column header
- by refactoring the sort icons to a more known type

### Out-of-scope

- number values (e.g. date) should be filtered by a `min` and a `max`. The custom UI will be implemented in a separate ticket
- The `debounce` component introduces too many cycles in the react tree. Improving the display will be addressed in a different ticket
- the `datalist` approach uses a native HTML component. An improved version using UI widgets (`Select`) will be implemented in a separate ticket.

### Before
screenshot-localhost_3000-2023 10 31-16_39_04

### After  
screenshot-localhost_3000-2023 11 01-11_09_43

Screenshot 2023-11-01 at 11 11 02

Screenshot 2023-11-01 at 11 10 33

Screenshot 2023-10-31 at 16 40 55

### DIFF
--- a/hivemq-edge/src/frontend/src/components/PaginatedTable/PaginatedTable.spec.cy.tsx
+++ b/hivemq-edge/src/frontend/src/components/PaginatedTable/PaginatedTable.spec.cy.tsx
@@ -77,21 +77,21 @@ describe('PaginatedTable', () => {
     cy.get('th').eq(0).should('have.text', 'item')
     checkRowOrder()
     cy.get('th').eq(0).click()
-    cy.get('th').eq(0).should('have.text', 'item 🔼')
+    cy.get('th').eq(0).should('have.attr', 'aria-sort', 'ascending')
     checkRowOrder('asc')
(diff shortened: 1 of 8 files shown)


sfrehse (approved):
LGTM and works awesome

PR #178, opened 2023-11-03
adds first sample event for opc ua

schaebo:
**Changes**
Adds first sample event for OPC UA adapter


sfrehse (commented):
Tested and works fine! 👍

simon622 (approved):
Looks good.

PR #179, opened 2023-11-05
fix(17445) Refactor filters to utilise Selct component

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/17445/details/

This PR fixes quite a few problems with the current implementation of the filter component. It replaces the lifecycle-failing debounced input with the more tested `Select` component already used across the app. 

The principles of the filtering remain the same:
- only applies to selected, string-only, columns
- unique values of the column are extracted to serve as options of the component
- a clear button is present as usual, allowing the users to clear the current filter

### Before

screenshot-localhost_3000-2023 11 05-17_26_29

### After 
 
screenshot-localhost_3000-2023 11 05-17_26_08
screenshot-localhost_3000-2023 11 05-17_34_47


sfrehse (approved):
Very snappy! :-)

PR #184, opened 2023-11-08
Add Version updater plugin

DC2-DanielKrueger:
**Motivation**

Reduce manual work on the release

**Changes**

Adds plugin to update the version
(https://hivemq.kanbanize.com/ctrl_board/57/cards/17519/details/)


schaebo:
@cla-bot check

hivemq-edge/src/main/java/com/hivemq/edge/ManifestUtils.java
```diff
@@ -0,0 +1,79 @@
+package com.hivemq.edge;
```
schaebo, line 1: Can you also include the Tests for ManifestUtils ?

build.gradle.kts
```diff
@@ -107,3 +110,23 @@ val hivemqEdgeZip by tasks.registering(Zip::class) {
     }
 }
 
+
+
+val edgeProjectsToUpdate = setOf(
+    "hivemq-edge",
+    "hivemq-edge-module-http",
```
schaebo, line 117: possible improvement: can we determine the list of modules dynamically from checking the modules folder ? This would remove a potential source for mistakes.

schaebo (approved):

build.gradle.kts
DC2-DanielKrueger: hm i guess we could derive it from the dependencies as we have an own configuration for the modules, but unfortunately I dont know how this works as the configurations are not builds

schaebo (approved):

PR #195, opened 2023-11-16
feat(17448): Add a reduced version of the event log to the overview of entities in the workspace

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/17448/details/

This PR extends the content of the overview available in the side panel when the user clicks on one of the entities (adapter or bridge) in the workspace 

It adds a limited version of the event log:
- only showing event for the selected entity (using the filter)
- only shows the 5 most recent events
- columns restricted to the most meaningful ones
- no access to the full payload
- no pagination 
- sorting still works (but might be turned off)

A CTA allows the user to directly access the full `Event Log`, with the filter pre-populated by the selected entity's name (id)

### Out-of-scope
- the payload of events is not available in the summary. The UX necessary to display it requires extended functionalities as used in the `Event Log`. 

### Before 
screenshot-localhost_3000-2023 11 16-12_03_22

### After 
screenshot-localhost_3000-2023 11 16-12_02_52


sfrehse (approved):
Awesome feature. I like the reduced version of it. However, having the payload would be still nice :-P

PR #203, opened 2023-11-17
Bugfix/17745 license check

DC2-DanielKrueger:
**Motivation**

Resolves #<issue>

**Changes**


hivemq-edge/src/test/java/com/hivemq/adapter/AdapterConfigFormatTest.java
```diff
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2019-present HiveMQ GmbH
```
sauroter, line 2: @DC2-DanielKrueger I think this is actually 2023. (Not that it makes a huge difference, but 4 years more IP coverage would be nice ;) )

PR #204, opened 2023-11-17
fix(17700): Improve the UX of the metrics panel in the workspace

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/17700/details/

The PR improves the rendering of metrics in the property side panel of an entity in the workspace and its creation flow.

- metric names are now properly rendered as an human-readable text, both in the sample component and the selector
- in the selector, available metrics are restricted to those pertinent to the selected entity. Metrics already displayed in the panel are disabled 
- the metrics selector is now only available by clicking on the "expandable" icon
- the rendering of a metric has been improved for accessibility (font colour) and for readability (use of numerical increase rather than percentage)
- numbers are properly formatted with locale 

The PR also adds a "copy-to-clipboard" CTA on the side of every Stat panel. It will copy the metric name as a string into the clipboard

### Out-of-scope
- persistency of the selected metrics will be done in a separate ticket

### Before
screenshot-localhost_3000-2023 11 17-16_34_22

### After
screenshot-localhost_3000-2023 11 21-13_02_51

screenshot-localhost_3000-2023 11 21-13_03_52

screenshot-localhost_3000-2023 11 21-13_04_16

screenshot-localhost_3000-2023 11 21-13_04_39

screenshot-localhost_3000-2023 11 21-13_05_27


sfrehse:
Might it be possible to show the complete metric string since that string is used as a direct address of the metric in other systems, like Prometheus/Grafana? Let's link a copy-to-clipboard icon with the tooltip of the raw metric string.

vanch3d:
> Might it be possible to show the complete metric string since that string is used as a direct address of the metric in other systems, like Prometheus/Grafana? Let's link a copy-to-clipboard icon with the tooltip of the raw metric string.

CTA added for every stat panel. Copy the metric in context: 
```json
{"metricName":"com.hivemq.edge.bridge.simons-bridge-there.forward.publish.count","timestamp":"2023-11-21T13:05:20.520Z","value":622308}
```

screenshot-localhost_3000-2023 11 21-13_05_27

sfrehse:
The copy to the clipboard is an excellent feature. Currently, I assume the following use-case/user journey.

The user will check out metrics in the workspace and wants to use them in Grafana / Prometheus.  To get it running in Grafana, the user only needs the metricName field content, which means, just the metric. Users will likely put it directly in the Grafana query editor which requires only the metricName,

vanch3d:
> The copy to the clipboard is an excellent feature. Currently, I assume the following use-case/user journey.
> 
> The user will check out metrics in the workspace and wants to use them in Grafana / Prometheus. To get it running in Grafana, the user only needs the metricName field content, which means, just the metric. Users will likely put it directly in the Grafana query editor which requires only the metricName,

Change made and description updated

sfrehse (approved):

PR #208, opened 2023-11-24
fix(17869): Improve date/time rendering for seconds

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/17869/details/

The PR fixes the rendering of relative date/time for seconds, using the proposed design system, see https://zeroheight.com/9c4326648/p/5495a8-date-and-time

To improve readability, it changes patterns like `10 seconds ago` to `a few seconds ago` and `45 seconds ago`  to `1 minute ago`. It also fixes a bug with `in 0 seconds` for a null duration

### Before
screenshot-localhost_3000-2023 11 24-11_40_02

### After
screenshot-localhost_3000-2023 11 24-11_38_23

### DIFF
--- a/hivemq-edge/src/frontend/src/components/DateTime/DateTimeRenderer.tsx
+++ b/hivemq-edge/src/frontend/src/components/DateTime/DateTimeRenderer.tsx
@@ -1,6 +1,7 @@
 import { FC } from 'react'
 import { Text, Tooltip, type TextProps } from '@chakra-ui/react'
 import { DateTime } from 'luxon'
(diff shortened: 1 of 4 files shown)


sfrehse (approved):
Actually, when I read the notification text, we cannot guarantee the 2 seconds. What about just write "a moment".

vanch3d:
> Actually, when I read the notification text, we cannot guarantee the 2 seconds. What about just write "a moment".

Good point. Will change that quickly

vanch3d:
screenshot-localhost_3000-2023 11 24-15_32_29

PR #213, opened 2023-11-28
Bugfix: Fix missing buffering before first connect of bridge

DC2-DanielKrueger:
**Motivation**

Resolves #<issue>

**Changes**


hivemq-edge/src/main/java/com/hivemq/bridge/mqtt/RemoteMqttForwarder.java
```diff
@@ -194,32 +195,53 @@ private PUBLISH convertPublishAfterBridge(@NotNull PUBLISH publish, int hopCount
...
-        return newPublish;
+        return mqtt5Builder.build();
     }
 
+
     @SuppressWarnings("ResultOfMethodCallIgnored")
-    private void sendPublishToRemote(
+    private synchronized void sendPublishToRemote(
             @NotNull PUBLISH publish, @NotNull String queueId, @NotNull PUBLISH origPublish) {
         final Mqtt5Publish mqtt5Publish = convertPublishForClient(publish);
```
schaebo, line 205: Improvement: The convert could be moved after the following if block, in the case that the bridge is not connected it might be ran twice otherwise

schaebo (approved):

PR #214, opened 2023-11-28
feat(17821): Add support for "persistence" capability 

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/17821/details/

This PR creates the frontend infrastructure required to handle broker capabilities in general (handling the API, verification of a capability) and "persistence" capability.

It adds a tab to the bridge configuration panel, if and only if the `mqtt-persistence` is defined. The content is based on the newly added `persist` property of the `Bridge`. 

### Out-of-scope
- The is no notification to the user when the capability is not available (error or not added to broker). 

### Before
screenshot-localhost_3000-2023 11 28-15_26_47

### After
screenshot-localhost_3000-2023 12 04-13_12_19


sfrehse:
Thanks. I guess we need to shorten the tab titles, however, let's sync with @DC2-DanielKrueger and Owen on that.

vanch3d:
Refactored the translations as per discussion and added a restart warning, activated on after change to the checkbox

screenshot-localhost_3000-2023 12 05-17_51_00

PR #217, opened 2023-11-29
Improvement/mqtt persistence

DC2-DanielKrueger:
**Motivation**

Resolves #<issue>

**Changes**


hivemq-edge/src/main/java/com/hivemq/mqtt/services/PublishDistributorImpl.java
```diff
@@ -148,25 +189,58 @@ private ListenableFuture<PublishStatus> handlePublish(@NotNull final PUBLISH pub
...
+            final boolean shared,
+            final boolean retainAsPublished,
+            @Nullable final ImmutableIntArray subscriptionIdentifier,
+            @Nullable final Long queueLimit) {
+
+
+        Long appliedQueueLimit = queueLimit;
+        int appliedQoS = subscriptionQos;
+        if (shared) {
+            // update with the configuration of the bridge, if it is a bridge client
```
schaebo, line 215: check if id starts with forwarder then we don't need to iterate the bridges if it is not.

Could be placed in `handlePublish` there we already do the qos and queue size checks for normal subs

hivemq-edge/src/main/resources/config.xsd
```diff
@@ -1015,6 +1016,10 @@
         </xs:all>
     </xs:complexType>
 
+    <xs:complexType name="moduleEntity">
```
schaebo, line 1019: this can be removed

schaebo (approved):

hivemq-edge/src/main/java/com/hivemq/HiveMQEdgeMain.java
```diff
@@ -96,18 +103,38 @@ public void bootstrap() throws HiveMQEdgeStartupException, InterruptedException
...
-        injector.initEagerSingletons();
+        Objects.requireNonNull(injector).persistences();
+        try {
+            persistenceStartup.finish();
+        } catch (InterruptedException e) {
+            throw new HiveMQEdgeStartupException(e);
+        }
+        Objects.requireNonNull(injector).initEagerSingletons();
+
         log.trace("Initialized classes in {}ms", (System.currentTimeMillis() - startInit));
```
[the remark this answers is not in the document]
DC2-DanielKrueger: The persistence startup has to finish before the initialization of the other classes as the "normal" persistences need functional local persistences at this point IIRC, I moved the start time measurement after finish. This way we get the correct measurement again.

hivemq-edge/src/main/java/com/hivemq/mqtt/services/PublishDistributorImpl.java
DC2-DanielKrueger: yeah makes sense. I'll put it there.

PR #223, opened 2023-12-01
feat(18003): add localStorage for the configs of the metric panels

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18003/details/

This PR is a follow-up to the main metrics initiative, see https://github.com/hivemq/hivemq-edge/pull/215. It adds persistence to the obervability panels created by the users, ensuring consistency across session. 

The persistency is done through `localStorage`, under individual keys (i.e. `edge.reports-#####`).

### Out-of-scope
- There is no UI-based mechanism to clear all the metric panels, especially the orphans.


simon622 (approved):
Added graphs and stats, refresh pages moved about and noted my configuration was persisted as expected - great work.

PR #226, opened 2023-12-04
Bugfix: Fix bridge Buffering

DC2-DanielKrueger:
**Motivation**

Resolves #<issue>

**Changes**

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/bridge/mqtt/RemoteMqttForwarder.java
+++ b/hivemq-edge/src/main/java/com/hivemq/bridge/mqtt/RemoteMqttForwarder.java
@@ -245,8 +245,8 @@ public class RemoteMqttForwarder implements MqttForwarder {
             } else {
                 perBridgeMetrics.getPublishForwardSuccessCounter().inc();
             }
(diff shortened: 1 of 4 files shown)


hivemq-edge/src/main/java/com/hivemq/bridge/mqtt/RemoteMqttForwarder.java
```diff
@@ -216,14 +236,17 @@ private synchronized void sendPublishToRemote(
...
+        outflightQueue.add(outflightPublishInformation);
         publishResult.whenComplete((mqtt5PublishResult, throwable) -> {
             if (throwable != null) {
                 handlePublishError(publish, throwable);
-                finishProcessing(originalQoS, publish.getUniqueId(), queueId);
             } else {
                 perBridgeMetrics.getPublishForwardSuccessCounter().inc();
-                finishProcessing(originalQoS, publish.getUniqueId(), queueId);
             }
+            outflightQueue.remove(outflightPublishInformation);
```
schaebo, line 248: minor: would be good if we keep the same order between outflightQueue.remove and finishProcessing here and in line 260+

schaebo (approved):

PR #228, opened 2023-12-05
fix(18092): Fix the update of status for adapters and bridges in the wokspace

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18092/details/

This PR changes the way statuses are updated in the `Workspace`. Up to now, any update is relying on the re-mounting of the component every time the users navigate to the page; the list of nodes (adapter and bridges) and edges (links!) would be regenerated with the newest status. 

With the `Workspace` persistence in the workflow, this will not work anymore. 

The PR correctly "listens" to changes in the status of all the entities in the workspace, dynamically changing the graph's nodes and edges as their internal data changes. 

Note: To experience the live update (since navigating between pages will recreate the `Workspace`), open two browsers with the `Workspace` in one and the adapters in the other. Stopping/starting adapters will immediately (give or take the pooling delay) the statuses on the `Worskapce`. 

### After

https://www.loom.com/share/036969c02ef84b679f420afd2665cb3e?sid=ac87304b-4cf8-4747-be29-1760e9dfae33

### DIFF
--- a/hivemq-edge/src/frontend/src/modules/EdgeVisualisation/components/controls/StatusListener.tsx
+++ b/hivemq-edge/src/frontend/src/modules/EdgeVisualisation/components/controls/StatusListener.tsx
@@ -1,22 +1,35 @@
 import { useEffect } from 'react'
 import { useReactFlow } from 'reactflow'
+import { useTheme } from '@chakra-ui/react'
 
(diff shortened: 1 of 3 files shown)


simon622 (approved):
Work nicely. Browsed to workspace - updated connection, status was reflected quickly in the viewer.

PR #232, opened 2023-12-11
feat(18214): Add editor for the group metadata

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18214/details/

This PR adds a simple editor to the group's side panel, allowing users to change the two pieces of information defined by its creation: title and colour scheme.

The editor follows the same pattern as the metrics editor, with a simple collapsible panel with the form.

The colour scheme input is a `Colour Picker` designed to work with the default schemes defined within the `Chakra UI` theme. The number of colours added to the pickers is customisable. 

As with defining a `colourScheme` for other components, the scheme selected for the group will use the two hues `.100` and `.500` for the background and the border respectively

### Before 
screenshot-localhost_3000-2023 12 11-08_59_16

### After 
screenshot-localhost_3000-2023 12 11-08_58_47


simon622 (approved):
love it, I have a rainbow groupings

PR #234, opened 2023-12-11
fix(18218): Fix the availability of the "remove metrics" button for Stats

vanch3d:
fix(18218): Fix the availability of the "remove metrics" button for Stats

See https://hivemq.kanbanize.com/ctrl_board/57/cards/18218/details/

This PR fixes a small oversight: a `Stat` chart (the simple value and increase/decrease indicators) could still be removed from a panel even if the editor was "collapsed".

This was a breach of the design pattern applied to the other charts. Charts cannot be edited or removed from the panel unless the editor is "expanded" and ready for usage.

### Before
screenshot-localhost_3000-2023 12 11-09_18_55

### After 
screenshot-localhost_3000-2023 12 11-09_18_11


simon622 (approved):
all fine

PR #235, opened 2023-12-11
fix(18216): Fix the propagation of status to a collapsed group's edge

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18216/details/

This PR fixes the computation of the summarised status as displayed on the single edge when the group is collapsed 

In an "expanded" view, each one of the grouped nodes still shows its own edge to the `Edge` node. When collapsed, the group node summarises all the edges into a single one connecting it to the `Edge`. 

Its status is based on the union of all the included nodes:
- if one of the nodes is `STOPPED` then the group's status will be `STOPPED`
- if one of the nodes is `DISCONNECTED` then the group's status will be `DISCONNECTED`
- if one of the nodes has no topic, then the group's edge will be shown as `DATA NOT FLOWING`

The purpose is to convey a summary of the group of nodes by singling out the `DISCONNECTED` and `NOT FLOWING` aspects of the group.

### Before
screenshot-localhost_3000-2023 12 11-09_25_53

### After
screenshot-localhost_3000-2023 12 11-09_25_26

### DIFF
--- a/hivemq-edge/src/frontend/src/modules/Workspace/utils/status-utils.spec.ts
+++ b/hivemq-edge/src/frontend/src/modules/Workspace/utils/status-utils.spec.ts
@@ -167,6 +167,11 @@ interface StatusStyleSuite {
 describe('getEdgeStatus', () => {
   const color = MOCK_THEME.colors.status.connected[500]
   const edge: EdgeStyle = {}
+
+  edge.data = {
+    hasTopics: true,
+    isConnected: true,
+  }
   edge.style = {
     strokeWidth: 0.5,
     stroke: color,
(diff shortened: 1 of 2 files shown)


simon622 (changes requested):
Unable to render workspace - see screen shots with errors

Screenshot 2023-12-13 at 14 54 56

Screenshot 2023-12-13 at 14 54 27

vanch3d:
> Unable to render workspace - see screen shots with errors

This is an issue arising with the workspace persisted data (i.e. the graph) getting out of sync with the actual list of adapters/bridges, as returned by the API. The edge case of a backend "reset" has not been considered and is missing a corresponding reset of the `localStorage`. As discussed, we have a quick solution and it will be fixed in a separate ticket.

sfrehse (approved):

PR #236, opened 2023-12-11
fix(18246): fix focus/selected status of nodes in workspace

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18246/details/

This PR fixes a long-standing accessibility issue with the custom nodes in the graph. 

Their `selected` and `focus-visible` statuses are now properly materialised by the similar blue shadow box used for all the other widgets in the UI.

In the graph, it means that nodes (`HiveMQ Edge`, `Adapter`, `Bridge` and `Group`) and the edge's observability button are now properly rendered and responding to keyboard navigation.

### Out-of-scope
- the side panel for the interactive elements is still based on a double-click, which is not keyboard-friendly. It will be fixed in a separate PR.

### Before 
screenshot-localhost_3000-2023 12 11-11_13_35

### After
screenshot-localhost_3000-2023 12 11-11_12_57

### DIFF
--- a/hivemq-edge/src/frontend/package.json
+++ b/hivemq-edge/src/frontend/package.json
@@ -9,7 +9,7 @@
     "lint:eslint": "eslint src --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "lint:prettier": "prettier --check .",
     "lint:prettier:write": "prettier --write .",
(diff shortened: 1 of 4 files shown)


simon622 (approved):
love the softer edges. looks nice

PR #237, opened 2023-12-12
feat(17823): Add form field for the limit of persisted queue message

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/17823/details/

This PR adds a new "queue limit" field to the local subscription form. 

It is part of the persistence feature and the field has conditions:
- if `mqtt-persistence` is not part of the Edge capabilities, the field is no shown
- if the `persist` flag is not turned on, the field is disabled
- if the `maxQoS` for the subscription is set at 0, the field is disabled 

### Out-of-scope
- The range of valid numbers for the limit is undefined. Users could set any number from 10s to millions of messages. With no default, the initial value will be treated at 0. Using the UI "helpers" for number fields (spinners and mouse wheel) cannot deliver supportive behaviour unless an appropriate "step" is defined (e.g. increment by 1000s). The default increment is 1.

### Before
screenshot-localhost_3000-2023 12 12-13_27_55

### After 
screenshot-localhost_3000-2023 12 12-13_26_36

screenshot-localhost_3000-2023 12 12-13_29_42

### DIFF
--- a/hivemq-edge/src/frontend/package.json
+++ b/hivemq-edge/src/frontend/package.json
@@ -11,7 +11,7 @@
     "lint:prettier:write": "prettier --write .",
     "lint:stylelint": "stylelint '{src}/**/*.{css}'",
     "lint:all": "eslint src --ext ts,tsx --report-unused-disable-directives --max-warnings 0 && prettier --check .",
-    "dev:openAPI": "openapi --input '../../../../hivemq-edge/ext/hivemq-edge-openapi-2023.8.yaml' -o ./src/api/__generated__ -c axios --name HiveMqClient --exportSchemas true",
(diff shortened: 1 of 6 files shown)


sfrehse:
I think it make sense to put the numbers of QoS to the labels or similar, since people are used to these numbers.

sfrehse:
> I think it make sense to put the numbers of QoS to the labels or similar, since people are used to these numbers.

I think that it is even more important since we say use the numbers in the queue part.

vanch3d:
> I think it make sense to put the numbers of QoS to the labels or similar, since people are used to these numbers.

Screenshot 2023-12-12 at 18 41 22

PR #244, opened 2023-12-14
refactor(18077): Align the design of the charts with the Edge theme

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18077/details/

The PR refactors the charts in the Observability panel, in order to alig ntheir design with the theme used for the Edge application. 

- The color schemes for the "series" are now based on the `ChakraUI` theme
- each data series in a chart is assign one of the available color scheme
- the tooltip is now consistent across the line and bar charts 
- the legend of the line and barcharts are now properly disaplyed
- the x-axis of the line and chart is now using a proper timestamp formatttng 

The PR also fixes the nature of the group side panel, properly differentiating between "Overview" and "Observability", as for the other nodes.

### Out-of-scope
- The display of multiple series within a chart will be dealt with in another ticket. See https://hivemq.kanbanize.com/ctrl_board/57/cards/18215/details/
- The handling of persisted data is limited to the last 10 samples, and are not persisted in any way. Changing this behaviour, if required, will be dealt with in another ticket. See https://hivemq.kanbanize.com/ctrl_board/57/cards/17964/details/

### Before
screenshot-localhost_3000-2023 12 14-18_16_34

screenshot-localhost_3000-2023 12 14-18_19_41

### After
screenshot-localhost_3000-2023 12 14-18_15_55

screenshot-localhost_3000-2023 12 14-14_55_33

screenshot-localhost_3000-2023 12 15-10_12_38

### DIFF
--- a/hivemq-edge/src/frontend/src/locales/en/translation.json
+++ b/hivemq-edge/src/frontend/src/locales/en/translation.json
@@ -633,6 +633,7 @@
     "property": {
       "header_ADAPTER_NODE": "Adapter Overview",
       "header_BRIDGE_NODE": "Bridge Overview",
+      "header_CLUSTER_NODE": "Group Overview",
       "modify_ADAPTER_NODE": "Modify the adapter",
       "modify_BRIDGE_NODE": "Modify the bridge",
       "eventLog": {
@@ -703,7 +704,7 @@
     "charts": {
       "LineChart": {
         "ariaLabel": {
-          "legend": "timestamp (s)"
+          "legend": "timestamp (seconds ago)"
         }
       }
     },
--- a/hivemq-edge/src/frontend/src/modules/Metrics/Metrics.tsx
+++ b/hivemq-edge/src/frontend/src/modules/Metrics/Metrics.tsx
@@ -10,6 +10,7 @@ import {
   CardBody,
   SimpleGrid,
   useDisclosure,
+  useTheme,
 } from '@chakra-ui/react'
 import { useTranslation } from 'react-i18next'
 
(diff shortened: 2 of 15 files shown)


sfrehse:
When we use some colors, do you think it makes sense to align with the ZeroHeights colors? I guess we have some color to be used in products.

vanch3d:
> When we use some colors, do you think it makes sense to align with the ZeroHeights colors? I guess we have some color to be used in products.

Yes, I've got an epic that will shape the system towards that goal. See https://hivemq.kanbanize.com/ctrl_board/57/cards/18347/details/

sfrehse:
In this PR is wasn't able to create a bounding box around two adapters, but I was able to multi-select two adapters and the create the group.

sfrehse:
I would add some tooltips to the icons when adapters are grouped since people will not understand what functions these buttons have.

Screenshot 2023-12-18 at 21 49 06

vanch3d:
> In this PR is wasn't able to create a bounding box around two adapters, but I was able to multi-select two adapters and the create the group.

I might have an idea about what happened. The drag-and-drop only works while on the canvas of the graph. If your mouse goes outside (a small margin in the page around the canvas) then the select operation is cancelled. No access to the DnD I'm afraid.

vanch3d:
> I would add some tooltips to the icons when adapters are grouped since people will not understand what functions these buttons have.

 Indeed :-) : https://hivemq.kanbanize.com/ctrl_board/57/cards/18382/details/

sfrehse:
> > In this PR is wasn't able to create a bounding box around two adapters, but I was able to multi-select two adapters and the create the group.
> 
> I might have an idea about what happened. The drag-and-drop only works while on the canvas of the graph. If your mouse goes outside (a small margin in the page around the canvas) then the select operation is cancelled. No access to the DnD I'm afraid.

I only get the panning mouse icon.

PR #249, opened 2023-12-18
feat(18348): Add support for dark mode

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18348/details/

This PR is part of an initiative aiming at preparing the styling of `Edge` with the design system being investigated for HiveMQ products.

It activates support for the dark mode and makes every component in the app to be both supportive of the dark mode (e.g. remove hard-coded colours) but also better designed for theming (e.g. using either `colorScheme` or `variant`). 

The refactoring process in this PR tries to ensure that all components from this point forward will be easily styled by a single source of truth, the custom `ChakraUI` theme.

The dark mode is also a very welcome addition to accessibility for `Edge`

A few related issues are also being fixed:

- add a switch button on the navigation menu, allowing users to switch between "light" and "dark" mode
- the `Edge` node in the `Workspace` has a new design
- all the nodes (including the hosts) are now covered by a custom rendering, ensuring adequate theming
- a basic theme has been added to the two charts
- fix a bug with the forward reference of the colour picker  

### Out-of-scope
- Most of the SVGs would benefit from a redesign to support light/dark mode out-of-the-box. At the moment, they require different sources for the two modes. 
- Being part of the app, the login page is also rendered in dark mode. However the switch is not available until users have logged in. This behaviour might have to be changed 

### Before
screenshot-localhost_3000-2023 12 18-16_39_35

### After

screenshot-localhost_3000-2023 12 18-16_40_26

screenshot-localhost_3000-2023 12 18-16_40_48

screenshot-localhost_3000-2023 12 18-16_46_36

screenshot-localhost_3000-2023 12 18-16_47_04

screenshot-localhost_3000-2023 12 18-14_52_33


simon622 (approved):
Looks good and functions as expected

PR #253, opened 2023-12-20
refactor(18350): Refactor the theme strcture

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18350/details/

This PR is mostly a refactoring of the internal theme defined for `Edge`, in order to make it ready for integration with the design system we are building for the `HiveMQ` ecosystem 

It restructures the theme's file system, to allow better flexibility and automation (e.g. change by pull requests). 

It also refactors a few component to ensure consistency in the design and styling of their instances across the application, following these principles:
- styling should be primarily through `variants` of the base component, modifiable though a single source of truth (the theme)
- all instances of the same component should be styled exactly the same (e.g. all main CTA buttons using the `primary` variant)
- 

Notes: 
-  There will be components that are not abidning to the principles above. They will be corrected as soon as work is needed on them, on the basis of this enabling PR. 
- There are limitations in the alignment of components. Many of them are customised by `props` rather than `style` and these properties are not available through theming


vanch3d:
Rebased didn't work

PR #257, opened 2024-01-08
refactor(18025): Reduce size of the main bundles

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18025/details/

This PR is the (latest and correct) attempt to improve the performance (i.e. loading time) of the frontend of Edge. Two operations have been done:

- reducing the size of some of the big SVG images used in the app
- "code-splitting" of the routing components

Code-splitting consists of splitting the main JS bundle into smaller chunks, only initially loading the main chunk required for the application, while the rest is lazy-loaded on demand.

The result is the reduction of the main chunk by almost 50%, significantly increasing the initial loading time.

```
dist/assets/index-e5631a88.js                                1,996.19 kB │ gzip: 634.45 kB
```

```
dist/assets/index-088e36d9.js                                1,121.96 kB │ gzip: 357.19 kB
```

The side-effect of course is a slowing down of the first loading of every other cvhunks. Every "splitted" code needs to be imported when required, i.e. when the relevant component is mounted into the DOM. While being loaded (at first usage), a "loading spinner" will be visible while the page loads.


simon622 (approved):
All looks good and functions as I would expect

PR #260, opened 2024-01-10
fix(17211) - Cascase bridge and adapter metrics on delete / stop

simon622:
**Motivation**

**Changes**

When a bridge is stopped / deleted, the associated metrics are removed.
When an adaper is updated / deletedthe associated metrics are removed.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/bridge/metrics/PerBridgeMetrics.java
+++ b/hivemq-edge/src/main/java/com/hivemq/bridge/metrics/PerBridgeMetrics.java
@@ -19,11 +19,15 @@ import com.codahale.metrics.Counter;
 import com.codahale.metrics.MetricRegistry;
 import com.hivemq.extension.sdk.api.annotations.NotNull;
 import com.hivemq.metrics.HiveMQMetrics;
+import javassist.convert.TransformNew;
+
+import javax.sound.midi.VoiceStatus;
+import java.util.HashSet;
+import java.util.Set;
 
 public class PerBridgeMetrics {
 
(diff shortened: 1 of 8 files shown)


hivemq-edge/src/main/java/com/hivemq/bridge/metrics/PerBridgeMetrics.java
```diff
@@ -34,54 +39,68 @@ public class PerBridgeMetrics {
...
+        loopPreventionRemoteDropCounter = createBridgeCounter(metricRegistry,
                 bridgeName,
                 "remote.publish.loop-hops-exceeded",
-                "count"));
+                "count");
+    }
+
+    private Counter createBridgeCounter(final @NotNull MetricRegistry metricRegistry, @NotNull final String... names){
+        final String metricName = MetricRegistry.name(BRIDGE_PREFIX, names);
+        synchronized (mutex){
```
DC2-DanielKrueger, line 100: What is the point of creating a mutex object and not synchronizing on 'this'? Would have the same semantic as the mutex is a normal field, right?

DC2-DanielKrueger (approved):

hivemq-edge/src/main/java/com/hivemq/bridge/metrics/PerBridgeMetrics.java
simon622: Generally I consider synchronizing on "this" a bad pattern, it means external objects can cause your encapsulated state change to be affected by external conditions. Nothing is stopping arbitrary code sync'ing the reference and locking you up. Encapsulation of these semantics wins in highly contended threaded code

PR #264, opened 2024-01-17
Feature/basic integration

DC2-DanielKrueger:
**Motivation**

Resolves #<issue>

**Changes**

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/edge/modules/ModuleLoader.java
+++ b/hivemq-edge/src/main/java/com/hivemq/edge/modules/ModuleLoader.java
@@ -133,8 +133,10 @@ public class ModuleLoader {
         final ArrayList<Class<? extends T>> classes = new ArrayList<>();
         for (EdgeModule module : modules) {
             try {
+                log.info("Trying to load class '{}' from file '{}'.", serviceClazz, module.root.getAbsoluteFile());
                 final Iterable<Class<? extends T>> loaded = classServiceLoader.load(serviceClazz, module.classloader);
                 for (Class<? extends T> foundClass : loaded) {
+                    log.info("Adding class '{}' from file '{}'.", foundClass, module.root.getAbsoluteFile());
                     classes.add(foundClass);
                 }
             } catch (IOException | ClassNotFoundException e) {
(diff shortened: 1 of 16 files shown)


hivemq-edge/src/main/java/com/hivemq/edge/modules/ModuleLoader.java
```diff
@@ -133,8 +133,10 @@ protected void loadFromModulesDirectory(final ClassLoader parentClassloader) {
         final ArrayList<Class<? extends T>> classes = new ArrayList<>();
         for (EdgeModule module : modules) {
             try {
+                log.info("Trying to load class '{}' from file '{}'.", serviceClazz, module.root.getAbsoluteFile());
```
schaebo, line 136: I think these should be trace/debug instead

hivemq-edge/src/main/java/com/hivemq/edge/modules/ModuleLoader.java
```diff
@@ -133,8 +133,10 @@ protected void loadFromModulesDirectory(final ClassLoader parentClassloader) {
...
                 final Iterable<Class<? extends T>> loaded = classServiceLoader.load(serviceClazz, module.classloader);
                 for (Class<? extends T> foundClass : loaded) {
+                    log.info("Adding class '{}' from file '{}'.", foundClass, module.root.getAbsoluteFile());
```
schaebo, line 139: I think these should be trace/debug instead

hivemq-edge/src/main/java/com/hivemq/extensions/core/RestComponentsService.java
```diff
@@ -0,0 +1,17 @@
+package com.hivemq.extensions.core;
+
+import com.hivemq.api.resources.GenericAPIHolder;
+import com.hivemq.extension.sdk.api.annotations.NotNull;
+
+public class RestComponentsService {
```
schaebo, line 6: may make sense to add an interface here as it is exposed to the modules

DC2-DanielKrueger: yeah ill do so

schaebo (approved):

PR #267, opened 2024-01-22
Feature/handler integration

DC2-DanielKrueger:
https://hivemq.kanbanize.com/ctrl_board/57/cards/16335/details/

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/bootstrap/netty/initializer/AbstractChannelInitializer.java
+++ b/hivemq-edge/src/main/java/com/hivemq/bootstrap/netty/initializer/AbstractChannelInitializer.java
@@ -17,12 +17,14 @@ package com.hivemq.bootstrap.netty.initializer;
 
 import com.google.common.base.Preconditions;
 import com.hivemq.bootstrap.ClientConnection;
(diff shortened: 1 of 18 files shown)


hivemq-edge/src/main/java/com/hivemq/bootstrap/netty/initializer/AbstractChannelInitializer.java
```diff
@@ -106,6 +108,12 @@ protected void initChannel(final @NotNull T ch) throws Exception {
         //MQTT_5_FLOW_CONTROL_HANDLER is added here after CONNECT
         ch.pipeline()
                 .addLast(MQTT_MESSAGE_BARRIER, new MessageBarrier(channelDependencies.getMqttServerDisconnector()));
+
+        final @Nullable HandlerPackage handlerPackage = channelDependencies.getHandlerProvider().get();
```
schaebo, line 112: Not necessary for this, but it could make sense to also offer an insertAfter with the previous handler as param instead of doing addLast, then the position in the pipeline is not fixed to the end

schaebo (approved):

PR #269, opened 2024-01-23
refactor(18840): Change the heading for titles in the node forms

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18840/details/

The PR changes the default heading used for titles in the JSONSchema form to `h2`.

It fixes an accessibility issue with the heading preferred to be incremented by 1 on nesting (the previous value was `h5`). 

### Before
screenshot-localhost_3000-2024 01 23-10_40_57

### After
screenshot-localhost_3000-2024 01 23-10_43_36

### DIFF
--- a/hivemq-edge/src/frontend/src/extensions/datahub/components/helpers/ReactFlowSchemaForm.tsx
+++ b/hivemq-edge/src/frontend/src/extensions/datahub/components/helpers/ReactFlowSchemaForm.tsx
@@ -11,11 +11,27 @@ import {
   getUiOptions,
   ErrorListProps,
   TranslatableString,
+  TitleFieldProps,
 } from '@rjsf/utils'
 import { GenericObjectType } from '@rjsf/utils/src/types.ts'
 import validator from '@rjsf/validator-ajv8'
-import { Alert, AlertTitle, Box, FormControl, List, ListIcon, ListItem, Text } from '@chakra-ui/react'
+import { Alert, AlertTitle, Box, Divider, FormControl, Heading, List, ListIcon, ListItem, Text } from '@chakra-ui/react'
 import { WarningIcon } from '@chakra-ui/icons'
+
+// overriding the heading definition
+function TitleFieldTemplate<T = unknown, S extends StrictRJSFSchema = RJSFSchema>({
+  id,
+  title,
+}: TitleFieldProps<T, S>) {
+  return (
+    <Box id={id} mt={1} mb={4}>
+      <Heading as={'h2'} size={'lg'}>
+        {title}
+      </Heading>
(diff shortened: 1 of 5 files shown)


sfrehse (approved):

hivemq-edge/src/frontend/src/extensions/datahub/components/helpers/ReactFlowSchemaForm.tsx
```diff
@@ -11,12 +11,28 @@ import {
...
   TranslatableString,
+  TitleFieldProps,
 } from '@rjsf/utils'
 import { GenericObjectType } from '@rjsf/utils/src/types.ts'
 import validator from '@rjsf/validator-ajv8'
-import { Alert, AlertTitle, Box, FormControl, List, ListIcon, ListItem, Text } from '@chakra-ui/react'
+import { Alert, AlertTitle, Box, Divider, FormControl, Heading, List, ListIcon, ListItem, Text } from '@chakra-ui/react'
 import { WarningIcon } from '@chakra-ui/icons'
 
+// overriding the heading definition
```
h2xd, line 21: question(non-blocking): do we actually need the comment?

Because to me personally it's very clear what the code is doing.

```suggestion
```

h2xd (approved):
LGTM! 🐊 

Nothing blocking from my side.

hivemq-edge/src/frontend/src/extensions/datahub/components/helpers/ReactFlowSchemaForm.tsx
vanch3d: Ah, that's a good one. 
Technically no. 
But then I'm replicating the original piece of code, just to change two characters. But I don't have the history of the changes to see the difference. So I keep track of the reason for the override. 
But then, I got lazy! I used the shorthand for the object construction, reusing the same name:
```
      templates={{
        DescriptionFieldTemplate,
        FieldTemplate,
        ErrorListTemplate,
        TitleFieldTemplate,
      }}
```
So, the best way should be to rename my template in a meaningful way, e.g. `TitleFieldTemplateWithH2` and get no too lazy:

```
      templates={{
        DescriptionFieldTemplate,
        FieldTemplate,
        ErrorListTemplate,
        TitleFieldTemplate: TitleFieldTemplateWithH2,
      }}
```

vanch3d: I got a ticket for rationalising the templates across both `Edge`, `DataHub` and potentially an external module. Will get a naming/commenting scheme in place

PR #276, opened 2024-01-31
feat(18844): Add a simple UX connection between Edge and DataHub

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18844/details/

This PR creates the first step in allowing the users to naturally navigate between the core functionality of `Edge` and `DataHub`

The `Topic Filters` element of the data policy flow now accepts an optional `Adapter` in the configuration panel. The select widget contains the list of all live adapters and can be searched for instances. 

Once selected, two extra features will be enabled: 
- in the `Workspace`, the link that connects this adapter to `Edge` will now contain a second icon, which allows users to navigate to the `DataHub` sketch directly, selecting the relevant node
- in the `Topic Filters` configuration panel, the list of strings now runs an extra validation step: if the edited topic filters don't match one of the topic subscriptions defined in the adapter, an error will be raised.

Note that selecting an adapter is optional and doesn't change the current approach of filters. They still have to be edited individually (and therefore can be generated by an external source to define the `matching` property of the payload).

The PR also adds a small fix to the `Vite`/`TypeScript` configuration, defining the `@datahub` alias for the file import and allowing proper containerisation of the "extension".

### Out-of-scope
- The configuration panel only supports `Adapter` for the time being. Support for topics from `Bridge` will be added in a subsequent ticket.

### Before
  
screenshot-localhost_3000-2024 01 31-17_51_29
screenshot-localhost_3000-2024 01 31-17_51_59

### After
screenshot-localhost_3000-2024 01 31-17_50_13

screenshot-localhost_3000-2024 01 31-17_50_26

### DIFF
--- a/hivemq-edge/src/frontend/src/extensions/datahub/api/__generated__/schemas/TopicFilterData.json
+++ b/hivemq-edge/src/frontend/src/extensions/datahub/api/__generated__/schemas/TopicFilterData.json
@@ -2,6 +2,11 @@
   "type": "object",
   "description": "To start a policy, you have to indicate a topic (or a topic filter) to apply it on. This\n         is a simple way to handle this operation, until we can use the Topic Tree Selector.\n          ",
   "properties": {
+    "adapter": {
+      "type": "string",
+      "title": "Adapter source",
+      "description": "If an adapter is selected, the filters created below will be validated from the subscriptions"
+    },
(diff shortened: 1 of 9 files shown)


h2xd (approved):
LGTM! 🐊

PR #278, opened 2024-02-01
refactor(18991): Refactor "Transform" Operation to include serialisers

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18991/details/

This PR changes the UX for the `DataHub.transform` Operation definition to add the `deserialiser` and `serialiser` handles to the node. It allows the pipeline not to have to explicitly use the two dedicated nodes and makes the abstraction more meaningful to the users. It also changes the schema, to allow multiple functions to be added.

The API payload is not changing: the node will be converted to the three individual actions when submitted. 

### Before
screenshot-localhost_3000-2024 02 01-09_29_10

### After
screenshot-localhost_3000-2024 02 01-09_27_18

### DIFF
--- a/hivemq-edge/src/frontend/src/extensions/datahub/components/nodes/OperationNode.tsx
+++ b/hivemq-edge/src/frontend/src/extensions/datahub/components/nodes/OperationNode.tsx
@@ -3,27 +3,44 @@ import { useTranslation } from 'react-i18next'
 import { NodeProps, Position } from 'reactflow'
 import { HStack, Text, VStack } from '@chakra-ui/react'
 
-import { DataHubNodeType, OperationData } from '../../types.ts'
-import NodeIcon from '../helpers/NodeIcon.tsx'
-import NodeParams from '../helpers/NodeParams.tsx'
-import { NodeWrapper } from './NodeWrapper.tsx'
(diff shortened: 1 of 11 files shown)


hivemq-edge/src/frontend/src/extensions/datahub/components/nodes/OperationNode.tsx
```diff
@@ -3,27 +3,44 @@ import { useTranslation } from 'react-i18next'
...
+import { CustomHandle } from '@datahub/components/nodes/CustomHandle.tsx'
 
 export const OperationNode: FC<NodeProps<OperationData>> = (props) => {
   const { t } = useTranslation('datahub')
   const { data, id, type } = props
   const { functionId, metadata } = data
 
+  const isTransform = metadata?.hasArguments && data.functionId === 'DataHub.transform'
+  const isSerialiser =
+    metadata?.hasArguments && (data.functionId === 'Serdes.serialize' || data.functionId === 'Serdes.deserialize')
```
h2xd, line 19: suggestion: does it make sense to put the magic strings into a constant?

h2xd (approved):
LGTM! 🐊

PR #300, opened 2024-02-29
fix(19825): Ensure resources are fully published before handling policy requests

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/19825/details/

The PR reorganises the batch of resource requests (e.g. `script` and `schema`1) to ensure that they are all settled before the main policy request (e.g. `data policy` or `behavior policy`) is being processed. 

Each request triggers an error or success toast, as expected. 

The policy request will only be executed if EVERY resource request is successful. A success or error toast is also triggered 

Note that is a resource already exists, the relevant resource request will fail, triggering a fail for the policy request. This behaviour might need to be changedwe consider that an existing resource is still a correct one for the policy to reference to.


h2xd (approved):
LGTM! 🐊 

question(non-blocking): Would it make sense to create a `const latestVersion = 'latest'` and utilise that? Unsure if that will change later and supports faster changes and identification of the concern later on.

PR #302, opened 2024-03-05
fix(19348) - Modbus change data calculation should only publish deltas

simon622:
When the checkbox only publish data items that have changed since last poll was set, the entire previous graph (which may have contained multiple indices) was sent with each poll if a single items had changed. The calculation now considered each item individually.

### DIFF
--- a/modules/hivemq-edge-module-modbus/src/main/java/com/hivemq/edge/adapters/modbus/ModbusAdapterConfig.java
+++ b/modules/hivemq-edge-module-modbus/src/main/java/com/hivemq/edge/adapters/modbus/ModbusAdapterConfig.java
@@ -27,9 +27,7 @@ import com.hivemq.extension.sdk.api.annotations.NotNull;
 import com.hivemq.extension.sdk.api.annotations.Nullable;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 
 public class ModbusAdapterConfig extends AbstractPollingProtocolAdapterConfig {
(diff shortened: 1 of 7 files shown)


modules/hivemq-edge-module-modbus/src/test/java/com/hivemq/edge/adapters/modbus/ModbusProtocolAdapterTest.java
```diff
@@ -65,4 +68,39 @@ void test_captureDataSample_expectedPayloadPresent()
...
     }
 
+    @Test
+    void test_deltaSamples() {
+
+        final ModBusData data1 = createSampleData(10);
+        final ModBusData data2 = createSampleData(10);
+
+        Assertions.assertEquals(0,
+                AdapterDataUtils.margeChangedSamples(data1.getDataPoints(), data2.getDataPoints()).size(), "There should be no deltas");
```
DC2-DanielKrueger, line 78: Should this method really be named margeChangedSamples? I couldnt find an explanation for the word. Is this a typo and it should be named merge..?

modules/hivemq-edge-module-modbus/src/main/java/com/hivemq/edge/adapters/modbus/ModbusAdapterConfig.java
```diff
@@ -100,9 +103,31 @@ public static class Subscription extends AbstractProtocolAdapterConfig.Subscript
...
         public @NotNull AddressRange getAddressRange() {
             return addressRange;
         }
+
+        @Override
+        public boolean equals(final Object o) {
+            if (this == o) return true;
+            if (!(o instanceof Subscription)) return false;
+            Subscription that = (Subscription) o;
+            return Objects.equals(addressRange, that.addressRange);
```
DC2-DanielKrueger, line 124: shouldnt be QoS and Destination also be part of equals and hashcode? If there is a reason to not have them inside please make a comment so we avoid a future bug.

DC2-DanielKrueger (approved):
LGTM, two things to check.

modules/hivemq-edge-module-modbus/src/test/java/com/hivemq/edge/adapters/modbus/ModbusProtocolAdapterTest.java
simon622: Typo! Doh! Will fix

sfrehse:
@DC2-DanielKrueger could ¥ou please resolve the conflict and let merge this PR.

PR #304, opened 2024-03-07
feat(19565) - Arbitrary properties on subscriptions

simon622:
Any adapters using the abstract polling technology will benefit from the ability of putting arbitrary properties on their application messages whether in single or multi tag mode


hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/AbstractProtocolAdapter.java
```diff
@@ -143,6 +147,13 @@ protected static TagSample createTagSample(final @NotNull ProtocolAdapterDataSam
         return new TagSample(includeTagName ? dataPoint.getTagName() : null, dataPoint.getTagValue());
     }
 
+    protected AbstractProtocolAdapterJsonPayload decoratePayloadMessage(ProtocolAdapterDataSample<T> sample, AbstractProtocolAdapterJsonPayload payload){
+        if(sample.getUserProperties() != null && !sample.getUserProperties().isEmpty()){
+            payload.setUserProperties(sample.getUserProperties());
```
DC2-DanielKrueger, line 152: it could make sense to make an add method that only adds UPs. Otherwise we lose existing ones. (which are none right now, so its not a real problem)

DC2-DanielKrueger (approved):

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/AbstractProtocolAdapter.java
simon622: This is only called by the abstract once, and is meant as a hook. If others wish to utilise it, they should override the method and provide their own decoration (probably called super.decorate first)

PR #307, opened 2024-03-08
refactor(20139): Add validation and error handling for PROTOBUF code

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/20139/details/

The PR fixes a few bugs with the text version of PROTOBUF in the `Schema` node and its side panel. 

A field for selecting the active message has been added to the JSON Schema of the PROTOBUF configuration. 
Validation of the code and error reporting are now supported. 

The PROTOBUF code is now compiled to a descriptor buffer and then encoded to a `base64` fully on the browser side. It means the end-user has access to the full flow of PROTOBUF schemas, from the editing (or copy-and-paste) to the validation, then to the submission to the backend, then to the listing of created schemas, then finally to the modifying of the original code. 

The client-side encoding relies on a self-described experimental feature of the `protobufjs` library (and a 5 years old extension) so an encoding-decoding test has been added to validate the PROTOBUF source. 

### Out-of-scope
- loading of a PROTOBUF compiled descriptor, as part of a policy, will be handled in a different ticket

### After

screenshot-localhost_3000-2024 03 08-10_41_06

screenshot-localhost_3000-2024 03 08-10_42_22

### DIFF
--- a/hivemq-edge/src/frontend/package.json
+++ b/hivemq-edge/src/frontend/package.json
@@ -60,6 +60,7 @@
     "i18next": "^22.4.15",
     "luxon": "^3.3.0",
     "mermaid": "^10.8.0",
+    "protobufjs": "^7.2.6",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-hook-form": "^7.43.9",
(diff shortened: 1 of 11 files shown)


h2xd (approved):
LGTM! 🐊

PR #309, opened 2024-03-08
refactor(20148): Add a feature flag to disable behaviour policies

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/20148/details/

The PR adds a feature flag (`VITE_FLAG_DATAHUB_PROTOBUF_ENABLED`) to disabled the behavior policies from the frontend:
- Behavior policies are not listed on the Data Hub table
- Drag-and-drop icons for the behavior policy and transition nodes are hidden in the toobar

By default, the flag is disabled. To change the flag, edit your `.env.local` (or `.env` for a github release) with the following property: 

```.env
VITE_FLAG_DATAHUB_BEHAVIOR_ENABLED=true
```

### After


h2xd (approved):
LGTM! 🐊

PR #311, opened 2024-03-11
fix(20147): allow data policy to be created without validators

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/20147/details/

The PR fixes a bug with the check of `Data Policies`, which is now accepting them to be without any explicit validation. 

It also fixes translations, in particular the `Dat Hub` navigation menu

### After

screenshot-localhost_3000-2024 03 12-11_23_49


oli-hivemq (approved):
LGTM 🦣

PR #313, opened 2024-03-12
Fix(18091) - Payload in HTTP adapter

simon622:
**Motivation**

Resolves #18091

I have added a new checkbox “Assert JSON Payload” which, when checked will always attempt to parse the payload as JSON regardless of the content type on the response object. I considered using a dropdown with multiple content types for the response, but actually the reality is that it will almost always be either dynamic (existing behaviour) OR JSON.

**Changes**

### DIFF
--- a/modules/hivemq-edge-module-http/src/main/java/com/hivemq/edge/adapters/http/HttpProtocolAdapter.java
+++ b/modules/hivemq-edge-module-http/src/main/java/com/hivemq/edge/adapters/http/HttpProtocolAdapter.java
@@ -187,7 +187,7 @@ public class HttpProtocolAdapter extends AbstractPollingProtocolAdapter<HttpAdap
                         payloadData = objectMapper.readTree(bodyData);
                     } catch (Exception e){
                         if(log.isDebugEnabled()){
-                            log.debug("Invalid json data was [{}]", bodyData);
+                            log.debug("Invalid JSON data was [{}]", bodyData);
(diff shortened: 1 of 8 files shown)


modules/hivemq-edge-module-http/src/main/java/com/hivemq/edge/adapters/http/HttpProtocolAdapter.java
```diff
@@ -178,16 +179,21 @@ protected HttpData readResponse(@NotNull final HttpAdapterConfig config, final @
...
+             if(bodyData != null){
                 responseContentType = response.headers().firstValue(HttpConstants.CONTENT_TYPE_HEADER).orElse(null);
+                responseContentType = config.isAssertResponseIsJson() ? HttpConstants.JSON_MIME_TYPE : responseContentType;
                 if(HttpConstants.JSON_MIME_TYPE.equals(responseContentType)) {
                     try {
                         payloadData = objectMapper.readTree(bodyData);
                     } catch (Exception e){
-                        log.warn("Error encountered marshalling HTTP response data to json", e);
                         if(log.isDebugEnabled()){
                             log.debug("Invalid json data was [{}]", bodyData);
```
sfrehse, line 190:
```suggestion
                            log.debug("Invalid JSON data was [{}]", bodyData);
```

sfrehse (approved):

PR #314, opened 2024-03-12
refactor(18960): Add ordering of connected nodes

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18960/details/

This PR addresses the issue of ordering connected nodes when appropriates. There are two cases treated.

With `Validator` nodes, the configuration in the side panel showed a list of schema object, to be created or ordererd. That list is now gone. Handling schemas and validator is done purely by connections, their order being unimportant.

With the `transform` functions of an `Operation` operation node, it is possible to connect multiple `Function` nodes to the `Operation`. But their order matters. So the configuration of the `Operation` side panel now shows the list of function names connected to the node. Their order can be changed (but they cannot be deleted or edited, this is handled directly by the connections or the nodes themselves). 

### After


hivemq-edge/src/frontend/src/extensions/datahub/designer/operation/OperationPanel.tsx
```diff
@@ -8,15 +8,28 @@ import useDataHubDraftStore from '@datahub/hooks/useDataHubDraftStore.ts'
...
 
   const formData = useMemo(() => {
     const adapterNode = nodes.find((e) => e.id === selectedNode) as Node<OperationData> | undefined
     if (!adapterNode?.data) return null
+
+    if (adapterNode.data.functionId === OperationData.Function.DATAHUB_TRANSFORM) {
+      const connectedFunctionIds = getIncomers(adapterNode, nodes, edges)
+        .filter(isFunctionNodeType)
+        .map((e) => e.data.name)
+      const gg = adapterNode.data.formData as unknown as OperationData.DataHubTransformType
```
h2xd, line 24: suggestion: check on the variable name `gg` might be confusing later

```suggestion
      const gg = adapterNode.data.formData as unknown as OperationData.DataHubTransformType
```

h2xd (approved):
LGTM! 🐊

PR #321, opened 2024-03-14
feat(20401): Support downloading payloads and read-only canvas

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/20401/details/

The PR adds two features allowing a partial release of the `Data Hub Designer`:
- the payload of the `policies`, `scripts` and `schemas` can be downloaded from the main tables. The JSON is exported and minified in a file automatically named from the `id` of the object and the timestamp of the download.
- access to the canvas of a saved policy is through a "view" command rather than the previous "edit". The canvas can still be explored, manipulated (and, unfortunately, modified) but it cannot be checked or published

The PR also fixes a small issue with the "clearing the canvas" action by navigating back to the right location. 

### Before
screenshot-localhost_3000-2024 03 14-12_54_22

### After
screenshot-localhost_3000-2024 03 14-12_53_34

screenshot-localhost_3000-2024 03 14-12_53_47

### DIFF
--- a/hivemq-edge/src/frontend/src/extensions/datahub/components/controls/ToolboxDryRun.tsx
+++ b/hivemq-edge/src/frontend/src/extensions/datahub/components/controls/ToolboxDryRun.tsx
@@ -12,7 +12,7 @@ import {
   AlertStatus,
   CloseButton,
 } from '@chakra-ui/react'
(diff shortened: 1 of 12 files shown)


h2xd (approved):
LGTM! 🐊

PR #323, opened 2024-03-15
fix(hotfix): Remove client filter and interpolation for the release version

vanch3d:
The PR hides the client filter button from the toolbar and the interpolation buttoins from the `Sytem.log` operation configuration panel 

They are hidden for the current release behind two internal feature flags

### After
screenshot-localhost_3000-2024 03 15-13_59_32

screenshot-localhost_3000-2024 03 15-14_02_06


sfrehse (approved):
👍

PR #348, opened 2024-04-02
featr(20797): Integrate workspace and data hub 

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/20797/details/

The PR improves the integration between the workspace and the Data Hub policies.

On every edge connecting an adapter or a bridge, if a topic in the subscriptions matches the `topic filter` of a policy, an icon is added to allow direct access to the `policy designer` 

If there are more policies than a certain threshold (currently set internally at 3), then a `+n` icon is added instead, linking to the (not-yet-filtered) list of policies. This is to ensure that the icons are not going to spread too widely on the workspace.

Not that the icons, in line with access to the Observability panel, are displayed consistently across both nodes and groups and reflect the collapse/expended status of groups

### Before 
screenshot-localhost_3000-2024 04 02-10_06_02

### After 
screenshot-localhost_3000-2024 04 02-10_04_25


hivemq-edge/src/frontend/src/modules/Workspace/components/edges/ObservabilityEdgeCTA.spec.cy.tsx
```diff
@@ -0,0 +1,24 @@
...
+describe('ObservabilityEdgeCTA', () => {
+  beforeEach(() => {
+    cy.viewport(400, 400)
+  })
+
+  it.only('should render the button', () => {
+    cy.mountWithProviders(<ObservabilityEdgeCTA source="1" onClick={cy.stub().as('onClick')} />)
+    cy.get('button').eq(0).should('not.have.attr', 'aria-describedby')
+    cy.get('button').eq(0).focus()
+    cy.get('button').eq(0).should('have.attr', 'aria-describedby')
```
tolumq, line 12: If possible, we should test the value of the attribute in this case, wdyt?
something like:
```
cy.get(`[aria-describedby="the id of the descriptor"]`)
```

hivemq-edge/src/frontend/src/modules/Workspace/hooks/useGetPoliciesMatching.ts
```diff
@@ -0,0 +1,65 @@
...
+//TODO[NVL] return isLoading, isError and maybe error?
+export const useGetPoliciesMatching = (id: string) => {
+  const hasDataHub = useGetCapability(CAPABILITY.DATAHUB)
+  const { isLoading: isDataLoading, data: dataPolicies, isError: isDataError } = useGetAllDataPolicies()
+  const { nodes: workspaceNodes } = useWorkspaceStore()
+  const { data: protocols } = useGetAdapterTypes()
+
+  if (!hasDataHub) return undefined
+  if (isDataLoading) return undefined
+  if (isDataError) return undefined
```
tolumq, line 20: I think we can combine this 3 into one:
```
const undefinedState = !hasDataHub | isDataLoading | isDataError;
if(undefinedState) return undefined
```
wdyt?

hivemq-edge/src/frontend/src/modules/Workspace/hooks/useGetPoliciesMatching.ts
```diff
@@ -0,0 +1,65 @@
...
+    return dataPolicies.items?.filter((policy) => allTopics.includes(policy.matching.topicFilter))
+  }
+
+  if (sourceNode.type === NodeTypes.ADAPTER_NODE) {
+    return getPoliciesForAdapter(sourceNode)
+  }
+
+  if (sourceNode.type === NodeTypes.CLUSTER_NODE) {
+    const group = sourceNode.data as Group
+    const adapterIDs = group.childrenNodeIds.map<Node | undefined>((e) => workspaceNodes.find((x) => x.id === e))
```
tolumq, line 52: I think a `filter` just after the `map` might be more appropriate here, so we don't need `line 55` and can do without an `undefined` type here. wdyt?

hivemq-edge/src/frontend/src/modules/Workspace/components/edges/ObservabilityEdgeCTA.spec.cy.tsx
```diff
@@ -0,0 +1,24 @@
...
+    cy.getByTestId(`icon-button-tooltip`).eq(0).should('contain.text', 'Open the Observability panel')
+    cy.get('button').eq(0).click()
+    cy.get('@onClick').should('have.been.called')
```
tolumq, line 15: [nit]: I see the `click` event on line 14, but you might want to test when the `@onClick` `.not.to.have.been.called` was true, so the test itself can tell a story, and easily catch possible future bugs.

hivemq-edge/src/frontend/src/modules/Workspace/hooks/useGetPoliciesMatching.ts
```diff
@@ -0,0 +1,65 @@
...
+    return dataPolicies.items?.filter((policy) => allTopics.includes(policy.matching.topicFilter))
+  }
+
+  const getPoliciesForAdapter = (node: Node): DataPolicy[] | undefined => {
+    const adapterProtocol = protocols?.items?.find((e) => e.id === node.data.type)
+    if (!adapterProtocol) return undefined
+
+    const allTopics = discoverAdapterTopics(adapterProtocol, node.data.config)
+      .map((e) => ({ topic: e }))
+      .map((e) => e.topic)
```
tolumq, line 40: It looks like we don't need this two lines, wdyt?

hivemq-edge/src/frontend/src/modules/Workspace/hooks/useGetPoliciesMatching.ts
```diff
@@ -0,0 +1,65 @@
...
+    const group = sourceNode.data as Group
+    const adapterIDs = group.childrenNodeIds.map<Node | undefined>((e) => workspaceNodes.find((x) => x.id === e))
+
+    const policies: DataPolicy[] = []
+    for (const node of adapterIDs) {
+      if (node) {
+        const hh = getPoliciesForAdapter(node)
+        if (hh) policies.push(...hh)
+      }
+    }
```
tolumq, line 60: we could instead have:
```
const policies = group.childrenNodeIds.map(e => workspaceNodes.find((x) => x.id === e)).filter(node => !!node).map(node => ...getPoliciesForAdapter(node));

// to avoid making readability complex, we can then continue with line 61, wdyt?
```

tolumq (commented):
Good PR, thank you. I left some suggestions, nothing too serious.

hivemq-edge/src/frontend/src/components/MQTT/TopicCreatableSelect.spec.cy.tsx
```diff
@@ -97,7 +97,7 @@ describe.only('MultiTopicsCreatableSelect', () => {
     cy.get('#react-select-3-listbox').contains('Add the topic ... 123')
     // cy.get('#my-id').type('{Enter}')
 
-    cy.get('#react-select-3-option-4').click()
+    cy.get('#react-select-3-option-3').click()
```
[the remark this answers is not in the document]
vanch3d: Thanks, nicely spotted. Unfortunately, this part of the `react-select` doesn't allow extra props to be added easily. 
But refactored the selector in a slightly better way 

```
    cy.get('[role="option"]').eq(0).click()

```

hivemq-edge/src/frontend/src/modules/Workspace/components/edges/ObservabilityEdgeCTA.spec.cy.tsx
vanch3d: You mean making sure it's not.to.have.been.called BEFORE the click? 
Do you think it's a good pattern or an overkill? 
It should be done for all the click and have.been.called

hivemq-edge/src/frontend/src/modules/Workspace/hooks/useGetPoliciesMatching.ts
vanch3d: Whoa. Nicely spotted. Not sure how this came into existence!

hivemq-edge/src/frontend/src/modules/Workspace/hooks/useGetPoliciesMatching.ts
vanch3d: I tried. Unfortunately, filter doesn't change the type so you still need a cast or a test to exclude the undefined for the next map. It gets too complicated.

hivemq-edge/src/frontend/src/modules/Workspace/components/edges/ObservabilityEdgeCTA.spec.cy.tsx
tolumq: Thank you for the question.
It's neither an overkill, nor does it have anything to do with patterns.
It is in fact, the only way to prove that your action on line 14 is what led to the line 15 being true, else, one can argue that if you tested `have.been.called` on line 12, it might have too. It makes it easier to catch nasty bugs.

hivemq-edge/src/frontend/src/modules/Workspace/hooks/useGetPoliciesMatching.ts
tolumq: Another approach would be to put the `filter` first before the `map` if you're worried about the types.
However, a much cleaner solution can be something like this:

```
const policies = group.childrenNodeIds.reduce((acc: DataPolicy[], curr: string) => {
      const node = workspaceNodes.find((n: Node) => n.id = curr);
      if(node) {
        const policy = getPoliciesForAdapter(node);
        if (policy) {acc.push(...policy)}
      }
      return acc;
    }, [])
```
as it harnesses the benefits of `reduce`

vanch3d: Yes, much better indeed. Thx!

tolumq (approved):
LGTM

PR #360, opened 2024-04-08
fix(20433): Remove Schema Ref message from the list

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/20433/details/

AJV8, used to validate the JSONSchema form, generates a "no schema with key or ref" error which is displayed at the first load of an adapter. It has not been possible to fix the error. 

The PR filters the message out of the list to be displayed on the summary. 

Note that the generated error message has no other identifier in the payload, except the "stack" attribute.


sfrehse:
The error object I got:

```
[
    {
        "stack": "no schema with key or ref \"https://json-schema.org/draft/2020-12/schema\""
    }
]
```

h2xd (approved):
LGTM! 🐊

PR #362, opened 2024-04-09
refactor(21032): Add status of terminal transitions to the node

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/21032/details/

The PR improves the rendering of a `behaviour policy` by highlighting, when appropriate, the terminal status of a transition. 

The highlight is materialised by a `success` or `failure` handle of the `Transition` node and, by doing so, aligns with a similar status materialised for the `data policy` 

The PR also adds `success` disconnected transitions that are missing from the generated model.

### Out-of-scope
- The rendering of the `transition` node only shows the transition name, not the end state of the FSM. It certainly creates ambiguity, since the transition names are not unique. Guards, indicating further conditions to the application of the transition, will be added in a future ticket, ensuring unicity of names.

### Before
screenshot-localhost_3000-2024 04 09-14_36_31

### After 
screenshot-localhost_3000-2024 04 09-14_36_10


h2xd (approved):
LGTM! 🐊

PR #364, opened 2024-04-10
Renovate: Split Dependency Updates between gradle and npm

DC2-DanielKrueger:
**Motivation**

Resolves #<issue>

**Changes**

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -9,5 +9,208 @@
     ],
     "useBaseBranchConfig": "merge",
     "branchPrefix": "renovate/hivemq-edge/",
-    "prConcurrentLimit": 20
+    "prConcurrentLimit": 20,
+    "packageRules": [
+        {
+            "matchPackagePatterns": [
+                "^io\\.confluent[.:]",
+                "^org\\.apache\\.kafka[.:]"
+            ],
+            "registryUrls": [
+                "https://packages.confluent.io/maven"
+            ]
+        },
+        {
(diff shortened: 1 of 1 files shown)


renovate.json5
```diff
@@ -9,5 +9,208 @@
...
+            "matchPackagePatterns": [
+                "^com\\.github\\.everit-org\\.json-schema[.:]"
+            ],
+            "registryUrls": [
+                "https://jitpack.io"
+            ]
+        },
+        {
+            "groupName": "all patch dependencies",
+            "groupSlug": "all-patch",
```
vanch3d, line 49: D o we want to keep the all-patches ?

renovate.json5
```diff
@@ -9,5 +9,208 @@
...
+        // GitHub Actions minor & patch
+        {
+            "matchManagers": [
+                "github-actions"
+            ],
+            "groupName": "all non-major GitHub Actions dependencies",
+            "groupSlug": "github-actions-non-major",
+            "matchUpdateTypes": [
+                "minor",
+                "patch"
```
vanch3d, line 170: Same here. Is minor not enough?

vanch3d (approved):
All good to me. Non-blocking questions about keeping patches in the renovate groups

renovate.json5
DC2-DanielKrueger: i split them up, I think we still want the patch dependencies that can be applied automotically

PR #370, opened 2024-04-11
refactor(19824): Fix the UX flow in the designer toolbox

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/19824/details/

The PR improves the user experience of the Designer toolbar, making the create/check/publish flow much easier to understand and use.

- the selection of a policy node is now more resilient to change and navigation within the Designer. The rules remain the same:
  - select a data or behaviour policy node to activate the "check" button
  - deselect the node or select a non-policy node and the "check" button will be disabled
- on a successful validity check, the alert message contains a link to the publish stepper. Manual navigation has been preferred to automated as it will give time to the user to check the message. 
- After a publish, successful or not, the check process will be reset: the stepper will navigate back to the "check" step, the message will be cleared but the policy node will still be selected. 

The icons have been redesigned and the rendering of the "expand" icon for the toolbox button improved. 

The PR also relocates the CTA button for "new draft" to its natural location on the top of the page. This is to abide by the design style but also to remove the ambiguity of the CTA splitting the table body from the navigation panel,

### Before

screenshot-localhost_3000-2024 04 11-17_20_30

### After 

screenshot-localhost_3000-2024 04 11-17_19_48

screenshot-localhost_3000-2024 04 12-09_52_44


oli-hivemq (approved):
Based on reviewing the code changes, the description of the PR and the before/after screenshots, it looks good to me 🦣.
The new icons look better 👍

Note that I haven't run and tested manually. I would have liked to, hopefully we can figure out how to provide a testing environment per PR without having to run the backend locally.

PR #376, opened 2024-04-15
fix(21009): Add navigation to the error node's panel

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/21009/details/

The PR fixes the missing navigation to the configuration panel of a node flagged as an error in the validation check. 

Every node in a verified policy with an error in the report provides access to both its location in the Designer and to its configuration side panel. 

### After
screenshot-localhost_3000-2024 04 15-10_56_44


hivemq-edge/src/frontend/src/extensions/datahub/components/controls/DesignerToolbox.tsx
```diff
@@ -138,7 +142,15 @@ const DesignerToolbox: FC = () => {
                     <>
                       <Box pt={5} h="100%">
                         {activeStep === 0 && <ToolboxNodes />}
-                        {activeStep === 1 && <ToolboxDryRun onActiveStep={onActiveStep} />}
+                        {activeStep === 1 && (
```
oli-hivemq, line 145: suggestion: Instead of using _magic numbers_ for the steps, using constants named accordingly to describe what each step is would make the code easier to follow.

Plus, if steps were to change, it would be less prone to errors.

hivemq-edge/src/frontend/src/extensions/datahub/components/controls/DesignerToolbox.tsx
```diff
@@ -138,7 +142,15 @@ const DesignerToolbox: FC = () => {
...
+                          <ToolboxDryRun
+                            onActiveStep={onActiveStep}
+                            onShowNode={(node) => fitView({ nodes: [node], padding: 3, duration: 800 })}
```
oli-hivemq, line 148: suggestion: Assuming that as _Edge_ grows, more time-based interactions and timeouts will keep appearing. For consistency, I suggest defining time constants that you'll be able to reuse.

vanch3d:
I've made changes according to your suggestions. 

> suggestion: Assuming that as Edge grows, more time-based interactions and timeouts will keep appearing. For consistency, I suggest defining time constants that you'll be able to reuse

I've created a constant for it, in a placeholder that should be used to similar values. But I also think they should be part of the theme (and design system), as part of the consistent UX we are building across our product. ChakraUI already provides a good framework through its theme tokens for defining such constants, see https://v2.chakra-ui.com/docs/styled-system/css-variables

PR #381, opened 2024-04-22
Fix: Define versions explicitly to avoid compile issue

DC2-DanielKrueger:
**Motivation**

Resolves #<issue>

**Changes**

### DIFF
--- a/hivemq-edge/src/frontend/.prettierignore
+++ b/hivemq-edge/src/frontend/.prettierignore
@@ -7,4 +7,3 @@ coverage/
 public/
 src/api/__generated__
 .gradle/
-package.json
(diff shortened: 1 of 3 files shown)


hivemq-edge/src/frontend/.prettierignore
```diff
@@ -7,3 +7,4 @@ coverage/
 public/
 src/api/__generated__
 .gradle/
+package.json
```
vanch3d: Don't exclude `package.json` from prettier. Instead, make sure `prettier --write .` is run at every file save (or manually once before committing)

vanch3d (approved):

PR #383, opened 2024-04-22
fix(20360): Fix the serialisation of multiple scripts

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/20360/details/

The PR fixes the publishing process of multiple scripts in a transformation node. The current process was only saving the first script connected to the `Operation` node. 

The PR also slighlty improves the rendering of the successful report on a `dry run` and the default layout for (multiple) schemas and scripts.

### After
screenshot-localhost_3000-2024 04 22-16_41_51

### DIFF
--- a/hivemq-edge/src/frontend/src/extensions/datahub/designer/operation/OperationNode.utils.ts
+++ b/hivemq-edge/src/frontend/src/extensions/datahub/designer/operation/OperationNode.utils.ts
@@ -15,7 +15,12 @@ import {
   WorkspaceAction,
   WorkspaceState,
 } from '@datahub/types.ts'
-import { checkValidityJSScript, loadScripts } from '@datahub/designer/script/FunctionNode.utils.ts'
+import {
+  checkValidityJSScript,
+  formatScriptName,
+  loadScripts,
+  parseScriptName,
+} from '@datahub/designer/script/FunctionNode.utils.ts'
 import { checkValiditySchema, loadSchema } from '@datahub/designer/schema/SchemaNode.utils.ts'
 import { PolicyCheckErrors } from '@datahub/designer/validation.errors.ts'
 import { isFunctionNodeType, isSchemaNodeType } from '@datahub/utils/node.utils.ts'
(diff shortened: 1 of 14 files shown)


hivemq-edge/src/frontend/src/extensions/datahub/designer/operation/OperationNode.utils.ts
```diff
@@ -94,9 +95,23 @@ export function checkValidityTransformFunction(
     ]
   }
 
-  // TODO[19497] This should not have to happen on the client side!
-  const formattedScriptName = (functionNode: Node<FunctionData>): string => {
-    return `fn:${functionNode.data.name}:latest`
+  const { transform } = operationNode.data.formData as unknown as OperationData.DataHubTransformType
+  const defaultOrder = transform.length ? transform : scriptNodes.map((e) => e.data?.id)
```
h2xd, line 99: suggestion(non-blocking): let's use `scriptNode` instead of `e` to define what the iteratee is

```suggestion
  const defaultOrder = transform.length ? transform : scriptNodes.map((scriptNode) => scriptNode.data?.id)
```

hivemq-edge/src/frontend/src/extensions/datahub/designer/operation/OperationNode.utils.ts
```diff
@@ -94,9 +95,23 @@ export function checkValidityTransformFunction(
...
+
+  const allTransformScripts: DryRunResults<PolicyOperation, Script>[] = []
+  for (const scriptId of defaultOrder) {
+    const script = scriptNodes.find((e) => e.data?.id === scriptId)
```
h2xd, line 103: suggestion(non-blocking): let's use `scriptNode` instead of `e` to define what the iteratee is

```suggestion
    const script = scriptNodes.find((scriptNode) => scriptNode.data?.id === scriptId)
```

h2xd (approved):
LGTM! 🐊 

Found just 2 small little things

PR #385, opened 2024-04-23
refactor(20321): Refactoring the policy loading 

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/20321/details/

The PR refactors how the policy payload is transformed into a designer graph. 

The previous process was creating the graph elements (nodes and edges) from the payload and inserting them iteratively into the Design store. This created a re-render race condition, with parent nodes not yet inserted into the graph when their connecting edges were added. The error message at the first loading was the visual result of the faulty operation. 

The new process iteratively creates ALL the graph elements BEFORE inserting them in two distinct operations: nodes first then edges. 

The PR also fixes a bug with the behaviour policy, in which pipelines were connected to every transition created, regardless of the real dependency. 

### Before 
screenshot-localhost_3000-2024 04 23-13_03_01

screenshot-localhost_3000-2024 04 23-13_03_27

### After 
 
screenshot-localhost_3000-2024 04 23-13_02_26

### DIFF
--- a/hivemq-edge/src/frontend/src/extensions/datahub/components/pages/PolicyEditorLoader.tsx
+++ b/hivemq-edge/src/frontend/src/extensions/datahub/components/pages/PolicyEditorLoader.tsx
@@ -64,7 +64,9 @@ export const DataPolicyLoader: FC<PolicyLoaderProps> = ({ policyId }) => {
       )
 
       store.onNodesChange(nodeChanges)
-      edgeConnects.map((connection) => store.onConnect(connection))
+      for (const connection of edgeConnects) {
+        store.onConnect(connection)
+      }
       store.setStatus(DesignerStatus.LOADED, { name: dataPolicy.id })
     } catch (error) {
       let message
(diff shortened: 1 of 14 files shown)


antpaw (changes requested):
there are a couple of `as NodeAddChange` and `as Connection` statements. This is an unsafe way of casting to an object. If this happens to variable declaration or return types you can use `satisfies NodeAddChange` instead. For return types you could also declare the return type of the function and remove the casting

```
// example
export const loadBehaviorPolicy = (behaviorPolicy: BehaviorPolicy): NodeAddChange => {
  // ... stuff
  return { item: behaviorPolicyNode, type: 'add' }
}

export const loadClientFilter = (behaviorPolicy: BehaviorPolicy, behaviorPolicyNode: Node<BehaviorPolicyData>): (NodeAddChange | Connection)[] => {
  /// ..stuff
  return [
    { item: topicNode, type: 'add' },
    { source: topicNode.id, target: behaviorPolicyNode.id, sourceHandle: null, targetHandle: null },
  ]
}
```

antpaw (approved):

hivemq-edge/src/frontend/src/extensions/datahub/components/pages/PolicyEditorLoader.tsx
```diff
@@ -44,12 +45,26 @@ export const DataPolicyLoader: FC<PolicyLoaderProps> = ({ policyId }) => {
...
+        const nodeChange = element as NodeAddChange
+        if (nodeChange.item && !allIds.includes(nodeChange.item.id)) acc.push(nodeChange)
+        return acc
+      }, [])
+      const edgeConnects = allArtefactsLoaded.filter(
+        (element): element is Connection => !!(element as Connection).source
+      )
+
+      store.onNodesChange(nodeChanges)
+      edgeConnects.map((connection) => store.onConnect(connection))
```
antpaw, line 67: why is this a map, i dont see anyone collecting it

PR #388, opened 2024-04-24
feat(21481): Add node duplication and copy-and-paste support

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/21481/details/

This PR improves the UX of the `Data Hub Designer` by allowing users to copy nodes from the graph and duplicate them as new elements. 

The operation is an internal "copy-and-paste" of the selection in the designer:
- one or many nodes can be selected manually from the graph  
- on "copy", the subgraph containing the selected nodes (i.e. including all the edges that have both start and end nodes in the selection) is duplicated (i.e. new id, new shifted position but the same configuration of the nodes
- on "paste", the new nodes are added to the graph but not connected to existing nodes. 
- subsequent "paste" will duplicate the selected nodes in a shifting pattern on the graph, allowing multiple duplications
- "copy" is bound to the usual "META + C" and "paste" to "META + V". "ESC" will clear the selection and the existing copied content

A small feedback widget has been added to the bottom centre of the Designer. It contains a "copy" indicator, with the number of nodes copied.

Note that copied content is not shared in the clipboard and its scope is limited to the current draft. Leaving the Designer will wipe the buffer. The copy-and-paste feature is also only available on DRAFT policies.

The PR also adds a "cheat sheet" with all the keyboard shortcuts supported by `React Flow` or the `Data Hub Designer`. The modal is available from the canvas toolbox.

### After
screenshot-localhost_3000-2024 04 25-09_30_37

screenshot-localhost_3000-2024 04 25-09_30_51

screenshot-localhost_3000-2024 04 25-09_31_05

### DIFF
--- a/hivemq-edge/src/frontend/package.json
+++ b/hivemq-edge/src/frontend/package.json
@@ -75,6 +75,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-hook-form": "^7.43.9",
+    "react-hotkeys-hook": "^4.5.0",
     "react-i18next": "^12.3.0",
     "react-icons": "^5.0.1",
     "react-router-dom": "^6.11.2",
--- a/hivemq-edge/src/frontend/pnpm-lock.yaml
+++ b/hivemq-edge/src/frontend/pnpm-lock.yaml
@@ -1,9 +1,5 @@
 lockfileVersion: '6.0'
 
-settings:
-  autoInstallPeers: true
-  excludeLinksFromLockfile: false
-
 dependencies:
   '@chakra-ui/anatomy':
     specifier: ^2.1.1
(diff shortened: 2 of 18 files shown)


antpaw:
I would suggest to change the name of the "Meta" key to something dynamic like this

```
  const shortcut = useMemo(() => {
    const os = window.navigator.platform;
    if (os.startsWith('Mac')) {
      return '⌘K';
    } else {
      return '^K';
    }
  }, [])
  ```

vanch3d:
> I would suggest to change the name of the "Meta" key to something dynamic like this
> 
> ```
>   const shortcut = useMemo(() => {
>     const os = window.navigator.platform;
>     if (os.startsWith('Mac')) {
>       return '⌘K';
>     } else {
>       return '^K';
>     }
>   }, [])
> ```

I cannot use the symbols, as I have multiple keys and non-key binding. So I'm using the official modifier name (`Command` and `Ctrl`)

screenshot-localhost_3000-2024 04 29-09_07_56

PR #390, opened 2024-04-25
feat(21481): Create nodes from dropping an edge from a source

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/21481/details/

This PR enables the creation of nodes directly from the dropping of an edge created from a source node onto the canvas. 

The action is based on the strong typing of the connection between different nodes: it is possible, by starting an edge from the handle of a node to know what target node is expected, therefore unambiguously creating it when the user drops the edge

Notes
- The drop of the edge MUST happen on the canvas itself, away from other nodes and edges
- The process is bidirectional, either from the "source" or "target" handle of any node; the direction of the created edge will be properly defined 
- A few nodes (SCHEMA, specific instances of OPERATION) are ambiguous and therefore don't support the create-on-drop.

### After

https://www.loom.com/share/31dc58eb39f54f2b91b58c975567bbd0?sid=73ffb344-0c51-46ff-86f6-0abf84fbad38

### DIFF
--- a/hivemq-edge/src/frontend/src/extensions/datahub/components/pages/PolicyEditor.tsx
+++ b/hivemq-edge/src/frontend/src/extensions/datahub/components/pages/PolicyEditor.tsx
@@ -151,6 +151,7 @@ const PolicyEditor: FC = () => {
           onConnectStart={onConnectStart}
           onConnectEnd={onConnectEnd}
           onConnect={onConnect}
+          connectionLineComponent={ConnectionLine}
           onInit={setReactFlowInstance}
           fitView
           snapToGrid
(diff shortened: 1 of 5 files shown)


hivemq-edge/src/frontend/src/extensions/datahub/components/pages/PolicyEditor.tsx
```diff
@@ -59,6 +78,60 @@ const PolicyEditor: FC = () => {
...
+
+      if (isTargetCanvas && edgeConnectStart.current && reactFlowInstance) {
+        const { type, handleId, nodeId } = edgeConnectStart.current
+
+        const droppedNode = getConnectedNodeFrom(type, handleId)
+        if (droppedNode) {
+          const id = getNodeId()
+          const newNode: Node = {
+            id,
+            position: reactFlowInstance?.screenToFlowPosition({
```
antpaw, line 105: the `?` is not needed here, you are checking for it on line 97

antpaw (approved):

PR #392, opened 2024-04-26
fix(21625): Prevent cycles in the Designer graphs

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/21625/details/

The PR prevents cycles from being created in the Designer when connecting nodes either to themselves or back to the beginning of a chain. 

This is the case exclusively of the `OPERATION` node, as being the only content that can be chained. 

### Before
screenshot-localhost_3000-2024 04 25-20_41_26

### After
screenshot-localhost_3000-2024 04 26-15_43_40

### DIFF
--- a/hivemq-edge/src/frontend/src/extensions/datahub/utils/node.utils.ts
+++ b/hivemq-edge/src/frontend/src/extensions/datahub/utils/node.utils.ts
@@ -91,6 +91,7 @@ export const isValidPolicyConnection = (connection: Connection, nodes: Node[], e
       if (outgoer.id === connection.source) return true
       if (hasCycle(outgoer, visited)) return true
     }
+    return false
   }
 
   if (!source || !destination) {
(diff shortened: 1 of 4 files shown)


hivemq-edge/src/frontend/src/extensions/datahub/utils/node.utils.ts
```diff
@@ -78,18 +78,41 @@ export const validConnections: ConnectionValidity = {
...
-  if (!source) {
+  const hasCycle = (node: Node, visited = new Set()) => {
+    if (visited.has(node.id)) return false
+
+    visited.add(node.id)
+
+    for (const outgoer of getOutgoers(node, nodes, edges)) {
+      if (outgoer.id === connection.source) return true
+      if (hasCycle(outgoer, visited)) return true
+    }
```
antpaw, line 93: i would add `return false` at the end

antpaw (approved):

hivemq-edge/src/frontend/src/extensions/datahub/utils/node.utils.ts
vanch3d: Oh nicely spotted! 
I wonder if I should consider using strict true/false for the tests too

PR #403, opened 2024-04-30
feat(19255): Add safeguards for deleting nodes or edges on the Designer

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/19255/details/

The PR remove the uncontrolled deletion of nodes and edges on the `Rect Flow` canvas by adding user confirmation and safeguards. 

The PR also fixes https://hivemq.kanbanize.com/ctrl_board/57/cards/20301/details/

### Design
- Both `DELETE` and `BACKSPACE` keys are now supported 
- Apply deletion on individually selected nodes or edges or on grouped selections. In both cases, all orphan edges will also be deleted 
- If deletion is possible, a confirmation modal will require user intervention before being applied. There is no further feedback once the deletion is done (elements are immediately removed) 
- If the deletion is not possible (see below), an error toast will provide reasons for the error
- The deletion operation cannot be reversed 

The deletion verifies the validity of the process before being applied. The following rules are in place:
- If the `Designer` is in `READ-ONLY` mode (triggered by "View" CTA), no deletion is possible
- If the `Designer` is in `DRAFT` mode (triggered by "Create new" CTA), every element on the canvas can be deleted
- If the `Designer` is in `LOADED` (triggered by "Edit" CTA) or `MODIFIED` (triggered by a change in the policy), deletions are authorised except for the following reasons:
  - The `Data Policy` or `Behaviour Policy` nodes cannot be deleted once they have been published
  - The `Topic Filter` node cannot be deleted 
  - The connection between a `Topic Filter` and a `Data Policy` cannot be deleted

### Before
screenshot-localhost_3000-2024 04 30-17_56_07

### After
screenshot-localhost_3000-2024 04 30-17_54_49

screenshot-localhost_3000-2024 04 30-17_55_12


hivemq-edge/src/frontend/src/modules/Bridges/components/panels/SubscriptionsPanel.spec.cy.tsx
```diff
@@ -65,13 +65,16 @@ describe('SubscriptionsPanel', () => {
     cy.percySnapshot('Component: SubscriptionsPanel')
   })
 
-  it('should initialise with OpenAPI defaults', () => {
+  it.only('should initialise with OpenAPI defaults', () => {
```
h2xd, line 68: question(non-blocking): guess all tests should run but only this once will be executed once merged

```suggestion
  it('should initialise with OpenAPI defaults', () => {
```

h2xd (approved):
LGTM! 🐊

PR #407, opened 2024-05-01
refactor(renovate): Upgrade MSW + Cypress + Vitest

vanch3d:
- MSW upgraded to v2 & intercepts refactored
- Cypress upograded to v13


oli-hivemq (commented):
The changes look good to me.
Are the build & e2e tests run as part of the CICD process on github? Or would it be safer to check locally?
And if we're choosing to run locally (which I'm happy to try), is there a backend I could point the frontend to so I don't need to run it locally (given that I haven't set it up)?

vanch3d:
> The changes look good to me. Are the build & e2e tests run as part of the CICD process on github? Or would it be safer to check locally? And if we're choosing to run locally (which I'm happy to try), is there a backend I could point the frontend to so I don't need to run it locally (given that I haven't set it up)?

Yes they are all running on the CI pipeline, see the `.github/` actions on the top repo.

Re: backend, actually that's a good point: Cypress needs the frontend running to work (just `run dev`) but the backend is not needed since I'm intercepting all the requests (most of them at least). So even if you cannot run the frontend, you should be able to rn the test. 

`MSW` used to be set up for running the frontend when there was no API available. I need to try it again; with all the mocks, you could have a decent-enough environment locally.

oli-hivemq (approved):
LGTM 🦣

PR #408, opened 2024-05-02
Epic: concrete

DC2-DanielKrueger:
**Motivation**

Resolves #<issue>

**Changes**


hivemq-edge/src/main/java/com/hivemq/HiveMQEdgeBootstrap.java
```diff
@@ -0,0 +1,229 @@
...
+    private void awaitPersistenceStartup() {
+        Objects.requireNonNull(injector).persistences();
+        try {
+            persistenceStartup.finish();
+        } catch (InterruptedException e) {
+            throw new HiveMQEdgeStartupException(e);
+        }
+
+        // configService is always set in caller
+        assert configService != null;
```
codepitbull, line 132: Why isn't this asserted at the begin of the method?

hivemq-edge/src/main/java/com/hivemq/HiveMQEdgeBootstrap.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.Objects;
+
+public class HiveMQEdgeBootstrap {
+
+    private static final @NotNull Logger log = LoggerFactory.getLogger(HiveMQEdgeBootstrap.class);
+
+    private final @NotNull MetricRegistry metricRegistry;
+    private final @NotNull SystemInformation systemInformation;
+    private final @NotNull ModuleLoader moduleLoader;
+    private @Nullable ConfigurationService configService;
```
codepitbull, line 51: I am not too familiar with the structure of invocations but looking at this field and how it is set/accessed it might be accessed from multiple threads.
Since init is delayed this field should probably be volatile.

hivemq-edge/src/main/java/com/hivemq/extensions/core/CommercialModuleLoaderDiscovery.java
```diff
@@ -15,40 +15,73 @@
...
+                instance.generalBootstrap(generalBootstrapService);
+            }
+        } catch (Exception e) {
+            log.error("Error when bootstrapping general information", e);
+        }
+    }
+
+    public void persistenceBootstrap(final @NotNull PersistenceBootstrapService persistenceBootstrapService) {
+        try {
+            for (ModuleLoaderMain instance : instances) {
```
codepitbull, line 66: discoverModuleLoaderMainClasses can be triggered while iterating which would lead to a. ConcurrentModificationException.
Either iterate using an iterator or an index-based for takes care of that.

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterMetrics.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static com.hivemq.metrics.HiveMQMetrics.ADAPTERS_CURRENT;
+import static com.hivemq.metrics.HiveMQMetrics.PROTOCOL_ADAPTER_PREFIX;
+
+@Singleton
+public class ProtocolAdapterMetrics {
+
+    private final @NotNull MetricRegistry metricRegistry;
+    private final AtomicInteger currentAdapters = new AtomicInteger();
```
codepitbull, line 17: Why the explicit atomic?
Codahale has metricRegistry..counter() which provides the very same functionality.

codepitbull (changes requested):
A few minor comments.

hivemq-edge/src/main/java/com/hivemq/HiveMQEdgeBootstrap.java
DC2-DanielKrueger: yeah, good point. I streamline it with Preconditions

hivemq-edge/src/main/java/com/hivemq/HiveMQEdgeBootstrap.java
DC2-DanielKrueger: It should only be accessed by the main thread. Still making it volatile makes sense to be future-proof.

hivemq-edge/src/main/java/com/hivemq/extensions/core/CommercialModuleLoaderDiscovery.java
DC2-DanielKrueger: yeah if the method would be called a second time, that is true. It is supposed to be called once before bootstraping everything. So right now it is safe, however for the future you are right. If somebody calls it a second time and concurrently, we get that exception. Index based is not safe as the List could get shorter though (theoretically).

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterMetrics.java
DC2-DanielKrueger: we wanted it to be a Gauge to be consistent with the other metrics. This is why we added the AtomicInteger although a counter would do the exact same job.

codepitbull (approved):
LGTM

PR #423, opened 2024-05-13
feat(20786): update the openAPI stubs

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/20786/details/

The PR updates the stubs and schemas to the OpenAPI specs `2024.4`. It adds the missing hooks (FSM) and mockd payload. 

It also change the internal type for Validator type to the enum generated from the schema This is a potential breaking change (from `schema` to `SCHEMA` but backward compatibility has been looked at).


oli-hivemq (approved):
All checks have passed so LGTM 🦣

PR #430, opened 2024-05-21
fix(22182): Handle readonly and editable Designer

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/22182/details/

The PR refactors the way policies are handling their `read-only` status when loaded on the Designer. 

The User Flow for the Designer is now clearly based on 4 states for the content edited in the canvas:
- `LOADED` when an existing policy is loaded onto the Designer; this is a `read-only` status
- `MODIFIED` when a `LOADED` policy has the `read-only` flag turned off and its content is being modified by the user
- `DRAFT` when a new policy is created on the canvas, either from empty content or from a copied policy

When in `LOADED` mode, the Designer is in read-only and its various elements now reflect that status:
- The nodes/check/publish toolbar is disabled
- The nodes on the canvas can be dragged around but cannot be deleted or re-connected
- The configuration panel, opened by double-clicking on a node, is in read-only mode: the configuration can be seen but cannot be modified, the validation is not activated

The policy toolbar now contains a CTA to switch the 'read-only` mode off,  allowing users to edit a loaded policy. Draft content are not allowed to switch the status (since they are by default editable)

The `read-only` status of the Designer is indicated in the status bar at the middle bottom of the screen

### Out-of-scope
- The `read-only` and `editable` status of the Designer enables a proper contextualisation of the modification of policy. The modification process itself will be implemented in a subsequent ticket

### Before 
screenshot-localhost_3000-2024 05 21-16_16_53

### After
screenshot-localhost_3000-2024 05 21-16_19_33 (1)

screenshot-localhost_3000-2024 05 21-16_16_13


h2xd (approved):

h2xd:
Sorry pressed the wrong button. 🤦

hivemq-edge/src/frontend/src/components/rjsf/__internals/ChakraIconButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+      {...otherProps}
+      isDisabled={props.disabled}
+      icon={icon as ChakraIconButtonProps['icon']}
+    />
+  )
+}
+
+ChakraIconButton.displayName = 'ChakraIconButton'
+
+export default memo(ChakraIconButton) as typeof ChakraIconButton
```
h2xd, line 33: question(non-blocking): is the type casting correct would have expected `Memoized<ChakraIconButton>` or similar here

hivemq-edge/src/frontend/src/extensions/datahub/components/controls/DeleteListener.spec.cy.tsx
```diff
@@ -132,4 +134,25 @@ describe('DeleteListener', () => {
       cy.wrap(w[3]).should('contain.text', 0)
     })
   })
+
+  it.only('should not delete if the designer is readonly', () => {
```
h2xd, line 138: suggestion(blocking): watch out there is an `only` here

```suggestion
  it('should not delete if the designer is readonly', () => {
```

hivemq-edge/src/frontend/src/extensions/datahub/components/pages/PolicyEditor.tsx
```diff
@@ -39,11 +40,12 @@ const PolicyEditor: FC = () => {
   const { t } = useTranslation('datahub')
   const reactFlowWrapper = useRef(null)
   const [reactFlowInstance, setReactFlowInstance] = useState<ReactFlowInstance | null>(null)
-  const { nodes, edges, onNodesChange, onEdgesChange, onConnect, onAddNodes } = useDataHubDraftStore()
+  const { status, nodes, edges, onNodesChange, onEdgesChange, onConnect, onAddNodes } = useDataHubDraftStore()
   const edgeConnectStart = useRef<OnConnectStartParamsNode | undefined>(undefined)
-
   const nodeTypes = useMemo(() => CustomNodeTypes, [])
 
+  const isEditable = status !== DesignerStatus.LOADED
```
h2xd, line 47: question(non-blocking): should this one be computed via `useMemo`?

```suggestion
  const isEditable = useMemo(() => status !== DesignerStatus.LOADED, [status])
```

h2xd (approved):
LGTM! 🐊 
Added couple of questions and found one `only` in the test suite.
I've added an approved to unblock the process but highly suggest to remove the `only` statement.

hivemq-edge/src/frontend/src/components/rjsf/__internals/ChakraIconButton.tsx
vanch3d: Copy of the original file. Temporary hack, will get a PR in their repo to achieve my fix to the disabled state

PR #434, opened 2024-05-22
fix(22188): Prevent creation of dropped node when connecting

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/22188/details/

The PR prevents the creation of an extra node when dropping a connection too close to a connectable handle.

If the distance of the drop to the handle was to small, the nodes would be connected AND a new one create on the spot as well. 

Connecting exisitng nodes is now taking priority on a drop. The "connection radius" has been increase to materialise the preference and safeguard added to prevent the extra creation if a node is connected.


h2xd (approved):
LGTM! 🐊

PR #455, opened 2024-06-11
feat(22339): Remove hardcoded uiSchema from UI

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/22339/details/

The PR removes the hardcoded `uiSchema` defined for every adapter as a combined artefact in the UI. Instead, it now extracts the customisation out of the specification of individual adapters from the API.
A minimalist `uiSchema`, containing the internal requirements for the form, is maintained in the front end.

The PR also fixes the error in the `uiSchema` of some of the adapters (pure English strings, duplicated config, non-existing properties)

The PR also fixes the dynamic validation of `id`, see https://hivemq.kanbanize.com/ctrl_board/57/cards/22997/details/


h2xd (approved):
I've found nothing :D

LGTM! 🐊

PR #469, opened 2024-06-19
fix(23330): Only include selectable options in the browser

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/23330/details/

The PR corrects the logic of the tag browser, only allowing selectable options to be visible in the auto-complete options. 

The `selectable` flag is internal and received from the backend.

The PR also customises the error message when no options are visible, to identify the context (no options available, no selectable options available, no option matching filter)

The PR adds a "breadcrumb" representation of the path of an item in the tree. It could be used for rendering purposes, as an alternative to present a fully, non-accessible treeview

### Out-of-scope
- The application of the `selectable` flag might result in no options available, making the auto-complete component unworkable. An alternative UX, if non-discoverable configuration is still required, will be implemented in a different ticket

### Before
screenshot-localhost_3000-2024 06 18-13_13_23

screenshot-localhost_3000-2024 06 18-13_14_05

### After
screenshot-localhost_3000-2024 06 19-17_16_41

screenshot-localhost_3000-2024 06 19-17_16_59

### DIFF
--- a/hivemq-edge/src/frontend/src/components/rjsf/Widgets/AdapterTagSelect.spec.cy.tsx
+++ b/hivemq-edge/src/frontend/src/components/rjsf/Widgets/AdapterTagSelect.spec.cy.tsx
@@ -63,19 +63,19 @@ describe('AdapterTagSelect', () => {
     cy.get('#react-select-dataPoint-container').click()
     cy.get('#react-select-dataPoint-listbox').should('be.visible')
     cy.get('#react-select-dataPoint-listbox').find('[role="option"]').as('options')
-    cy.get('@options').should('have.length', 11)
-    cy.get('@options').eq(0).find('[data-testid="dataPoint-name"]').should('have.text', 'Object')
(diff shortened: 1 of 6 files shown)


oli-hivemq:
Your 2nd screenshot shows "None of the tabs are selectable" but the UI also shows that this is a required field. What does this mean for the end user? Can they continue or will they be blocked?
What's the user experience like?

vanch3d:
> Your 2nd screenshot shows "None of the tabs are selectable" but the UI also shows that this is a required field. What does this mean for the end user? Can they continue or will they be blocked?

It's a major issue we will be tackling in the next tickets. There is a contradiction:
- you cannot just add a nonsensical info, the adapter will generate an error at runtime. So auto-complete select makes sense
- you cannot have a select without options. It means the adapter doesn't accept any configuration and, being required cannot function

h2xd (approved):
LGTM! 🐊

PR #476, opened 2024-06-27
fix future handling

DC2-DanielKrueger:
https://hivemq.kanbanize.com/ctrl_board/57/cards/22351/details/

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImpl.java
@@ -150,7 +150,8 @@ public class ProtocolAdaptersResourceImpl extends AbstractApi implements Protoco
 
     @Override
     public @NotNull Response getAdapter(final @NotNull String adapterId) {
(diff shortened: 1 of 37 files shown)


hivemq-edge/src/main/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImpl.java
```diff
@@ -290,16 +321,27 @@ public int getDepth() {
...
+                switch (command.getCommand()) {
+                    case START:
+                        protocolAdapterManager.start(adapterId).get();
+                        break;
+                    case STOP:
+                        protocolAdapterManager.stop(adapterId).get();
+                        break;
+                    case RESTART:
+                        protocolAdapterManager.stop(adapterId)
+                                .thenRun(() -> protocolAdapterManager.start(adapterId))
```
sfrehse, line 334: What happens if the stop never ends?

DC2-DanielKrueger: I added timeouts to all adapter interactions. This way we should be safe.

sfrehse (approved):

PR #478, opened 2024-07-01
Add new AdapterPollingService

DC2-DanielKrueger:
https://hivemq.kanbanize.com/ctrl_board/57/cards/23793/details/   

**Motivation**

Resolves #<issue>

**Changes**


hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/polling/PollingTask.java
```diff
@@ -0,0 +1,210 @@
...
+                        if (ExceptionUtils.isInterruptedException(throwable)) {
+                            handleInterruptionException(throwable);
+                        } else {
+                            handleExceptionDuringPolling(throwable);
+                        }
+                    }
+                });
+            }
+        } catch (Throwable t) {
+            // the sampler shouldnt throw a exception, but better safe than sorry as we might to
```
hlohse, line 86:
```suggestion
            // the sampler shouldn't throw a exception, but better safe than sorry as we might to
```
...but I am still wondering what the comment wanted me to know at the end. 😄

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/polling/PollingTask.java
```diff
@@ -0,0 +1,210 @@
...
+            if (log.isInfoEnabled()) {
+                log.info("Sampler Encountered Error In Notification", samplerError);
+            }
+        }
+    }
+
+    private void reschedule(int errorCountTotal) {
+        final long delayInMillis = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - nanosOfLastPolling);
+        // a negative delay means that the last polling attempt took longer to be processed than the specified delay between polls
+        if (delayInMillis < 0) {
```
hlohse, line 166: I don't see how the condition could every be true: `System.nanoTime()` will always be >= `nanosOfLastPolling`. You probably wanted to compare it to `sampler.getPeriod()`?

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/polling/PollingTask.java
```diff
@@ -0,0 +1,210 @@
...
+                            sampler.getAdapterId(),
+                            sampler.getPeriod()))
+                    .withSeverity(Event.SEVERITY.WARN)
+                    .fire();
+        }
+
+        long nonNegativeDelay = Math.max(0, delayInMillis);
+
+        if (errorCountTotal == 0) {
+            schedule(nonNegativeDelay);
```
hlohse, line 183: Don't you want to consider `sampler.getPeriod()` here? Otherwise, if the polling returns quickly enough, there would be barely any delay between polls.
Looks like `ProtocolAdapterPollingSampler#getPeriod()` is not respected here in general, seems accidental to me?

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/polling/PollingTask.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    @Override
+    public void run() {
+        try {
+            nanosOfLastPolling = System.nanoTime();
+
+            // synchronized with stop so that the change of the executor future field and the setting of the continueScheduling
+            // has exclusive access by either of the two methods.
+            synchronized (this) {
```
hlohse, line 66: Redundant/obsolete?

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/polling/PollingTask.java
```diff
@@ -0,0 +1,210 @@
...
+                    }
+                });
+            }
+        } catch (Throwable t) {
+            // the sampler shouldnt throw a exception, but better safe than sorry as we might to
+            handleExceptionDuringPolling(t);
+        }
+    }
+
+    public synchronized void stopScheduling() {
```
hlohse, line 91: Redundant/obsolete?
```suggestion
    public void stopScheduling() {
```

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/polling/PollingTask.java
```diff
@@ -0,0 +1,210 @@
...
+        continueScheduling.set(false);
```
hlohse, line 92: Not sure if possible, but it would be nice to speed up a shutdown if we could interrupt the thread here, and not reschedule it. But I am not sure if we get access to the thread.

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/polling/ProtocolAdapterPollingServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(ProtocolAdapterPollingServiceImpl.class);
+
+    private final @NotNull ScheduledExecutorService scheduledExecutorService;
+    private final @NotNull EventService eventService;
+    private final @NotNull Map<ProtocolAdapterPollingSampler, PollingTask> samplerToTask = new ConcurrentHashMap<>();
+
+    @Inject
+    public ProtocolAdapterPollingServiceImpl(
+            final @NotNull ScheduledExecutorService scheduledExecutorService,
```
hlohse, line 49: I have trouble figuring out which instance is created and injected here: How many threads does this have?

DC2-DanielKrueger: yeah, I used the existing one that was already used before

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/polling/PollingTask.java
DC2-DanielKrueger: that is the problem. Before we interrupted the threads of the service that were waiting on the future. However the thread that really did the work, was never touched as this thread could be any thread. It could be a thread started by the adapter, a fork-join-pool thread or anything basically. We do not need to interrupt as the tasks should be fast as we do not wait for the future anymore in the loop. Unless an adapter blocks in the poll() method, which no adapter should do, but use the async api we offer.

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/polling/PollingTask.java
DC2-DanielKrueger: the delay should include the period time already. Likely I made a mistake here.

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/polling/PollingTask.java
DC2-DanielKrueger: yeah that is missing for sure

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/polling/PollingTask.java
DC2-DanielKrueger: yeah I wanted to return/expose the future of the sampler.execute in stop(), but it changes frequently and is often null after the execution and before the next execution, so this seemed useless to me as stop basically kills the next poll attempt anyway, which seems semantically good enough.

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/polling/PollingTask.java
DC2-DanielKrueger: oh yeah, I fixed the comment. xD

hlohse (approved):

PR #483, opened 2024-07-09
fix(23881): Add fallback editor when Monaco is not loading

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/23881/details/

The Monaco Editor (https://github.com/suren-atoyan/monaco-react) relies on an external CDN to load all its components. When Edge is used "offline", the loading obviously fails. However, the visual feedback only shows a permanent "Loading ..." text, without any error message (in the dev console).

The PR provides a short-term solution for the problem by falling back on a standard `textarea` input.

### Design
- The fallback is created from the `textarea` widget provided by the React JSONSchema Form library, ensuring seamless integration
- A warning about the lack of syntax highlighting (and validation) is added below the form.
- The "loading" state of the Monaco Editor has been changed to the Edge's common spinner.
- The Monaco Editor doesn't seem to offer access to the loading errors. The fallback widget is therefore displayed FIRST while the editor is loading. Upon success, It will be replaced. This might result in a quick flicker 

### Out-of-scope
- An alternative to Monaco Editor, or at least an offline version of it, will be implemented in another ticket

### DIFF
--- a/hivemq-edge/src/frontend/src/extensions/datahub/components/forms/CodeEditor.tsx
+++ b/hivemq-edge/src/frontend/src/extensions/datahub/components/forms/CodeEditor.tsx
@@ -1,14 +1,39 @@
-import { useMemo } from 'react'
+import { useEffect, useMemo, useState } from 'react'
 import { labelValue, WidgetProps } from '@rjsf/utils'
-import { Editor } from '@monaco-editor/react'
-import { FormControl, FormLabel, VStack } from '@chakra-ui/react'
+import { Editor, useMonaco } from '@monaco-editor/react'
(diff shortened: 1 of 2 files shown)


oli-hivemq (approved):
I would have loved to see what it looks like with a screenshot.
But the description of the ticket & the code are clear enough.
So it looks good to go in my opinion 🦣

vanch3d:
> I would have loved to see what it looks like with a screenshot. But the description of the ticket & the code are clear enough. So it looks good to go in my opinion 🦣

I forgot to add my usual screenshots. Updated

PR #485, opened 2024-07-10
fix(23934): Remove the "seletable" filter from the auto-complete selector

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/23934/details/

This is a quick fix to the previous PR, which allows EVERY node discovered on the adapter to be selectable in the UI

### Design
- We have a significant ambiguity as to what makes a discovered node `selectable` in general
- Specifically, in `OPC-UA` the existing rule based on `VALUE` or `OBJECT` doesn't seem to be working
- The example given by one of our customers shows expected nodes that are `OBJECT` but `not selectable`
- We cannot make an informed decision of this treatment, based on our fake data. We need real-life expertise with real-life OPCA-UA devices deeding Edge
- The fix makes EVERY node selectable
  - users are not facing an empty list if their usable nodes are "unselectable"
  - if a wrong node is selected, the adapter will certainly fail and a relevant error message generated in the `event log`
  - the onus will be on the users to fix their mistake

### Before 
screenshot-localhost_3000-2024 07 10-12_01_52

For reference, this is the OPC-UA dummy :
Screenshot 2024-07-10 at 12 02 25

### After
screenshot-localhost_3000-2024 07 10-12_03_46

Note that it will affect EVERY adapter, including MODBUS, where the folder node will be displayed

screenshot-localhost_3000-2024 07 10-12_04_46

### DIFF
--- a/hivemq-edge/src/frontend/src/components/rjsf/Widgets/AdapterTagSelect.spec.cy.tsx
+++ b/hivemq-edge/src/frontend/src/components/rjsf/Widgets/AdapterTagSelect.spec.cy.tsx
@@ -63,18 +63,18 @@ describe('AdapterTagSelect', () => {
     cy.get('#react-select-dataPoint-container').click()
     cy.get('#react-select-dataPoint-listbox').should('be.visible')
     cy.get('#react-select-dataPoint-listbox').find('[role="option"]').as('options')
-    cy.get('@options').should('have.length', 8)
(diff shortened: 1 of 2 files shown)


DC2-DanielKrueger (approved):

h2xd (approved):

vanch3d:
The feature has been discontinued due to problems with the discovery process itself. 
This "fix" is now obsolescent

PR #493, opened 2024-07-15
fix(24143): Add configuration checks to the dry run validation of policies

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/24143/details/

The PR fixes a major gap from the validation run, in which connectivity between nodes on the Editor was checked bvut not their internal configuration. This often resulted is a validation returnin a green flag to publish, with the publish then failing for a mis-configured element. 

The PR also fixes a bug with the status update of nodes

### Out-of-order
- The errors are gathers according to the walk and might include multiple errors on the same node. A futher ticket will combine errors by nodes

### After


hivemq-edge/src/frontend/src/extensions/datahub/utils/node.utils.ts
```diff
@@ -376,3 +383,31 @@ export const renderResourceName = (
...
+    isValid,
+    errors: validate.errors,
+  }
+}
+
+export const checkValidityConfigurations = (allNodes: Node[]) => {
+  const allConfigurations: DryRunResults<unknown>[] = []
+  for (const node of allNodes) {
+    const nodeConfiguration = validateNode(node)
+    if (!nodeConfiguration.isValid) {
```
tolumq, line 402: to improve readability, you could change this to:
```
const invalidNodesWithErrors = !nodeConfiguration.isValid && nodeConfiguration?.errors?.length;

if(invalidNodesWithErrors) {
  for(const error of nodeConfiguration.errors) {
   ....
  }
}
```

tolumq (approved):
LGTM, left a suggestion

PR #494, opened 2024-07-15
chore(CYPRESS): Enable videos and verbose output

vanch3d:
To track the flaky component tests


oli-hivemq (approved):
Looks like the target branch for this should be fix/24143/dryrun-validation-config given that it seems to contain those changes.
Approving anyway as the other PR is approved already

PR #495, opened 2024-07-16
feat(24140): Add a context toolbar to the nodes on the Designer

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/24140/details/

The PR adds a conditional toolbar to every node on the Designer, containing buttons for the main tasks associated with nodes: edit (open the side panel), copy and delete

### Design
- The toolbar is conditional to a node being selected. 
- The toolbar is associated with the node (top)
- If multiple nodes are selected (like for copy), no toolbar is displayed
- The toolbar's size changes based on the zoom factor of the canvas, maintaining a consistent visual

### Out-of-scope
- Multiple selections might benefit from a toolbar, associated with the first node for example, and applying the operation to every node. This might be exploited in a subsequent ticket
- Dispatching the COPY and DELETE hit keys trigger an uncaught error with ReactFlow. This will be explored after eventual upgrade to v12

### Before
screenshot-localhost_3000-2024 07 16-13_44_40

### After
screenshot-localhost_3000-2024 07 16-13_44_20

### DIFF
--- a/hivemq-edge/src/frontend/src/components/Chakra/ContextMenu.tsx
+++ b/hivemq-edge/src/frontend/src/components/Chakra/ContextMenu.tsx
@@ -1,0 +1,75 @@
+import { MutableRefObject, useCallback, useEffect, useRef, useState } from 'react'
+import { useEventListener, Portal, Menu, MenuButton, PortalProps, MenuButtonProps, MenuProps } from '@chakra-ui/react'
+
+export interface ContextMenuProps<T extends HTMLElement> {
+  renderMenu: () => JSX.Element | null
+  children: (ref: MutableRefObject<T | null>) => JSX.Element | null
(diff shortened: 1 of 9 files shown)


hivemq-edge/src/frontend/src/components/Chakra/ContextMenu.tsx
```diff
@@ -0,0 +1,79 @@
...
+        setIsRendered(true)
+        setTimeout(() => {
+          setIsDeferredOpen(true)
+        })
+      })
+    } else {
+      setIsDeferredOpen(false)
+      const timeout = setTimeout(() => {
+        setIsRendered(isOpen)
+      }, 1000)
```
oli-hivemq, line 35: My only suggestion would be to define constant variables somewhere for timeouts. They could even become user preferences one day.

oli-hivemq (approved):
LGTM 🦣

PR #496, opened 2024-07-17
fix(24361): Fix bug with deprecated error

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/24361/details/

The PR removes a deprecated error throw. The case is dealt with by supporting an alternative field.

### Out-of-scope
- Since the discovery API requires an adapter `id` to work, it cannot be used when creating a new adapter. Without the discovery, there is no possible auto-complete. In such a case, the widget reverts to the default, uncontrolled, input. The issue was raised on the main PR.


oli-hivemq (approved):
LGTM 🦣

PR #498, opened 2024-07-18
refactor(24369): Turn discovery browser off

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/24369/details/

The PR turns the custom data point browser off and replace it with the standard text input. It will affect all adapters with the `DISCOVERY` feature and `discovery:tagBrowser` mapping in the `UiSchema`

### Before
screenshot-localhost_3000-2024 07 18-10_26_40

### After
screenshot-localhost_3000-2024 07 18-10_28_35


oli-hivemq (approved):
I was wondering why to turn the discovery off, but the ticket explains the reasons.

LGTM 🦣

PR #499, opened 2024-07-18
fix(24363): Fix flaky accessibility tests

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/24363/details/

The PR makes some changes to the Cypress and Axe configuration in order to more easily identify flaky tests

- Cypress and Axe have been updsate, to ensure false positives are up to date
- a more verbose violation report has been added, which should help to identify violated rules and faulty elements

A few false positive have been excluded


oli-hivemq (approved):
LGTM 🦣

But is there a plan for re-enabling color contrast tests at some point? Do you know how to address this issue?

vanch3d:
> LGTM 🦣
> 
> But is there a plan for re-enabling color contrast tests at some point? Do you know how to address this issue?

There is no real plan. These are genuinely flaky tests, as they occur too randomly to be anything else.
It's a pity to have a blanket exception but I'm always checking without the exceptions when I'm on the same test suite.
A better solution might take time

PR #519, opened 2024-08-06
feat(24592): Add the websocket configuration to the UI

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/24592/details/

The PR adds a new tab to the `Bridge` configuration panel, allowing users to enable websockets and their options

### Design
- Both `serverpath` and `subprotocol` are open string input, without any validation

### Out-of-scope
- There are now 5 tabs in the configuration panel, reaching a good-practice limit. Reorganising the options should be part of a subsequent ticket

### Before
screenshot-localhost_3000-2024 08 06-11_09_36

### After
screenshot-localhost_3000-2024 08 06-14_55_49

### DIFF
--- a/ext/hivemq-edge-openapi-2024.7.yaml
+++ b/ext/hivemq-edge-openapi-2024.7.yaml
@@ -1,0 +1,5173 @@
+openapi: 3.0.1
+info:
+  contact:
+    url: https://www.hivemq.com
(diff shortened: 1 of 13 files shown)


h2xd (approved):
Frontend LGTM! 🐊

PR #533, opened 2024-08-21
refactor(25329): Refactor the toolbars in the workspace

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/25329/details/

The PR contributes to the bi-directionality epics by streamlining the interactions with the "nodes" in the `workspace`

A contextual toolbar has been added to every interactive node: `adapters`, `edge`, `bridges` and `groups`. 
The lonesome "group" icon has been removed from the top-left corner of the screen and replaced by a contextual CTA in the node's toolbar.

It includes the following CTAs:
- `Open the overview panel` for the selected node
- `manage inward subscriptions` for adapters
- `manage outward subscriptions` for bidirectional adapters
- `group` for multiple selected adapters
- `expand`/`collapse` for the group
- `ungroup` for the group

The PR also adds configuration for the minimum and maximum zoom of the `workspace` and extends the default range (0.25 to 1.5)

### Design
- The toolbar occupies the left and right-hand sides of the node (as all connections are top or bottom, leaving the space)
- The toolbar is only shown for a selected node (by click or ENTER)
- When multiple nodes are selected (SHIFT+click or drag+select), the toolbar will be only shown on the first selected node
- The `inward subscription` and `outward subscription` CTAs are located in a way that again reflects the direction of the subscription mapping, from the adapter viewpoint. They match the visualisation of the subscribed topics on the node. 

### Out-of-scope
- The CTAs for subscription management are placeholder and will be implemented in a subsequent ticket (see https://hivemq.kanbanize.com/ctrl_board/57/cards/25263/details/)
- Selecting multiple nodes might have undesirable side effects when only the first node shows its toolbar. There is no mechanism to store the selected node in order of their selection. It might be implemented in a future ticket

### Before
screenshot-localhost_3000-2024 08 21-11_44_09

### After
screenshot-localhost_3000-2024 08 21-11_27_53

screenshot-localhost_3000-2024 08 21-11_28_06

screenshot-localhost_3000-2024 08 21-11_28_55


Florian-Limpoeck (approved):

hivemq-edge/src/frontend/src/modules/Workspace/components/nodes/ContextualToolbar.tsx
```diff
@@ -0,0 +1,139 @@
...
+}
+
+const ContextualToolbar: FC<ContextualToolbarProps> = ({ id, onOpenPanel, children }) => {
+  const { t } = useTranslation()
+  const { onInsertGroupNode, nodes } = useWorkspaceStore()
+  const theme = useTheme()
+
+  const selectedNodes = nodes.filter((node) => node.selected)
+  const selectedGroupCandidates = useMemo(() => {
+    // TODO[NVL] Should the grouping only be available if ALL nodes match the filter ?
```
oli-hivemq, line 29: That would be an interesting one to user test. My guess would be that a user would expect grouping to be available if at least 2 nodes are selected, as per your code. Hence, if I've experienced grouping before, I could get frustrated if grouping only was available if all nodes were selected. And how would I know why grouping isn't available?

hivemq-edge/src/frontend/src/modules/Workspace/components/nodes/ContextualToolbar.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+    const groupStatus: Status = {
+      runtime: selectedGroupCandidates.every((e: Node<Adapter>) => e.data.status?.runtime === Status.runtime.STARTED)
+        ? Status.runtime.STARTED
+        : Status.runtime.STOPPED,
+      connection: selectedGroupCandidates.every(
+        (e: Node<Adapter>) => e.data.status?.connection !== Status.connection.CONNECTED
+      )
+        ? Status.connection.CONNECTED
+        : Status.connection.DISCONNECTED,
```
oli-hivemq, line 64: am I misunderstanding this? I'm reading:
`if every connection is NOT "CONNECTED", then set connection = CONNECTED, otherwise set it to DISCONNECTED`.

It seems backwards to me 🤔 
If it's on purpose, then maybe add a comment to explain it

vanch3d:
> I haven't reviewed the code yet but from your screenshot it looks like a couple of icons should look similar but their width is different and the arrows look different too. And that bugs me !

It's annoying but it's Google's fault!
It's the best icons I've found to describe `inward` and `outward` subscriptions and I don't know why they are different in shape.
See https://react-icons.github.io/react-icons/ for her source of all icons

hivemq-edge/src/frontend/src/modules/Workspace/components/nodes/ContextualToolbar.tsx
vanch3d: The grouping affordance is skewed when selectable nodes are not suitable (e.g. already in another group or of a different type). 
I'm relying on a `select-then-action` paradigm, using the "global" selection mechanism. Hence some nodes might NOT end up in your group. 
A better alternative would be to implement `action-then-select` paradigm: since you are in a "grouping" context, only nodes that are suitable for grouping can be selected. It's a bit more difficult to implement.

hivemq-edge/src/frontend/src/modules/Workspace/components/nodes/ContextualToolbar.tsx
vanch3d: No, that's a bug. Thanks for spotting that one

PR #541, opened 2024-08-28
fix(25622): fix bug with compact table

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/25622/details/

The fix prevents the `RJSC` registry, which contains templates and widgets, to be permanently modified for the overrides needed by the compact table


vanch3d:
Not fixing the problem, creating others. Discarded

PR #548, opened 2024-09-02
fix(25622): remove compact table from adapter's UISchema

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/25622/details/

The PR prevents a significant bug with the compact table widget, removing the mapping from the existing adapters. 

### Out-of-scope
- the code for te RJSF widgets and templates is still in the frontend. It will be refactored to remove the unexpected side effect in a further ticket.


oli-hivemq (approved):
LGTM 🦣

PR #584, opened 2024-10-07
Feature/26414 tag rest api

DC2-DanielKrueger:
https://hivemq.kanbanize.com/ctrl_board/57/cards/26414/details/


hivemq-edge/src/main/java/com/hivemq/api/model/TagResourceExamples.java
```diff
@@ -0,0 +1,72 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hivemq.api.model;
+
+public class TagResourceExamples {
+
+    public static final String EXAMPLE_OPC_UA = "{\n" +
```
codepitbull, line 20: Either use a statix ObjectMapper or have a test for these.
Reason is that we might change the API and people won't catch the strings in here, breaking our docs.

codepitbull (changes requested):

codepitbull (approved):

PR #588, opened 2024-10-10
Switch to using the CachedConnectionManager

codepitbull:
**Motivation**

Resolves #26848

We saw issues with S7 which were (probably) related to restarts/reloads.

**Changes**

Replace the connection manager with the cached version.

This was outlined in another issue which ran into a similar problem: https://github.com/apache/streampipes/issues/2944


codepitbull:
The CachedConnectionManager is actually fully blocking which causes excessive waittime with the way we handle adapters.

PR #590, opened 2024-10-11
refactor(26909): Refactor device toolbar

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/26909/details/

This PR is part of the bi-directional epic and fixes a consistency issue with node toolbars on the workspace. 

The Device toolbar, to manage tags, is now on the left-hand side, consistent with the mapping toolbar for the adapter nodes 

### Out-of-scope 
- The device node remains underspecified (no information about devices in the system)
- The "overview" CTA on the right-hand side is non-existent. It also means the device node doesn't react to double-click or right-click like the other node. There is no visual feedback to support differences between the two states. This will be dealt with in a subsequent ticket.

### Before
screenshot-localhost_3000-2024_10_11-12_51_10

screenshot-localhost_3000-2024_10_11-12_50_56

### After

screenshot-localhost_3000-2024_10_11-12_51_45

### DIFF
--- a/hivemq-edge/src/frontend/src/locales/en/translation.json
+++ b/hivemq-edge/src/frontend/src/locales/en/translation.json
@@ -718,10 +718,13 @@
       "command": {
         "overview": "Open the overview panel",
         "group": "Group the selected adapters",
+        "device": {
+          "metadata": "Edit device tags"
+        },
         "mappings": {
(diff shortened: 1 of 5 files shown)


h2xd (approved):
LGTM! :cro

PR #592, opened 2024-10-11
Updated plc4x since the new version delivers some fixes for S7

codepitbull:
**Motivation**

PLC4X released a new version in Februray which includes several fixes for S7

**Changes**

Update to PLC4X 0.12.0


DC2-DanielKrueger (changes requested):
we explicitily downgraded because of an bug IIRC like 2 months ago

PR #596, opened 2024-10-15
Clean up brokn modules

codepitbull:
**Motivation**

There was some breakage in the toml-referenced from two modules

**Changes**

Fix the modules.


gradle/libs.versions.toml
```diff
@@ -118,6 +118,7 @@ wiremock-jre8-standalone = { module = "com.github.tomakehurst:wiremock-jre8-stan
 zeroAllocationHashing = { module = "net.openhft:zero-allocation-hashing", version.ref = "zeroAllocationHashing" }
 
 [plugins]
+cyclonedx = { id = "org.cyclonedx.bom", version = "1.10.0" }
```
DC2-DanielKrueger, line 121: what does this plugin do?

DC2-DanielKrueger (approved):

gradle/libs.versions.toml
codepitbull: This is for testing, I hope to replace our libraries stuff with this.
It's an SBOM generator.

PR #597, opened 2024-10-15
Fix/validate incoming update

codepitbull:
**Motivation**

We need to ensure incoming configs are actually valid.


gradle/libs.versions.toml
```diff
@@ -118,6 +118,7 @@ wiremock-jre8-standalone = { module = "com.github.tomakehurst:wiremock-jre8-stan
 zeroAllocationHashing = { module = "net.openhft:zero-allocation-hashing", version.ref = "zeroAllocationHashing" }
 
 [plugins]
+cyclonedx = { id = "org.cyclonedx.bom", version = "1.10.0" }
```
DC2-DanielKrueger, line 121: seems some changes leaked from the other PR. should be fine after merge and rebase.

DC2-DanielKrueger (approved):

PR #615, opened 2024-10-31
Submitting alpha version of the PostgreSQL protocol adapter.

anthonyolazabal:
**Motivation**

First contribution to the protocol adapter list with one for PostgreSQL. Original idea came from a discussion with a customer to be able to connect to a specific tool that rely on a database to collect metrics and add them into the enterprise UNS.
The adapter is quite simple, you define the connection to the server and then in each subscription you can execute a dedicated query. For performance a maximum of query is defined. You can chose the output as one single message with all data as JSON or one message per row formatted as JSON. 

Resolves #<issue>
None

**Changes**

Add new protocol adapter.


sfrehse:
@cla-bot check

PR #625, opened 2024-11-07
Bugfix/27689 opcua findings

DC2-DanielKrueger:
https://hivemq.kanbanize.com/ctrl_board/57/cards/27689/details/

### DIFF
--- a/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/mqtt2opcua/JsonToOpcUAConverter.java
+++ b/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/mqtt2opcua/JsonToOpcUAConverter.java
@@ -48,7 +48,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Field;
-import java.util.Base64;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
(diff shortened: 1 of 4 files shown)


modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/mqtt2opcua/JsonToOpcUAConverter.java
```diff
@@ -363,28 +251,247 @@ public JsonToOpcUAConverter(final @NotNull OpcUaClient client) throws UaExceptio
...
-                return builder.build();
+                return extractExtensionObject(jsonNode, binaryEncodingId);
         }
         throw createException(jsonNode, builtinDataType.name());
     }
 
+    private @NotNull Struct extractExtensionObject(
+            @NotNull final JsonNode jsonNode, @Nullable final NodeId binaryEncodingId) {
+        if (binaryEncodingId == null) {
+            throw new RuntimeException("Binary encoding id was null for nested struct. ");
```
codepitbull, line 262: Isn't that rather an IllegalArgumentException?
Also, why @Nullable and then throw an exception when it's null?

I'd still rather go with using optional as return type here and not throw an exception, it would make the flow more explicit.

codepitbull (changes requested):
Some null-weirdness

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/mqtt2opcua/JsonToOpcUAConverter.java
DC2-DanielKrueger: This is an exceptional case as it will hinder the parsing and in the end writing of the data. Optional.absent would not be a legit case. But yeah Runtime is too generic.

codepitbull (approved):

PR #628, opened 2024-11-08
fix(27781): Fix extraction of mapping schema

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/27781/details/

The PR fixes a bug with the extraction of mapping schemas from the main config schema, whereas potential definitions were not extracted. The fix takes care of both `definitions` and `$defs` syntax for the JSONSchema element.

The PR also fixed two small issues:
- the visual rendering of status on the links to devices in the workspace has been corrected to reflect the creation of outward mappings.
- the topics used for outward mappings are now displayed on the top-side of the adapter node, replicating the pattern introduced with inward mappings. 
  It is worth noting that the display is a temporary placeholder, using a "tag" rendering while in fact it shows the source topic filter. This is to create an expected placeholder, while the tag/topic mapping is being introduced

### Before 
screenshot-localhost_3000-2024_11_11-10_00_34

### After
screenshot-localhost_3000-2024_11_11-10_00_08


h2xd (approved):
LGTM! 🐊

PR #639, opened 2024-11-12
fix(27867): add device node to every adapter in the workspace

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/27867/details/

A simple fix to enable the `workspace` to show the `device` node of every `adapter` created. 

Initially, `devices` were only added for adapters supporting bi-directionality. 

However, the generalisation of `tags` for mapping, regardless of direction, enabled the co-location of the `tags` within the `device` namespace, and the necessity of their presence on the `workspace`.

The PR also fixes a bug with the `device` node not being deleted when their `adapter` is.

### Out-of-scope
- Content of the `device` nodes is still not accurate but will be delivered in subsequent tickets

### Before
screenshot-localhost_3000-2024_11_12-15_39_00

### After
screenshot-localhost_3000-2024_11_12-15_38_15


sfrehse (approved):
LGTM but is there a test possible to add?

h2xd (approved):
LGTM! 🐊

vanch3d:
I kind of forgot the tests. on their way :-)

PR #645, opened 2024-11-15
Feature/27616 field mapping api

DC2-DanielKrueger:


hivemq-edge/src/main/java/com/hivemq/protocols/AdapterConfigAndTagsAndFieldMappings.java
```diff
@@ -0,0 +1,154 @@
...
+
+    public AdapterConfigAndTagsAndFieldMappings(
+            final @NotNull ProtocolAdapterConfig adapterConfig,
+            final @NotNull List<? extends Tag> tags,
+            final @NotNull List<FieldMappings> fieldMappings) {
+        Objects.requireNonNull(adapterConfig);
+        Objects.requireNonNull(tags);
+        this.adapterConfig = adapterConfig;
+        this.tags = tags;
+        this.fieldMappings = fieldMappings;
```
codepitbull, line 56: Please make a null check for this field.

codepitbull (approved):
Small comment on a null

hivemq-edge/src/main/java/com/hivemq/protocols/AdapterConfigAndTagsAndFieldMappings.java
DC2-DanielKrueger: actually we do not test fields that are annotated with @NotNull as we are warned when handing over a nullable field. Unless we ignore the warning no null can ever be handed over.

PR #688, opened 2024-12-10
fix(28833): Ensure JSONSchema examples are array

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/28833/details/

There is a conflict in the definition of the `examples` keyword between two versions of JSONSchema
- v6 defines it as 
  ```examples?: JSONSchema6Type[] | undefined;```
- v7 defines it as 
  ``` examples?: JSONSchema7Type | undefined;```

The documentation for the validator used in the `RJSF` (AJV) states for v7:  
```
examples: an array of data instances. Ajv does not check the validity of these instances against the schema.
```
and validates the property as an array. This results in several schemas being rejected by the topic filter UX

The PR ensures that every `examples` handled by the frontend is created as an array of values


hivemq-edge/src/frontend/src/components/rjsf/MqttTransformation/components/schema/PropertyItem.tsx
```diff
@@ -40,6 +42,9 @@ const PropertyItem: FC<PropertyItemProps> = ({
   const propertyName = property.title
   const path = [...property.path, property.key].join('.')
 
+  // TODO[NVL] Only extracting the first value of any examples. Might want to provide something more user-friendly
+  const examples = Array.isArray(property.examples) ? property.examples[0] : property.examples
```
sfrehse, line 46: What happens if examples is an array but contains zero elements.

sfrehse (approved):

PR #699, opened 2024-12-12
fix(28866): Fix the configuration flags of the adapter nodes

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/28866/details/

The PR updates the old "topic tag" in the adapter node with the new mapping paradigm. 

- They cannot be called "tags" anymore since we now have "Device Tags" in the mapping
- As they are basically labels to the link handles on a node, they could be termed "configuration labels"
-  Northbound mappings provides the topics for the labels to the connection with the Edge node
- Southbound mappings provides the topics filters for the labels to the connection with the Device node

The PR also fixes the connection status of an adapter, as rendered by the colour of the links between the nodes. Bi-directionality is rendered by an arrowed link to the Device node, by contract to a plain link when uni-directional,

### Before
screenshot-localhost_3000-2024_12_12-10_59_13

### After

screenshot-localhost_3000-2024_12_12-10_57_46


sfrehse (approved):
Makes sense

PR #708, opened 2024-12-16
fix(28713): change generated name for array items

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/28713/details/

The PR change the pattern of the default name for items in a `RJSF` form. 
- the pattern in used is more readable: `name #1 - content`
- the default name is `item`
- if a `ui:title` is given, then it replaces the name

Note that the fix involves overriding a name that is hardcoded by the `RJSF` library. It might not be resilient to code updates.

### Before
screenshot-localhost_3000-2024_12_16-09_13_30

### After
screenshot-localhost_3000-2024_12_16-09_12_58


vanch3d:
> @vanch3d is the `ui:title` provided by all protocol adapters?

It's defined for `mappings`; any other array in forms will be `item`

sfrehse (approved):

PR #709, opened 2024-12-16
fix(28963): reintroduce batch upload in northbound mappings

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/28963/details/

The PR re-introduces the `batch upload` functionality for northbound mappings

### Out-of-scope
- The batch upload requires `tags` to be defined BEFORE the mappings are created. There is no way to verify this and there is no way to extend the batch process to create requirements and dependencies on the fly. The feature will need to be investigated in a further epic.

### Before 
screenshot-localhost_3000-2024_12_16-10_11_54

### After 
screenshot-localhost_3000-2024_12_16-10_08_56


vanch3d:
The ticket has been archived and changes cannot be made at this point. Closing

PR #713, opened 2024-12-17
Add version to config

DC2-DanielKrueger:
https://hivemq.kanbanize.com/ctrl_board/57/cards/27422/details/

Currently we have no way to find out for what edge version a given config.xml was written. This makes it hard to find out which migrations might be needed. A simple version field that indicates that will resolve this issue.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/configuration/entity/HiveMQConfigEntity.java
+++ b/hivemq-edge/src/main/java/com/hivemq/configuration/entity/HiveMQConfigEntity.java
@@ -34,6 +34,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * @author Dominik Obermaier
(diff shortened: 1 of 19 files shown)


hivemq-edge/src/main/java/com/hivemq/configuration/entity/HiveMQConfigEntity.java
```diff
@@ -44,6 +45,11 @@
 @SuppressWarnings({"FieldMayBeFinal", "FieldCanBeLocal"})
 public class HiveMQConfigEntity {
 
+    public static final int CURRENT_CONFIG_VERSION = 1;
```
codepitbull, line 48: We need something to remind us to increase the number when something actually changes.
Like a breaking test.
Also: What happens with the configs contributed by adapters?

hivemq-edge/src/main/java/com/hivemq/configuration/entity/adapter/ProtocolAdapterEntity.java
```diff
@@ -41,6 +43,9 @@ public class ProtocolAdapterEntity {
     @XmlElement(name = "protocolId", required = true)
     private @NotNull String protocolId;
 
+    @XmlElement(name = "protocolId")
```
codepitbull, line 46: Wrong

hivemq-edge/src/main/java/com/hivemq/configuration/entity/HiveMQConfigEntity.java
DC2-DanielKrueger: I added an integration test for the json schema that is generated by each adapter

codepitbull (approved):

PR #729, opened 2025-01-06
fix(28895): Fix rendering of metrics for groups

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/28895/details/

The PR fixes a bug in the rendering of the list of metrics available for groups. The process was considering device nodes in the list of available targets for metrics, resulting in a wrong list 

Device nodes are now properly excluded from the processing of metrics

### Before 
screenshot-localhost_3000-2025_01_06-18_43_53

### After
screenshot-localhost_3000-2025_01_06-18_44_43


sfrehse (approved):

oli-hivemq (approved):
LGTM 🦣

PR #730, opened 2025-01-06
fix(28933): Add device tags to the node visualisation

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/28933/details/

The PR fixes a glaring gap. Device tags are now rendered on the device node, similar to how topic filters and topics are shown on the adapters.

The PR also improves the layout of the device node

### Out-of-scope
- The "glued" relation between a device and its adapter adapter is rendered by a fixed distance between the two nodes, from top to top. The distance reflects the gap needed to render two tags in the device node. The layout should be more dynamic, by having a constant length of the link between the two nodes, i.e. from bottom to top. This will be done in a subsequent ticket

### Before 
screenshot-localhost_3000-2025_01_06-19_24_56

### After
screenshot-localhost_3000-2025_01_06-19_30_10


sfrehse (approved):

oli-hivemq (approved):
It'd be interested in having a visual designer propose a few tweaks to the styling, but otherwise it looks good to me 🦣

PR #735, opened 2025-01-09
refactor(27085): Redesign the policy dry-run flow for a better UX

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/27085/details/

The PR completely redesigns the UI for the validation (or "dry-run") of a policy created in the Data Hub `Designer`, in particular the main toolbar and the access  to the two `Check` and `Publish` operations

### Design
- The "step-by-step" toolbar has been removed, restricting the (top-left) toolbar to the construction toolbox and the summary info panel
- The contextual toolbar associated with each node has been redesigned
  - its style is consistent with the `Workspace`  
  - the `Check` and `Publish` CTAs are added to both `Data Policy` and `Behavior Policy` nodes
- the `Check` button runs the "dry-run" validation, as before. It is always available on the relevant nodes
- After a "dry-run", a `Show Report` CTA will be enabled on the toolbar, opening a side panel containing the result of the dry-run
  - a summary message (alert styled)
  - if `ERROR`, a list of errors, each with two buttons: the `Show on Designer` (zoom on the relevant node) and `Open Configuration` (replace the report with the config dialog for the relevant node)
  - if `SUCCESS`, the bottom of the report contains a `Publish` CTA 
- The `Publish` button is also enabled on the toolbar, so access to the side-panel report is not necessary in the case of `SUCCESS`
- The `Check` button, once a "dry-run" process has ran, had a composed `Clear` button besides it, allowing users to clear the whole validation (removing the marks on the node for example)

### Out-of-scope
- Successfully publishing a policy keeps the draft open. This behaviour should be changed in a subsequent ticket.
- The main toolbar and the visual identity of the nodes (as well as their accessibility), see comments below, will be part of a subsequent ticket.

### Before 
screenshot-localhost_3000-2025_01_10-09_15_57

### After
screenshot-localhost_3000-2025_01_10-09_13_36

screenshot-localhost_3000-2025_01_10-09_13_51

screenshot-localhost_3000-2025_01_10-09_14_05

screenshot-localhost_3000-2025_01_10-09_14_44


oli-hivemq:
These comments are not related to this PR in particular, but I'll comment on here anyway.

CleanShot 2025-01-10 at 18 30 45@2x
The alignments looks strange to me because of the group titles. Without the group titles, it wouldn't look strange to me.

CleanShot 2025-01-10 at 18 32 33@2x
I assume that...
- a green shadow means good
- a red shadow means something is not working
- what's a blue shadow? Information?

Additional questions:
- what does it mean to have a green arrow with the start of the arrow being red, and/or the end of the arrow having a red shadow?

Even if colours can be used to represent a state, I would also use another way to indicate what's going on because colour-blind people might not see the difference between green and red. Hence, maybe a alternative icon representing the same thing but with a cross showing there's an issue, and a tooltip explaining what it is, is something that might be worth considering.

Has some user testing be done or considered? (always asking but we're likely understaffed to make this happen)

oli-hivemq (approved):
I skimmed through this pretty quickly but LGTM 🦣

sfrehse:
The general changes on the toolbar directly on top of the data policy make sense and give context what "steps" are to be done.

The toolbar in the screenshot below doesn't look clean. Does it make sense to re-use the new color schema you have shared in the channel? However, I guess the margins are missing on the left and right. Is there a visual chance to clarify that the user is in "Draft Policy". The text is too easy to overread.

Screenshot 2025-01-13 at 11 31 45

PR #742, opened 2025-01-13
fix(29347): fix bug with empty targets of deletion

vanch3d:
See 

A quick fix for preventing an unnecessary confirmation message when there are no element to delete 

### Before 
screenshot-localhost_3000-2025_01_10-16_52_40


oli-hivemq (approved):
LGTM 🦣

sfrehse (approved):

PR #745, opened 2025-01-13
refactor(29358): Refactor the Designer visual identity

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/29358/details/

The PR introduces visual changes to most elements (nodes and edges) in the `Data Hub Designer`. The aim is to highlight the paradigms that are similar to the `Workspace` (selectable and draggable nodes, toolbar for commands, status) while ensuring the two tools are immediately differentiated 

### Design
- The layout of the nodes has been streamlined, to ensure handles (connection ports) are only on the left (target) or on the right (source) hand side of a node. It helps to convey a sense of "pipeline" with the `Designer`
- The visual attributes of nodes and edges have been changed to convey different aspects:
  - Handles (the port on a node used for connections) are visually consistent
     - round for source handles, square for target handles
     - plain for primary (policy pipeline) handles, hollow for secondary (noise configuration) handles
     - all handles are black, except for primary pipeline handles that convey success (green) or error (red) processing of the policy (e.g `onSuccess` on a `Data Policy` or on failed transition in a `Behaviour Policy)
  - The border of nodes is in a plain light grey, changing to blue (when the node is selected or focused), green (when its "dry-run" validation is successful) or red (when it's a failure)
  - Edges are now displayed as (smooth) stepped connectors, to highlight the flow between the handles of nodes
  - Edges are displayed in plain colour (grey), changing to green (success) or red (failure) based on the "dry run" status of the validation of their source
  - A coloured header is used in nodes to show the type of element: 
     - policies (data and behaviour) in orange 
     - resources (script and schema) in pink 
     - operators (any other constituent of a policy, e.g. topic filter, transition, etc.) in blue
  - Some of the internal properties of a node are carefully selected to be displayed inside the node, in a way that adds labelling to the more important handle 

The PR also improves the layout of the handles and, by effect, the alignment of nodes on a gid layout. It fixes https://hivemq.kanbanize.com/ctrl_board/57/cards/29435/details/

### Out-of-scope
- handles are positioned so that they align with the nodes on both sides of the connection. The utility is not bound to the ChakraUI Theme, it will be done in a subsequent ticket
 
### Before

screenshot-localhost_3000-2025_01_13-16_37_50

### After
screenshot-localhost_3000-2025_01_13-16_38_40

screenshot-localhost_3000-2025_01_20-15_11_09


hivemq-edge/src/frontend/src/extensions/datahub/designer/behavior_policy/BehaviorPolicyNode.tsx
```diff
@@ -1,34 +1,40 @@
...
       </NodeWrapper>
-      <CustomHandle type="target" position={Position.Left} id={BehaviorPolicyData.Handle.CLIENT_FILTER} />
-      <CustomHandle type="source" position={Position.Right} id={BehaviorPolicyData.Handle.TRANSITIONS} />
+      <CustomHandle
+        type="target"
+        position={Position.Left}
+        id={BehaviorPolicyData.Handle.CLIENT_FILTER}
+        style={{
+          top: `calc(var(--chakra-space-3) + 12px + 48px `,
+        }}
```
oli-hivemq, line 28: nit: this style could be saved as a const and reused a few lines below. This way if you need to edit the style you'd only do it once

hivemq-edge/src/frontend/src/extensions/datahub/designer/client_filter/ClientFilterNode.tsx
```diff
@@ -33,7 +26,7 @@ export const ClientFilterNode: FC<NodeProps<ClientFilterData>> = (props) => {
           id={`${id}-${index}`}
           key={`${id}-${client}-${index}`}
           style={{
-            top: `calc(var(--chakra-space-3) + 12px + ${index * 24}px + ${0.5 * index}rem)`,
+            top: `calc(var(--chakra-space-3) + 12px + 48px + ${index * 24}px + ${0.5 * index}rem)`,
```
oli-hivemq, line 29: Potential to think about providing a utility function that does this kind of calculation for you. A very similar calculation is done in `BehaviorPolicyNode.tsx`, why not creating a utility function to have 1 source of truth?

hivemq-edge/src/frontend/src/extensions/datahub/designer/data_policy/DataPolicyNode.tsx
```diff
@@ -43,7 +69,7 @@ export const DataPolicyNode: FC<NodeProps<DataPolicyData>> = (props) => {
         id={DataPolicyData.Handle.ON_ERROR}
         className={DataPolicyData.Handle.ON_ERROR}
         style={{
-          top: `calc(var(--chakra-space-3) + 10px + 16px + 0.5rem)`,
+          top: `calc(var(--chakra-space-3) + 12px + 16px + 0.5rem + 44px)`,
```
oli-hivemq, line 72: Similar point as made above. This calculation for the positioning could either be done from a utility function, or you could define this style as a const within this component as you're using the same line more than once

oli-hivemq: another point: you're mixing `rem` and `px`, which isn't an issue as long as the assumption `1rem=16px` doesn't change.

hivemq-edge/src/frontend/src/extensions/datahub/designer/operation/OperationNode.tsx
```diff
@@ -41,36 +41,78 @@ export const OperationNode: FC<NodeProps<OperationData>> = (props) => {
...
+      )}
+      {isSerialiser && (
+        <CustomHandle
+          type="target"
+          position={Position.Left}
+          id={OperationData.Handle.SCHEMA}
+          style={{
+            top: `calc(var(--chakra-space-3) + 12px + 16px + 1rem + 44px)`,
+            background: 'var(--chakra-colors-white)',
+            borderColor: 'var(--chakra-colors-black)',
```
oli-hivemq, line 76: I assume that this component doesn't support light & dark modes, right?

hivemq-edge/src/frontend/src/extensions/datahub/designer/operation/OperationNode.tsx
```diff
@@ -41,36 +41,78 @@ export const OperationNode: FC<NodeProps<OperationData>> = (props) => {
...
+          }}
+        />
+      )}
+      {isSerialiser && (
+        <CustomHandle
+          type="target"
+          position={Position.Left}
+          id={OperationData.Handle.SCHEMA}
+          style={{
+            top: `calc(var(--chakra-space-3) + 12px + 16px + 1rem + 44px)`,
```
oli-hivemq, line 74: Same comment as above regarding a utility function.
I will stop adding this kind of comment from now on because it applies in a lot of places apparently

hivemq-edge/src/frontend/src/extensions/datahub/designer/operation/OperationNode.tsx
```diff
@@ -41,36 +41,78 @@ export const OperationNode: FC<NodeProps<OperationData>> = (props) => {
...
         </VStack>
       </NodeWrapper>
-      <CustomHandle type="target" position={Position.Left} id={OperationData.Handle.INPUT} />
+      <CustomHandle
+        type="target"
+        position={Position.Left}
+        id={OperationData.Handle.INPUT}
+        style={{
+          top: `calc(var(--chakra-space-3) + 12px + 44px)`,
+          // background: 'green',
```
oli-hivemq, line 53: nit: cleanup? (same for line 64)

oli-hivemq (approved):
I would recommend to look into refactoring all the code around position as I saw multiple instances of `44px` and similar calculations all over the place.
It wouldn't take much to refactor this and have 1 source of truth (or at least minimise these sources of truth) and make your life easier when a refactoring request or an adjustment is needed.

Approving anyway not to block you 🦣

hivemq-edge/src/frontend/src/extensions/datahub/designer/behavior_policy/BehaviorPolicyNode.tsx
vanch3d: I've found a way (easy of course) to integrate them in ChakraUI Theme so will be done next iteration

hivemq-edge/src/frontend/src/extensions/datahub/designer/data_policy/DataPolicyNode.tsx
vanch3d: `rem` and `px` might change but, hidden in the magic CSS code is the fact that I'm copying the value from other elements I'm trying to match in height. Your utility will work nicely, in particular, if I can link directly to the Chakra design token for the value. No magic match!

hivemq-edge/src/frontend/src/extensions/datahub/designer/operation/OperationNode.tsx
vanch3d: It should. Bug to fix

sfrehse:
Once I want to save the ID of a data policy, it isn't stored; an error occurs after checking repeatedly.

sfrehse:
> Once I want to save the ID of a data policy, it isn't stored; an error occurs after checking repeatedly.

Pressing enter holds but clicking not

vanch3d:
> > > Once I want to save the ID of a data policy, it isn't stored; an error occurs after checking repeatedly.
> > 
> > 
> > Pressing enter holds but clicking not
> 
> Sorry, this is not true. The id of a data-policy is stored once the user left the focus one time from the data-policy node

I think I noticed that issue but haven't been able to pin it down. Will look again, and opening a new ticket for it

PR #746, opened 2025-01-16
Feature/28983 generate jersey from spec

DC2-DanielKrueger:
https://hivemq.kanbanize.com/ctrl_board/57/cards/28983/details/


hivemq-edge/src/main/java/com/hivemq/api/utils/BridgeUtils.java
```diff
@@ -0,0 +1,131 @@
...
+import com.hivemq.edge.api.model.Status;
+import com.hivemq.edge.api.model.WebsocketConfiguration;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.stream.Collectors;
+
+public class BridgeUtils {
+
+    public static @NotNull Bridge convert(final @NotNull MqttBridge mqttBridge, final @NotNull Status status) {
```
codepitbull, line 37: Would be nicer to have this directly on one of the classes as a to or from-method.

hivemq-edge/src/main/java/com/hivemq/api/utils/BridgeUtils.java
```diff
@@ -0,0 +1,131 @@
...
+                .host(mqttBridge.getHost())
+                .port(mqttBridge.getPort())
+                .clientId(mqttBridge.getClientId())
+                .keepAlive(mqttBridge.getKeepAlive())
+                .sessionExpiry(mqttBridge.getSessionExpiry())
+                .cleanStart(mqttBridge.isCleanStart())
+                .username(mqttBridge.getUsername())
+                .password(mqttBridge.getPassword())
+                .loopPreventionEnabled(mqttBridge.isLoopPreventionEnabled())
+                .loopPreventionHopCount(mqttBridge.getLoopPreventionHopCount() < 1 ?
```
codepitbull, line 48: Is this a config error or do we expect negative values?

codepitbull (approved):
Nice work :)

hivemq-edge/src/main/java/com/hivemq/api/utils/BridgeUtils.java
DC2-DanielKrueger: We expect 0 as unlimited, i guess negative numbers work the same logically.

hivemq-edge/src/main/java/com/hivemq/api/utils/BridgeUtils.java
DC2-DanielKrueger: not sure, we did this for data hub and it turned out to be a semi good idea. I think a 3rd component is actually correct, as the Bridge should not care about the MqttBridge and vice versa.

PR #757, opened 2025-01-23
refactor(29589): Refactor the nodes' initial position in the Designer

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/29589/details/
See https://hivemq.kanbanize.com/ctrl_board/57/cards/29579/details/

The PR fixes the computation of the initial position of the nodes on the canvas when a policy is loaded. It changes the basis of the layout to match the new design constraints (left-to-right orientation of policy graphs)

The positions are not persisted with the policy so the loaded policy needs to be re-layouted at load time. The process in place is a simple predecessor- and grid-based position shifting, not a full graph layout: overlaps are still expected. 

The PR also fixes an irritating anti-pattern: when constructing a node by dragging from the handle, the new node is created based on the expected position of the new handle, guaranteeing alignment.

### Before
screenshot-localhost_3000-2025_01_23-15_39_36

screenshot-localhost_3000-2025_01_23-15_41_11

### After

screenshot-localhost_3000-2025_01_23-15_40_08

screenshot-localhost_3000-2025_01_23-15_40_47


sfrehse (approved):
This is great and reduces friction a lot. However, I couldn't test yet, but the code looks fine for me.

PR #758, opened 2025-01-23
feat(29524): Enable modification of published policy

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/29524/details/

This PR fixes the publishing process of a policy by allowing users to modify a published policy.

In the previous releases, policies could only be created once, as the "modify" part of the process had never been implemented. 

The constraints on policies, both `Data` and `Behaviour` remain the same
- the policy node itself cannot be modified (in practice, its 'id`)
- the triggers (topic filter and client filter) cannot be modified
- Modifying an existing resource creates a new one (new version)

### Out-of-scope
- Multiple resources are not properly published, see https://hivemq.kanbanize.com/ctrl_board/57/cards/29734/details/
- Persistence of the policy layout is not yet supported, see https://hivemq.kanbanize.com/ctrl_board/57/cards/29735/details/

### Before 
screenshot-localhost_3000-2025_01_27-11_00_20

### After
screenshot-localhost_3000-2025_01_27-11_00_54


tolumq (approved):
LGTM

PR #763, opened 2025-01-27
fix(29608): Fix the display of resource version

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/29608/details/

The PR fixes the display of the resource's version in the Designer. The version number will be shown, `DRAFT` for a new resource and `MODIFIED` for a new version of an existing resource. The fix applies to both `Schema` and `Function`.

For compatibility purposes, the non-published version of a resource is internally identified by a number (9 ** 9 for DRAFT, incremented for MODIFIED). These fake numbers are unlikely to be reached by the `Data Hub`c version increment. But they should not be displayed as such 

### Out-of-scope
- There is a bug loading a new version, where `latest` is not found (nor the list of the different versions). It will be fixed in another ticket.

### Before 
screenshot-localhost_3000-2025_01_27-11_37_00

### After
screenshot-localhost_3000-2025_01_27-11_11_14
screenshot-localhost_3000-2025_01_27-11_11_36
screenshot-localhost_3000-2025_01_27-11_12_29
screenshot-localhost_3000-2025_01_27-11_19_42

### DIFF
--- a/hivemq-edge/src/frontend/src/extensions/datahub/utils/node.utils.spec.ts
+++ b/hivemq-edge/src/frontend/src/extensions/datahub/utils/node.utils.spec.ts
@@ -25,7 +25,6 @@ import {
   DesignerStatus,
   OperationData,
   ResourceWorkingVersion,
-  ResourceStatus,
   SchemaType,
   StrategyType,
   TransitionData,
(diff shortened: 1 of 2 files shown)


h2xd (approved):
LGTM! 🐊

PR #769, opened 2025-01-28
feat(29319): Add notification of not writable config

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/29319/details/

The PR adds a notification when the `config-writeable` capability is NOT defined

### Out-of-scope
- Capabilities MUST be curated in the OpenAPI specs, not replicated on the front end. This will be part of a further ticket
- The intended usage of the capability has been inverted. The critical aspect worth notifying users about is when the config is NOT writable, triggering error messages (403) at the front end. However, that state is (mis)detected by the capability (config is writable) being not present in the capabilities list. The `undefined` content could mean not supported but also not found or API to get the capabilities failed. As the logician says: "Absence of Evidence does not mean Evidence of Absence”. This will be fixed in a further update 

### After
screenshot-localhost_3000-2025_01_28-19_11_03

### DIFF
--- a/hivemq-edge/src/frontend/src/locales/en/translation.json
+++ b/hivemq-edge/src/frontend/src/locales/en/translation.json
@@ -30,7 +30,7 @@
   },
   "capabilities": {
     "WRITEABLE_CONFIG": {
-      "title": "Config cannot be manipulated via the REST API",
+      "title": "Config cannot be manipulated via the web app",
       "description": "Changes to the configuration made via the web app will NOT be persisted back into the config.xml. The requests will fail with an error message highlighting this situation."
     }
   },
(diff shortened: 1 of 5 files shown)


hivemq-edge/src/frontend/src/locales/en/translation.json
```diff
@@ -28,6 +28,12 @@
       "status_ERROR": "Error"
     }
   },
+  "capabilities": {
+    "WRITEABLE_CONFIG": {
+      "title": "Config cannot be manipulated via the REST API",
```
sfrehse, line 33:
```suggestion
      "title": "Config cannot be manipulated via the UI or REST API",
```

sfrehse (approved):

h2xd (approved):
LGTM! 🐊

hivemq-edge/src/frontend/src/locales/en/translation.json
vanch3d: UI won't cut it for users (and I guess REST API won't either). Replacing by `Config cannot be manipulated via the Web Application`

PR #794, opened 2025-02-10
Upgrade compile target to JDK 17

codepitbull:
**Motivation**

Resolves #30163

**Changes**
JDK 17 is our new baseline.


caoccao (approved):
Looks good to me.

PR #797, opened 2025-02-12
feat(30125): Create the OpoenAPI specs for the Data Combining feature

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30340/details/

The PR specifies the API endpoints and entities (schemas, error messages, etc.) as anticipated for the Data Combining feature. 

It will be used to generate the stubs for both backend and frontend

### Design
- TODO

### Notes
- Since it's a brand new feature, the OpenAPI has been created in a new empty YAML document, to allow reading and reviewing clear of the (huge) existing specs
- When existing elements (e.g. schemas, headers, ...) are reused, they are referenced using the `$ref` attribute, pointing to the relevant element from the `master` YAML file, e.g. 

```yaml
    DataCombining:
      type: object
      properties:
        instructions:
          type: array
          description: List of instructions to be applied to incoming data
          items:
            $ref: '../hivemq-edge-openapi-2025.3-SNAPSHOT.yaml#/components/schemas/Instruction'
```
- This split structure is still a work in progress
- Once accepted, the complete YAML master document will be updated


ext/openAPI/30125-data-combining.yaml
```diff
@@ -0,0 +1,271 @@
...
+      required:
+        - instructions
+    EntityReferenceList:
+      type: object
+      properties:
+        items:
+          type: array
+          items:
+            $ref: '#/components/schemas/EntityReference'
+    EntityType:
```
DC2-DanielKrueger, line 251: why do we care about the type? Shouldnt this just be TAG or TOPIC_FILTER? Logically we can not know whether a publish came from a bridge or from a normal client. It is all just MQTT traffic from Edge PoV.

ext/openAPI/30125-data-combining.yaml
```diff
@@ -0,0 +1,271 @@
...
+      type: string
+      description: These are the prime entities owning tags and topic filters
+      enum: [ADAPTER, DEVICE, BRIDGE, EDGE_BROKER]
```
DC2-DanielKrueger, line 254: Why is ADAPTER and DEVICE separate values here?

ext/openAPI/30125-data-combining.yaml
vanch3d: It's the encoding for the container, not for the type of data. `EntityReferenceList` is an `object` containing the property `items`, which is an `array` of `EntityReference`

```
/// EntityReferenceList
{ 
   items: []
}
```

`items` is an unfortunate choice but I've kept the pattern for all (almost) type returning a container

ext/openAPI/30125-data-combining.yaml
vanch3d: Because they are different entities in the frontend: you get the N/S mapping from the adapter, and the tags from the device. The backend doesn't have any knowledge of a thing called device (yet)

ext/openAPI/30125-data-combining.yaml
```diff
@@ -0,0 +1,271 @@
...
+          type: string
+          description: The user-facing name of the combiner
+        description:
+          type: string
+          description: The user-facing description of the combiner
+        sources:
+          $ref: '#/components/schemas/EntityReferenceList'
+        mappings:
+          $ref: '#/components/schemas/DataCombiningList'
+      required:
```
DC2-DanielKrueger, line 203: i think at least sources should be required as a combiner without them makes no sense, right?

DC2-DanielKrueger (approved):

vanch3d:
Rebased to epic btranch

PR #807, opened 2025-02-18
feat(30403): Support the main user flow for creating combiners

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30403/details/

The PR is the next step in building the UX for the data-combining feature of Edge, see https://hivemq.kanbanize.com/ctrl_board/57/cards/29097/details/. It creates the main entry point to the feature by supporting users in creating and defining a  `Combiner` that is used to put together sources of data points to combine

### Design
- The functionality is available through the `Workspace`. With multiple sources for data combining, a single table with homogenous content (e.g tag) would not suffice
- End users are expected to instantiate data combining in four steps
   1.  in the `Workspace`, select the entities that "own" the data sources (e.g. tags or topic filters) that must be combined. When the selection is valid (the entities are of the right types), the users can now create a `Combiner`, displayed on the `Workspace` like the other entities and connected to all selected sources and to the Edge 
   2. In the `Combiner` configuration panel, create a new `Mapping` that defines a destination `topic`. A schema MUST be associated with the topic.
   4. In the `Mapping`, select the data sources for the combining. It consists of filtering and adding to the "bag" the tags and topic filters strictly defined by the selected entities.  Data sources not owned by the selected entities will NOT be eligible for selection. The selected data sources expose their schema to the users
   6. The user can now create a series of `instructions` that define how individual properties from the source schemas are being mapped to individual properties of the destination topic. To be valid, every required property of the destination topic MUST be mapped to sources. 

- The working assumption is that only the following nodes are eligible sources for a combiner
   -  `ADAPTER` as a source of tags
   - `BRIDGE` as a source of topic filters (but bridge subscriptions are not supported)
   - `EDGE` as the owner of the topic filters (not actively linked, as a destination of the combiner) 
- At least one source MUST be selected for a `Combiner` to be created. 

### Out-of-scope
- There are issues with the OpenAPI specs (and the related JSON-Schema used for the `RJSF` forms) that will be fixed in a further ticket, see https://hivemq.kanbanize.com/ctrl_board/57/cards/30463/details/
- The Data Combining itself, while being supported by the "raw" `RJSF` form, needs a dedicated custom UX that will be implemented in a further ticket, see https://hivemq.kanbanize.com/ctrl_board/57/cards/30405/details/
- The sources of a `Combiner` must be manageable by the end users: adding, deleting, modifying. The CRUD operations will be supported, both in the configuration panel and on the `Workspace` in a further ticket

### After
HiveMQ-Edge-02-18-2025_11_02_AM

HiveMQ-Edge-02-18-2025_09_31_AM (1)

HiveMQ-Edge-02-18-2025_09_32_AM


hivemq-edge/src/frontend/src/modules/Workspace/components/nodes/NodeCombiner.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import { useContextMenu } from '@/modules/Workspace/hooks/useContextMenu.ts'
+import ContextualToolbar from '@/modules/Workspace/components/nodes/ContextualToolbar.tsx'
+
+const NodeCombiner: FC<NodeProps<Combiner>> = ({ id, selected, data, dragging }) => {
+  const { onContextMenu } = useContextMenu(id, selected, '/workspace/combiner')
+
+  return (
+    <>
+      <ContextualToolbar id={id} title={data.name} onOpenPanel={onContextMenu} dragging={dragging}></ContextualToolbar>
```
oli-hivemq, line 19: Would `<ContextualToolbar id={id} title={data.name} onOpenPanel={onContextMenu} dragging={dragging} />` work?

hivemq-edge/src/frontend/src/modules/Workspace/hooks/useGetFlowElements.ts
```diff
@@ -75,9 +83,45 @@ const useGetFlowElements = () => {
...
+          })
+          // TODO[] Error message for missing references
+          .filter((node) => node) || []
+
+      const { nodeCombiner, edgeConnector, sourceConnectors } = createCombinerNode(
+        combiner,
+        (index - deltaPosition) / nbCombiners,
+        sources as Node[],
+        theme
+      )
```
oli-hivemq, line 103: Is there a test for the positioning?
e.g. nbCombiners = 4
deltaPosition = Math.floor((4-1) / 2) = Math.floor(3/2) = 1
at index 0, line 1000 would be `(0 - 1) / 4` = -0.25
at index 1: (1 - 1) / 4 = 0
at index 2: (2 - 1) / 4 = 0.25
at index 3: (3 - 1) / 4 = 0.5

oli-hivemq (approved):
Left a few comments & questions.
Approving to move fast, nothing blocking 🦣

hivemq-edge/src/frontend/src/modules/Workspace/hooks/useGetFlowElements.ts
vanch3d: None but that's a very good idea. 
The reason for the computation above is that I prefer to have nodes aligned. 
The barycenter of the four combiners would be between index 1 and 2, so by doing it that way, index 1 will be vertically aligned with the rest of the graph. 
Nut that's a design decision that must be tested and checked for regression. More tests :-)

PR #809, opened 2025-02-18
Feature/30125: Backend Implementation of the Data Combining API

DC2-DanielKrueger:
https://hivemq.kanbanize.com/ctrl_board/57/cards/30125/details/


hivemq-edge/src/main/java/com/hivemq/configuration/entity/combining/DataCombinerEntity.java
```diff
@@ -0,0 +1,121 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import javax.xml.bind.ValidationEvent;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlElementWrapper;
+import java.util.List;
+import java.util.UUID;
+
+
+public class DataCombinerEntity {
```
codepitbull, line 14: Shouldn't the entities have a name for JSON/XML?

hivemq-edge/src/main/java/com/hivemq/configuration/service/impl/DataCombiningConfigurationServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class DataCombiningConfigurationServiceImpl implements DataCombiningConfigurationService {
+
+    private volatile @NotNull List<DataCombinerEntity> allConfigs = new ArrayList<>();
+
+    @Override
+    public @NotNull List<DataCombinerEntity> getAllConfigs() {
+        return allConfigs;
```
codepitbull, line 32: Shouldn't that be a copy of the list rather than the full reference?

hivemq-edge/src/main/java/com/hivemq/persistence/util/FutureUtils.java
```diff
@@ -76,4 +79,14 @@ public void onFailure(final Throwable throwable) {
             }
         }, MoreExecutors.directExecutor());
     }
+
+    public static void syncFuture(final @NotNull Future<?> future) {
+        try {
+            future.get();
+        } catch (final InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (final ExecutionException e) {
```
codepitbull, line 88: Shouldn't the caller know about the failure?

hivemq-edge/src/main/java/com/hivemq/protocols/ConfigPersistence.java
```diff
@@ -91,4 +93,48 @@ public synchronized void deleteAdapter(final @NotNull String adapterId, final @N
...
+
+    public @NotNull List<DataCombiner> allDataCombiners() {
+        return configurationService.dataCombiningConfigurationService()
+                .getAllConfigs()
+                .stream()
+                .map(DataCombiner::fromPersistence)
+                .collect(Collectors.toList());
+    }
+
+    public synchronized void updateDataCombiners(
```
codepitbull, line 106: This should probably be called updateDataCombiner

hivemq-edge/src/main/resources/config.xsd
```diff
@@ -59,6 +59,15 @@
                     </xs:sequence>
                 </xs:complexType>
             </xs:element>
+            <xs:element name="data-combiners" minOccurs="0">
```
codepitbull, line 62: There should be a follo up ticket to change this into a proper XML definition later on.

hivemq-edge/src/main/java/com/hivemq/combining/CombiningManager.java
```diff
@@ -0,0 +1,150 @@
...
+        return CompletableFuture.completedFuture(null);
+    }
+
+    public @NotNull CompletableFuture<Void> stopAll() {
+        // TODO
+        return CompletableFuture.completedFuture(null);
+    }
+
+    public @NotNull CompletableFuture<AddResult> addDataCombiner(final @NotNull DataCombiner dataCombiner) {
+        if (getCombinerById(dataCombiner.id()).isPresent()) {
```
codepitbull, line 75: A saver way is to use putIfAbsent. It returns null if the key wasn't defined before and you also stay within the guarantees for atomicity of a concurrenthashmap.

codepitbull (approved):
General comments:
DidnÄ't we decide against using Preconditions.checkNotNull in combination with @NotNull annotations?

It would be good to have a general pattern for to/from methods on records and other model classes.

I am still approving since I don't think any of my comments should be blocking you.

hivemq-edge/src/main/java/com/hivemq/persistence/util/FutureUtils.java
DC2-DanielKrueger: yeah method is not used yet. but i had the same thought. not sure

hivemq-edge/src/main/java/com/hivemq/configuration/service/impl/DataCombiningConfigurationServiceImpl.java
DC2-DanielKrueger: same as ProtocolAdapterConfigurationServiceImpl. Yeah a copy or an immutable view could make sense

hivemq-edge/src/main/java/com/hivemq/combining/EntityReference.java
```diff
@@ -0,0 +1,30 @@
...
+
+    }
+
+    public static @NotNull com.hivemq.edge.api.model.EntityReference toModel(final @NotNull EntityReference entityReference) {
+        return new com.hivemq.edge.api.model.EntityReference().id(entityReference.id)
+                .type(entityReference.type.toModel())
+                .isPrimary(entityReference.isPrimary);
+    }
+
+    public static @NotNull EntityReference fromPersistence(final @NotNull EntityReferenceEntity entityReference) {
```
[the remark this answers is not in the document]
DC2-DanielKrueger: not sure. this way the conversions are in one place and easy to find

hivemq-edge/src/main/java/com/hivemq/combining/CombiningManager.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        idToDataCombiner.put(dataCombiner.id().toString(), dataCombiner);
+        final CompletableFuture<AddResult> ret = startCombiner(dataCombiner);
+        configPersistence.addDataCombiner(dataCombiner);
+        return ret;
+    }
+
+    public boolean updateDataCombiner(final @NotNull DataCombiner dataCombining) {
+        Preconditions.checkNotNull(dataCombining);
```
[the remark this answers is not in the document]
DC2-DanielKrueger: relict from copied code

hivemq-edge/src/main/java/com/hivemq/api/resources/impl/CombinersResourceImpl.java
```diff
@@ -0,0 +1,179 @@
...
+        if (instance.isPresent()) {
+            return ErrorResponseUtil.errorResponse(new AlreadyExistsError(String.format(
+                    "DataCombiner already exists '%s'",
+                    combiner.getId())));
+        }
+
+
+        final DataCombiner dataCombiner = DataCombiner.fromModel(combiner);
+        try {
+            combiningManager.addDataCombiner(dataCombiner).get();
```
marregui, line 64: here we call get to materialise the future, and updateCombiner does call get too, but it does it implicitly, ce n'est pas symétrique

marregui (approved):
LGTM!!!

hivemq-edge/src/main/java/com/hivemq/configuration/service/impl/ConfigurationServiceImpl.java
```diff
@@ -156,6 +160,12 @@ public ConfigurationServiceImpl(
         return proxy(ProtocolAdapterConfigurationService.class, protocolAdapterConfigurationService);
     }
 
+    @Override
+    public @NotNull DataCombiningConfigurationService dataCombiningConfigurationService() {
+        // TODO is proxy correct
```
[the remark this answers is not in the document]
DC2-DanielKrueger: i think the proxy is not so much about serilializing calls, but.it also syncs the config file on disc IIRC. I added this TODO to check that on the cleanup ticket.

hivemq-edge/src/main/java/com/hivemq/api/resources/impl/CombinersResourceImpl.java
DC2-DanielKrueger: ce n'est pas symétrique -> what? :-D guess you mean it is not symmetric? Yeah, the methods are not fleshed out as the whole handling needs to be implemented. We do not have AsyncResponse, which means we have to sync the future at some point unfortunately. I wondered during the rough implementation whether async even makes sense then get() is called in the end.

PR #813, opened 2025-02-20
fix(30439): Enable relative path for redirection

caoccao:
https://hivemq.kanbanize.com/ctrl_board/57/cards/30439/details/

**Motivation**

This is to allow Edge to be accessible behind an Ingress that does TLS-termination we generate a redirect to an http-URL.

**Changes**

* Add server property `jersey.config.server.headers.location.relative.resolution.disabled` to jersey
* Add unit test case to verify `/` is redirected to `/app/`

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/http/JaxrsHttpServer.java
+++ b/hivemq-edge/src/main/java/com/hivemq/http/JaxrsHttpServer.java
@@ -95,7 +95,7 @@ public class JaxrsHttpServer {
                     ResourceConfig resources = resourceConfig == null ? new ResourceConfig() : resourceConfig;
                     // https://github.com/eclipse-ee4j/jersey/issues/2986
                     // This server property tells jersey not to resolve relative location
-                    // so that / -> /app/ doesn't get resolved to http://domain/app/.
+                    // so that the Location of the response headers is /path/ instead of http://host:port/path/.
                     resources.property(LOCATION_RELATIVE_RESOLUTION_DISABLED, true);
 
                     for (JaxrsHttpServerConfiguration config : configs) {
--- a/hivemq-edge/src/main/java/com/hivemq/http/JaxrsHttpServer.java
+++ b/hivemq-edge/src/main/java/com/hivemq/http/JaxrsHttpServer.java
@@ -58,9 +58,11 @@ public class JaxrsHttpServer {
 
     public static final int MAX_BINDING_PRIORITY = 1;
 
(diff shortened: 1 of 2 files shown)


hivemq-edge/src/test/java/com/hivemq/embedded/internal/EmbeddedHiveMQImplHttpTest.java
```diff
@@ -0,0 +1,119 @@
...
+            assertEquals(1, apiListeners.size(), "The API listener count should be 1");
+            assertEquals(randomApiPort, apiListeners.get(0).getPort(), "The Admin API port should match");
+
+            final HttpClient httpClient = HttpClient.newBuilder().followRedirects(HttpClient.Redirect.NEVER).build();
+            final HttpRequest httpRequest =
+                    HttpRequest.newBuilder().uri(URI.create("http://localhost:" + randomApiPort + "/")).GET().build();
+
+            final HttpResponse<String> httpResponse =
+                    httpClient.send(httpRequest, HttpResponse.BodyHandlers.ofString());
+            assertEquals(307, httpResponse.statusCode(), "The status code should be 307");
```
codepitbull, line 109: Sorry for not pointing it out.
For asserts we stick to AssertJ.
You will still see other assert styles in older code but the general rule of thumb is to go with AssertJ for new stuff.

codepitbull (approved):
Small comment on assertions, nothing to block merging.

hivemq-edge/src/main/java/com/hivemq/http/JaxrsHttpServer.java
```diff
@@ -91,6 +93,10 @@ public void startServer() {
                     System.setProperty(MAX_RESP_TIME, "45");
 
                     ResourceConfig resources = resourceConfig == null ? new ResourceConfig() : resourceConfig;
+                    // https://github.com/eclipse-ee4j/jersey/issues/2986
+                    // This server property tells jersey not to resolve relative location
+                    // so that / -> /app/ doesn't get resolved to http://domain/app/.
+                    resources.property(LOCATION_RELATIVE_RESOLUTION_DISABLED, true);
```
marregui, line 99: with this set to true, the response headers will not look like `http://host:port/path/` but rather like `/path/` . This is less information, is this ok?

caoccao: Right. I'll adjust the wording.

marregui (approved):
LGTM

PR #831, opened 2025-02-28
fix(30890): Ensure that topic filters are in the mapping selector

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30890/details/

The PR fixes a bug which prevented the `topic filter` from being available to the mappings.

If no `bridge` was added to the selected entities in the workspace, then `topic filters` would not be processed. This situation is not corrected by ensuring that the `Edge` entity, as the owner of the `topic filters` is automatically added to the list of sources

The PR also improves error message on the schema selector and remove duplicates from the list of tag/topic filters.

### Out-of-scope
- Sources, including the `edge` cannot be removed from the `combiner`. When that is the case, a decision on keeping the `edge` node as a default source must be taken
- Error message has been improved (mentioning the right type of source) but is still lacking enough context. This will be improved, see https://hivemq.kanbanize.com/ctrl_board/57/cards/30750/details/

### Before
HiveMQ-Edge-02-28-2025_03_45_PM

### After
HiveMQ-Edge-02-28-2025_03_43_PM

### DIFF
--- a/hivemq-edge/src/frontend/src/modules/Mappings/components/combiner/CombinedEntitySelect.tsx
+++ b/hivemq-edge/src/frontend/src/modules/Mappings/components/combiner/CombinedEntitySelect.tsx
@@ -5,8 +5,8 @@ import type { GroupBase, MultiValue, OptionBase } from 'chakra-react-select'
 import { Select } from 'chakra-react-select'
 import { Box } from '@chakra-ui/react'
 
-import { DataIdentifierReference } from '@/api/__generated__'
 import type { DomainTag, TopicFilter } from '@/api/__generated__'
+import { DataIdentifierReference } from '@/api/__generated__'
 import type { CombinerContext } from '@/modules/Mappings/types'
 
 interface EntityReferenceSelectProps {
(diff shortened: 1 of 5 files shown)


hivemq-edge/src/frontend/src/modules/Mappings/components/combiner/CombinedEntitySelect.tsx
```diff
@@ -60,7 +60,14 @@ const CombinedEntitySelect: FC<EntityReferenceSelectProps> = ({ id, tags, topicF
...
       }, []) || []
-    )
+
+    return combinedOptions.reduce<EntityOption[]>((acc, current) => {
+      const isAlreadyIn = acc.find((item) => item.value === current.value && item.type === current.type)
+      if (!isAlreadyIn) {
+        return acc.concat([current])
+      }
+      return acc
+    }, [])
```
oli-hivemq, line 70: Looks totally fine, but I'm just wondering whether it'd be easier to use [lodash's uniqWith](https://lodash.com/docs/4.17.15#uniqWith) function.

oli-hivemq (approved):
Left one question, but LGTM 🦣

hivemq-edge/src/frontend/src/modules/Mappings/components/combiner/CombinedEntitySelect.tsx
vanch3d: Possibly. But by a long habit, from when the size of the bundle was too big for any SPA, I've been staying clear of lodash

PR #832, opened 2025-02-28
fix(30892): remove the filter on "supported" properties

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30892/details/

Fix a bug with the "flattening" of properties that prevented nested properties from being displayed in the drag-and-drop editor

Notes
- The reason for the original filter has been lost ! 
- The fix will also impact the southbound mapping editor

### Before
HiveMQ-Edge-02-28-2025_05_10_PM

### After
HiveMQ-Edge-02-28-2025_05_10_PM (1)


oli-hivemq (approved):
LGTM 🦣

PR #841, opened 2025-03-04
fix(30955): Fix combiner's schema fetching for sources

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30955/details/

The PR fixes a small bug that meant schemas from tags were fetched from the wrong API endpoints, resulting in a misleading error message on the UI. 

### Before 
HiveMQ-Edge-03-03-2025_10_21_PM

HiveMQ-Edge-03-04-2025_10_28_AM

### After

HiveMQ-Edge-03-03-2025_10_21_PM (1)

HiveMQ-Edge-03-03-2025_10_22_PM


oli-hivemq (approved):
In terms of styling, I dislike the alignments here. They look wrong and it makes the UI look unprofessional as it's not polished. CleanShot 2025-03-05 at 14 16 03
Eric might have some idea about a quick fix for it?

In terms of code, it LGTM 🦣

PR #842, opened 2025-03-04
fix(30977) Fix the primary key selection in the combiner's mappings

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30977/details/

The PR refactors the UX that allows the primary key of a combiner's mapping to be defined. 

### Design
- The key is selectable from the list of data sources (both tags and topic filters) presented in the mapping editor
- When a selected key is removed from the source list, the primary key is set to `undefined`
- The `required` attributes of the the `RJSF` fomr have been changed to ensure the key is required

### Out-of-scope
- The design of the key selector is flawed in several aspects and will be resolved in subsequent tickets. See https://hivemq.kanbanize.com/ctrl_board/57/cards/30982/details/ and https://hivemq.kanbanize.com/ctrl_board/57/cards/30983/details/


oli-hivemq (approved):
LGTM 🦣

PR #852, opened 2025-03-10
Feature/30759 Initial version of the MTConnect protocol adapter

caoccao:
https://hivemq.kanbanize.com/ctrl_board/57/cards/30759/details/

This is the initial version with the following features.

1. MTConnect protocol comes from project hivemq-mtconnect-protocol.
2. Loose mode is implemented as a fallback for custom protocols that are not compatible with the official protocols.


marregui (approved):
epic work Sam

PR #855, opened 2025-03-18
Error states were incorrectly handled in ADS.

codepitbull:
**Motivation**

Resolves #30336

**Changes**
ADS is now only setting the state to connected if we successfully get the connection.
Before that we didn't really check and because of some internal error in PLC4X ( a type cast in case ADS wasn't able to connect ) we never got to see problems occuring.

### DIFF
--- a/modules/hivemq-edge-module-plc4x/src/main/java/com/hivemq/edge/adapters/plc4x/impl/AbstractPlc4xAdapter.java
+++ b/modules/hivemq-edge-module-plc4x/src/main/java/com/hivemq/edge/adapters/plc4x/impl/AbstractPlc4xAdapter.java
@@ -98,6 +98,7 @@ public abstract class AbstractPlc4xAdapter<T extends Plc4XSpecificAdapterConfig<
     public void poll(final @NotNull BatchPollingInput pollingInput, final @NotNull BatchPollingOutput pollingOutput) {
         final Plc4xConnection<T> tempConnection = connection;
         if (tempConnection != null && tempConnection.isConnected()) {
+            if(!tags.isEmpty()) {
                 tempConnection.read(tags)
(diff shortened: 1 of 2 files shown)


modules/hivemq-edge-module-plc4x/src/main/java/com/hivemq/edge/adapters/plc4x/impl/AbstractPlc4xAdapter.java
```diff
@@ -135,9 +135,28 @@ public void poll(final @NotNull BatchPollingInput pollingInput, final @NotNull B
...
+                        final Plc4xConnection<T> connection = createConnection();
+                        if(connection.isConnected()) {
+                            protocolAdapterState.setConnectionStatus(CONNECTED);
+                            output.startedSuccessfully();
+                        } else {
+                            protocolAdapterState.setConnectionStatus(ERROR);
+                            output.failStart(new Plc4xException("Unable to connect to device"), "Unable to connect to device");
+                        }
+
+                        this.connection = connection;
```
caoccao, line 154: This line is executed even if `output.failStart()` is called. Is this expected behavior?

caoccao (changes requested):
Just one concern in the comment.

modules/hivemq-edge-module-plc4x/src/main/java/com/hivemq/edge/adapters/plc4x/impl/AbstractPlc4xAdapter.java
codepitbull: Thanks for catching :)
Fixed.

caoccao (approved):
LGTM

PR #857, opened 2025-03-19
Rework TagManager to give us more visibility and better handling of concurrent structures

codepitbull:
**Motivation**

We need a way to know the Tags currently being used.
The TagManager is the only part of Edge that currently has that information so we have to expose it.

**Changes**
- Add methods to TagManager to find out which tags are used and by whom
- Rework internals of the TagManager to make better use of existing concurrent structures
- Small rework of the ProtocolAdapterManager to get rid of a sync on a future


hivemq-edge/src/main/java/com/hivemq/protocols/northbound/TagConsumer.java
```diff
@@ -24,4 +24,6 @@
 public interface TagConsumer extends Consumer<List<DataPoint>> {
 
    @NotNull String getTagName();
+
+   @NotNull String consumerName();
```
caoccao, line 28: `getConsumerName()`?

caoccao (approved):
LGTM

caoccao:
I wonder if `CopyOnWriteArraySet` could be a better option in this special case. It allows other threads to hold a snapshot of the collection so that other threads are not blocked. The snapshots also ensure the thread-safety when the tags are being changed.

The drawback is it is internally a `CopyOnWriteArrayList` so that `contains()` is O(N). But in this special case, tags are not update frequently.

hivemq-edge/src/main/java/com/hivemq/protocols/northbound/TagConsumer.java
codepitbull: Fixed

caoccao (approved):
Nice work!

PR #861, opened 2025-03-19
fix(30987): Add styling for the primary key in the tabular summary

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30987/details/

The PR adds visual cues to the data point, `tag` or `topic filter`, that is used as a `primary key` for the listed mapping. Only one of the presented data points must have the primary cues.

In this first iteration, cues are a `key` icon in a tag wrapping the data tag. The `key` icon has also been added to the selector in the main editor, for contextualization 

The PR also fixes an accessibility issue (label of the table)

### Before
HiveMQ-Edge-03-19-2025_05_23_PM

### After 
HiveMQ-Edge-03-20-2025_03_00_PM

HiveMQ-Edge-03-20-2025_03_01_PM


oli-hivemq:
Will the aria-label be enough for people to understand the meaning of the wider borders?

oli-hivemq:
What about something like this?
CleanShot 2025-03-20 at 08 29 49

oli-hivemq (approved):
Left a comment/suggestion - feel free to ignore.
Approving anyway to move fast.

PR #862, opened 2025-03-20
Implement hot-reloading

codepitbull:
**Motivation**

Resolves #31276

**Changes**

- Get rid of old writing/reading logic, especially the proxies
- make all services which are capable of reloading reloaded from the filesystem


hivemq-edge/src/main/java/com/hivemq/configuration/reader/BridgeExtractor.java
```diff
@@ -291,10 +328,12 @@ private static List<CustomUserProperty> convertCustomUserProperties(
...
     public void sync(final @NotNull HiveMQConfigEntity entity) {
-        List<MqttBridge> liveBridges = bridgeConfigurationService.getBridges();
-        List<MqttBridgeEntity> newList = liveBridges.stream().map(this::uncovert).collect(Collectors.toList());
-        entity.getBridgeConfig().clear();
-        entity.getBridgeConfig().addAll(newList);
+        var tmpBridges = bridgeEntities;
+        if(tmpBridges != null) {
+            List<MqttBridgeEntity> newList = tmpBridges.stream().map(this::uncovert).collect(Collectors.toList());
+            entity.getBridgeConfig().clear();
+            entity.getBridgeConfig().addAll(newList);
```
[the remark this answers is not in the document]
codepitbull: You are prefectly right.
Added a synchronized on the method.

hivemq-edge/src/main/java/com/hivemq/configuration/reader/BridgeExtractor.java
```diff
@@ -37,52 +36,92 @@
...
+                .build();
+
+        notifyConsumer();
+        configFileReaderWriter.writeConfigWithSync();
+    }
+
+    public @NotNull List<MqttBridge> getBridges() {
+        return new ImmutableList.Builder<MqttBridge>()
+                .addAll(bridgeEntities)
+                .build();
```
caoccao, line 87: If `bridgeEntities` is already immutable, why do we create another immutable list?

hivemq-edge/src/main/java/com/hivemq/configuration/reader/BridgeExtractor.java
```diff
@@ -37,52 +36,92 @@
...
+    }
+
+    public synchronized void removeBridge(final @NotNull String id) {
+        bridgeEntities = bridgeEntities.stream().filter(entry -> !entry.getId().equals(id)).toList();
```
caoccao, line 91: `bridgeEntities` is not immutable here. Why do we have the inconsistency here?

hivemq-edge/src/main/java/com/hivemq/bridge/BridgeService.java
```diff
@@ -248,6 +254,14 @@ public boolean restartBridge(
         return false;
     }
 
+    public @NotNull List<MqttBridge> getBridges() {
+        final var tmpBridges = bridges;
+        if(tmpBridges != null) {
+            return tmpBridges;
+        }
+        return List.of();
```
caoccao, line 262: If all bridges are removed temporarily, does that imply slight data loss?

caoccao (approved):
I have only a few small concerns. Not blocking.

hivemq-edge/src/main/java/com/hivemq/configuration/reader/BridgeExtractor.java
codepitbull: The method is synchronized which prevents modification. When we change to a different list type we have to reconsider.
Or do you mean the toList? That returns an unmodifiable list.

codepitbull: To clarify about synchronized:
All modifiying methods are synchronized, so that protects the collection.

caoccao: In this class, sometimes `bridgeEntities` is mutable, sometimes `bridgeEntities` is immutable. I'm referring to that type inconsistency. I didn't intend to question the synchronization.

PR #864, opened 2025-03-21
fix(30993): Improving toast management

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30993/details/

The PR fixes the `Designer` publish toasts, preventing duplicates to be shown when repeatedly pressing "publish".

The PR also ensures that every toasts in Edge are using the same base config.

### Before
screenshot-localhost_3000-2025_01_29-14_39_55

### After 
HiveMQ-Edge-03-21-2025_04_30_PM


oli-hivemq (approved):
LGTM 🦣

PR #868, opened 2025-03-24
fix(30741): Add checks for valid combiner creation

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30741/details/

The PR adds checks to the creation of a new combiner:
- Every selected adapter source must have the `COMBINE` capability. If not, the combiner will be created by excluding those without. 
- If a combiner using the same set of sources already exists, creating a new one will be prevented. The user will be redirected to the combiner's configuration panel, and adding mappings to this one will be suggested. 

### Before 
HiveMQ-Edge-03-24-2025_03_45

### After 
HiveMQ-Edge-03-24-2025_03_35_AM

HiveMQ-Edge-03-24-2025_03_39_AM (1)

### DIFF
--- a/hivemq-edge/src/frontend/src/modules/Workspace/utils/combiner.utils.ts
+++ b/hivemq-edge/src/frontend/src/modules/Workspace/utils/combiner.utils.ts
@@ -6,12 +6,6 @@ export const arrayContains = <T extends object>(container: T[], o: T) => {
   return container.some((c) => objectsEqual<T>(c, o))
 }
 
-// export const combinerWithSameEntities = (entities: EntityReference[]) => (combiner: Combiner) => {
-//   return entities.every((entity) => arrayContains<EntityReference>(combiner.sources.items, entity))
-//     ? combiner
-//     : undefined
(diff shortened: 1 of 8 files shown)


hivemq-edge/src/frontend/src/modules/Workspace/utils/combiner.utils.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const arrayContains = <T extends object>(container: T[], o: T) => {
+  return container.some((c) => objectsEqual<T>(c, o))
+}
+
+// export const combinerWithSameEntities = (entities: EntityReference[]) => (combiner: Combiner) => {
+//   return entities.every((entity) => arrayContains<EntityReference>(combiner.sources.items, entity))
+//     ? combiner
+//     : undefined
+// }
```
oli-hivemq, line 13: is it left here on purpose or should it be cleaned up?

oli-hivemq (approved):
LGTM 🦣

PR #869, opened 2025-03-24
fix(30538): Improve the combiner visual identity in the workspace

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30538/details/

The PR improves the visual identity of the `combiner` node in the `workspace`, in particular by adding the `topic` flags of the mappings to the node, similar to every other entity. 

The `edge` node is also refactored to display the `topic filters` flags on the node, as well as the instance name (not yet implemented, replaced by the generic product name)

### Before 
HiveMQ-Edge-03-24-2025_03_49

### After'
HiveMQ-Edge-03-24-2025_03_51


oli-hivemq (approved):
LGTM 🦣

PR #876, opened 2025-03-27
Fix/31448 opcua timestamp tagname

codepitbull:
**Motivation**

Resolves #31448

**Changes**

Resolves an issue where northbound messages for OPC UA tags would not honor the includTimestamp/includeTagName settings.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/protocols/northbound/NorthboundConsumerFactory.java
+++ b/hivemq-edge/src/main/java/com/hivemq/protocols/northbound/NorthboundConsumerFactory.java
@@ -18,6 +18,7 @@ package com.hivemq.protocols.northbound;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.hivemq.adapter.sdk.api.config.PollingContext;
 import com.hivemq.adapter.sdk.api.events.EventService;
+import com.hivemq.adapter.sdk.api.factories.DataPointFactory;
 import com.hivemq.adapter.sdk.api.services.ProtocolAdapterMetricsService;
 import com.hivemq.edge.modules.adapters.impl.ProtocolAdapterPublishServiceImpl;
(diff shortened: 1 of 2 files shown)


hivemq-edge/src/main/java/com/hivemq/configuration/reader/UnsExtractor.java
```diff
@@ -0,0 +1,93 @@
...
+    @Override
+    public void sync(final @NotNull HiveMQConfigEntity entity){
+        final ISA95 isa95 = config;
+        entity.getUns().getIsa95().setEnabled(isa95.isEnabled());
+        entity.getUns().getIsa95().setPrefixAllTopics(isa95.isPrefixAllTopics());
+        entity.getUns().getIsa95().setArea(isa95.getArea());
+        entity.getUns().getIsa95().setEnterprise(isa95.getEnterprise());
+        entity.getUns().getIsa95().setSite(isa95.getSite());
+        entity.getUns().getIsa95().setWorkCell(isa95.getWorkCell());
+        entity.getUns().getIsa95().setProductionLine(isa95.getProductionLine());
```
caoccao, line 67: Could we save `entity.getUns().getIsa95()` to a temp var?

Or, what if `setISA95()` is called during the `sync()`? Is that possible?

caoccao (approved):
Just one concern, but not blocking.

hivemq-edge/src/main/java/com/hivemq/configuration/reader/UnsExtractor.java
codepitbull: Good point.
Moved things into a temp variable and made the method synchronized.

caoccao: My personal practice is such feature usually belongs to `Isa95`, so I would create a method `Isa95#copyFrom(final Isa95 isa95)`.

PR #877, opened 2025-03-27
feat(30574): Add source schemas merging to the combiner editor

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30574/details/

The PR introduces an experimental sources merging process to "infer" a destination schema from the source schemas. 

### Design
- the "infer a schema" doesn't merge the underlying JSON-Schema of the selected source; it merges the "flattened" list of properties displayed on the "Source schemas" 
- To be activated, the sources must contain at least one valid schema, i.e. as represented by a list of properties, with their type, name, description and examples 
- The merge process creates a new list with all the nested properties in a single content
- The order of the properties is defined by the order of the source
- The top-level properties ("roots") are systematically renamed, to prevent override by duplication. The pattern for renaming is a prefix (type of source and index) to the original name
- When "Generate" is actioned, the new list of properties is used to create a new JSON-Schema, which is added to the form payload, ready for rendering and mapping.

### Out-of-scope
- Users MUST be allowed to edit the default renaming of duplicated property names. This will be added in a subsequent ticket. 
- Additional information on the schema (such as title, and description) SHOULD be editable as well. This will be added in a subsequent ticket. 

### Before
HiveMQ-Edge-03-27-2025_18_25

### After
HiveMQ-Edge-03-27-2025_15_41

HiveMQ-Edge-03-27-2025_15_41 (1)

HiveMQ-Edge-03-27-2025_15_42


hivemq-edge/src/frontend/src/modules/Mappings/combiner/DestinationSchemaLoader.tsx
```diff
@@ -0,0 +1,164 @@
...
+  ModalCloseButton,
+  VStack,
+  useDisclosure,
+} from '@chakra-ui/react'
+
+import type { DataCombining, Instruction } from '@/api/__generated__'
+import ErrorMessage from '@/components/ErrorMessage'
+import { MappingInstructionList } from '@/components/rjsf/MqttTransformation/components/MappingInstructionList'
+import type { FlatJSONSchema7 } from '@/components/rjsf/MqttTransformation/utils/json-schema.utils'
+import { getSchemaFromPropertyList } from '@/components/rjsf/MqttTransformation/utils/json-schema.utils'
```
oli-hivemq, line 23: Why not combining these two?
`import { getSchemaFromPropertyList, type FlatJSONSchema7 } from '@/components/rjsf/MqttTransformation/utils/json-schema.utils'`

hivemq-edge/src/frontend/src/modules/Mappings/utils/combining.utils.ts
```diff
@@ -0,0 +1,98 @@
+import type { UseQueryResult } from '@tanstack/react-query'
+
+import type { DataCombining, DomainTag, EntityReference, JsonNode } from '@/api/__generated__'
+import { DataIdentifierReference, type TopicFilter } from '@/api/__generated__'
```
oli-hivemq, line 4: is it best practice to separate types and the rest?
Actually on this occasion it's a real mix, so you could merge:
```
import {
  DataIdentifierReference,
  type DataCombining,
  type DomainTag,
  type EntityReference,
  type JsonNode,
  type TopicFilter
} from '@/api/__generated__'
```

hivemq-edge/src/frontend/src/modules/Mappings/utils/combining.utils.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  const selectedReferences = allDataReferences?.filter((dataReference) => indexes.includes(dataReference.id)) || []
+  return selectedReferences.reduce<DataReference[]>((acc, current) => {
+    const isAlreadyIn = acc.find((item) => item.id === current.id && item.type === current.type)
+    if (!isAlreadyIn) {
+      return acc.concat([current])
+    }
+    return acc
+  }, [])
+}
```
oli-hivemq, line 72: Using `find` all the time doesn't seem super efficient.

Two potential alternatives:
1. more readable: always concatenating and then deduplicating
```
import { uniqBy } from 'lodash'

const selectedReferences = uniqBy(
  allDataReferences?.filter((dataReference) =>
    indexes.includes(dataReference.id)
  ) || [],
  (item) => `${item.id}_${item.type}`
)
```

2. more efficient: using a Set or Map to track ids + types already seen
```
export const getFilteredDataReferences = (formData?: DataCombining, formContext?: CombinerContext) => {
  const tags = formData?.sources?.tags || []
  const topicFilters = formData?.sources?.topicFilters || []
  const indexes = new Set([...tags, ...topicFilters])

  const allDataReferences = getDataReference(formContext)

  const seen = new Set<string>()

  const selectedReferences = allDataReferences?.filter((dataReference) =>
    indexes.has(dataReference.id)
  ).filter((dataReference) => {
    const key = `${dataReference.id}_${dataReference.type}`
    if (seen.has(key)) {
      return false
    }
    seen.add(key)
    return true
  }) || []

  return selectedReferences
}
```

hivemq-edge/src/frontend/src/modules/Mappings/combiner/SchemaMerger.tsx
```diff
@@ -0,0 +1,131 @@
...
+  VStack,
+} from '@chakra-ui/react'
+import { useTranslation } from 'react-i18next'
+
+import type { DataCombining } from '@/api/__generated__'
+import { DataIdentifierReference } from '@/api/__generated__'
+import { useGetCombinedDataSchemas } from '@/api/hooks/useDomainModel/useGetCombinedDataSchemas'
+import PropertyItem from '@/components/rjsf/MqttTransformation/components/schema/PropertyItem'
+import type { FlatJSONSchema7 } from '@/components/rjsf/MqttTransformation/utils/json-schema.utils'
+import { getPropertyListFrom } from '@/components/rjsf/MqttTransformation/utils/json-schema.utils'
```
oli-hivemq, line 24: Merging these import lines?

hivemq-edge/src/frontend/src/modules/Mappings/combiner/SchemaMerger.tsx
```diff
@@ -0,0 +1,131 @@
...
+            if (property.path[index]) {
+              property.path[index] = newTitle
+            }
+          })
+        }
+      })
+      acc.push(...properties)
+      return acc
+    }, [])
+  }, [formData?.sources?.tags, formData?.sources?.topicFilters, references, schemaQueries])
```
oli-hivemq, line 82: Repeating `.findIndex()` in every `.forEach()` loop is inefficient.
Instead, you could precompute a Map of tag & topicFilter IDs -> index once outside the .reduce():
```
const tagIndexMap = new Map(formData?.sources?.tags?.map((tag, index) => [tag, index]))
const topicFilterIndexMap = new Map(formData?.sources?.topicFilters?.map((filter, index) => [filter, index]))
```

Then you can replace `.findIndex()` by `.get(reference.id)` e.g. `tagIndexMap.get(reference.id)`

oli-hivemq (approved):
I left minor comments about imports, and more important comments about inefficiencies of using `.findIndex` in loops, with suggestions.
Approving to move fast.

hivemq-edge/src/frontend/src/modules/Mappings/utils/combining.utils.ts
vanch3d: Never checked the efficiency of find. I'm using a similar pattern in quite a few places so I'll see to get a reusable piece of code like your proposal (I'm not using lodash)

PR #884, opened 2025-04-02
fix(31710): Prevent crash with edge status

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/31710/details/

The PR adds missing optional chaining  on accessing a data payload that was also not properly typed. Enough said


oli-hivemq (approved):
LGTM 🦣

PR #893, opened 2025-04-07
refactor(31833): Upgrade RJSF and update templates

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/31833/details/

The PR upgrades `RJSF` to its latest version and update some of the temapltes (in particular to add consistency for testing purpose)


hivemq-edge/src/frontend/src/components/rjsf/Fields/CompactArrayField.tsx
```diff
@@ -1,6 +1,6 @@
 import type { FC } from 'react'
 import type { FieldProps } from '@rjsf/utils'
-import type { RJSFSchema } from '@rjsf/utils/src/types.ts'
+import type { RJSFSchema } from '@rjsf/utils'
```
oli-hivemq, line 3: nit: duplicate imports

hivemq-edge/src/frontend/src/components/rjsf/utils/getChakra.ts
```diff
@@ -0,0 +1,32 @@
...
+     * Normally, it prevents ChakraProps from being passed to the DOM.
+     * In this case we just want to delete the unknown props. So we flip the boolean.
+     */
+    if (shouldForwardProp(key)) {
+      delete (chakraProps as never)[key]
+    }
+  })
+
+  return chakraProps
+}
```
oli-hivemq, line 32: I find this a bit confusing. For example `shouldForwardProp` is used to remove props.
And delete in a loop might not be super efficient.

Proposal (not tested):
```ts
export function getChakra({ uiSchema = {} }: GetChakraProps): ChakraProps {
  const rawChakraProps = uiSchema['ui:options']?.chakra ?? {}

  const cleanedChakraProps = Object.entries(rawChakraProps).reduce((acc, [key, value]) => {
    // Only include props that should NOT be forwarded to the DOM
    if (!shouldForwardProp(key)) {
      acc[key] = value
    }
    return acc
  }, {} as ChakraProps)

  return cleanedChakraProps
}
```

oli-hivemq (approved):
Left a few comments.
Approving to move fast

hivemq-edge/src/frontend/src/components/rjsf/utils/getChakra.ts
vanch3d: The file is a copy from `RJSF` (I've added a comment and q code coverage instruction).
I'm flagging the lack of export on the library. Will be removed once fixed

PR #894, opened 2025-04-07
Added modules as compileOnly dependency to reliably reload changes when testing

codepitbull:
When hivemq-edge is included in hivemq-edge-test changes to the protocol adapter modules wouldn't trigger the module to be rebuilt.

Only using includeBuild results in modules not being rebuilt when hivemq-edge is included in other builds.
This is due to a not-declared dependency between the modules and edge.

I made the protocol adapter modules compileOnly depdencies instead of only doing includeBuild which gives gradle the required information to react to changes in the modules.


marregui (approved):
LGTM

PR #896, opened 2025-04-07
fix(31771): Improve the toolbar when multiple nodes are selected

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/31771/details/

The PR improves. the toolbar in the workspace, to deal with situations when multiple nodes are selected.

### Design
- As before, only one toolbar is displayed on the workspace
- When multiple nodes are selected, the toolbar will be displayed on the topmost node, ensuring a better experience 
- The `title` of the toolbar will show either the name of the component (single selection) or a text such as "3 elements selected" (multiple selection)
- Only the CTAs for selection-based commands are shown on the toolbar: `group` and `combine`

### Before
HiveMQ-Edge-04-07-2025_11_43

### After
HiveMQ-Edge-04-07-2025_11_42

### DIFF
--- a/hivemq-edge/src/frontend/src/modules/Workspace/components/nodes/ContextualToolbar.tsx
+++ b/hivemq-edge/src/frontend/src/modules/Workspace/components/nodes/ContextualToolbar.tsx
@@ -58,7 +58,20 @@ const ContextualToolbar: FC<ContextualToolbarProps> = ({
   const navigate = useNavigate()
   const { fitView } = useReactFlow()
 
-  const selectedNodes = nodes.filter((node) => node.selected)
+  const selectedNodes = useMemo(() => {
+    return nodes.filter((node) => node.selected)
+  }, [nodes])
+
+  const topSelectedNode = useMemo(() => {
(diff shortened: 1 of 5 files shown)


hivemq-edge/src/frontend/src/modules/Workspace/components/nodes/ContextualToolbar.tsx
```diff
@@ -58,7 +58,20 @@ const ContextualToolbar: FC<ContextualToolbarProps> = ({
...
   const { fitView } = useReactFlow()
 
-  const selectedNodes = nodes.filter((node) => node.selected)
+  const selectedNodes = useMemo(() => {
+    return nodes.filter((node) => node.selected)
+  }, [nodes])
+
+  const topSelectedNode = useMemo(() => {
+    const [firstNode] = nodes
+      .filter((node) => node.selected)
```
oli-hivemq, line 67: Why not reusing `selectedNodes` here instead of these 2 lines?

oli-hivemq (approved):
Left a simple comment.

PR #903, opened 2025-04-08
Fixed cases where processing stopped because of a broken consumer

codepitbull:
**Motivation**

Resolves #31949

**Changes**
When a new tag value came in we iterated over all consumers to inform them about the new value.
A broken consumer could throw an exception which would stop the iteration.

The fix is to catch and report the errors and then continue processing.


caoccao (approved):
The fix looks good. I wonder in the long term silently log a warning might not be a good solution.

codepitbull:
> The fix looks good. I wonder in the long term silently log a warning might not be a good solution.

Good point, I raised it to error.

PR #906, opened 2025-04-10
Bugfix/packet id is returned twice

codepitbull:
**Motivation**

Resolves 31972

**Changes**
Fixing: The broker returns every packet ID twice when the publish flow is complete. Therefore, the same packet ID can be assigned to two publishes. The OrderedTopicService maps the packet ID to a future, which is complete when the publish flow is complete. When two publishes have the same packet ID, the map entry will be replaced, and one of the futures will never be complete.


caoccao (approved):
It seems most of the changes are refactoring changes. The dedup is to only route the message to the first publish, right? Looks good to me.

PR #922, opened 2025-04-14
test: check the workflow syntax

blacksd:
**Motivation**

Test condition to skip running the `check` workflow when pushed to frontend's main path


blacksd:
path filtering and required checks are a bad match

PR #923, opened 2025-04-15
fix(31362): Add feedback to schema validation for destination

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/31362/details/

The PR adds the missing error messages when the validation of the schema for the destination topic fails. 

The translations for the validation messages have been improved, in particular mentioning the correct Integration point (tag, topic or topic filter) for which the schema is being validated

### Out-of-scope
- The payloads that contains inbound schemas (tag and topic filter) are not consistent in shape and are not maintaining connection with the type of its integration point. The result is a series of hack in the frontend code. The issue should be fixed in subsequent ticket (see https://hivemq.kanbanize.com/ctrl_board/57/cards/30744/details/)
### After

### DIFF
--- a/hivemq-edge/src/frontend/src/locales/en/translation.json
+++ b/hivemq-edge/src/frontend/src/locales/en/translation.json
@@ -1148,20 +1148,7 @@
     },
     "error": {
       "loading": "We cannot load the topic filters. Please try again later.",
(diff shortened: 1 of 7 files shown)


oli-hivemq (approved):
LGTM 🦣

PR #927, opened 2025-04-17
fix(32339): fix bug with react-flow upgrade

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/32339/details/ 

The PR fixes a bug that crept in out of the latest major upgrade of `react-flow`, preventing nodes to be selected

### End


oli-hivemq (approved):
LGTM 🦣

PR #933, opened 2025-04-22
fix(32398): Fix the rendering of combiner events in the Event Log

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/32398/details/

The PR fixes the rendering of `Combiner` sources in the `Event Log`, making events consistent with other sources like `Adapter` or `Bridge`

### Design
- If the `Combiner` doesn't exist anymore, the `id` is simply rendered as it is
- If the `Combiner` still exists, the `name` (if defined) is displayed with the combiner icon and a link that redirects to the `Combiner` configuration panel on the `Workspace`

### Out-of-scope
- Filtering sources is still based on the internal values received by the API, i.e. the `id`. It creates a discrepancy between displayed and filtered values. This will be fixed in a future ticket.


oli-hivemq (approved):
LGTM 🦣

PR #937, opened 2025-04-23
Fix multiple inconsistencies for combiner events

codepitbull:
**Motivation**

Resolves #31411

**Changes**
- Use Combiner instead of Data Combining in event messages
- Add additional sources, especially COMBINER
- Update combiner messages to use the name when available
- Create proper events to reflect what is actually happening.


caoccao (approved):
LGTM

ext/hivemq-edge-openapi-2024.1.yaml
```diff
@@ -2094,6 +2094,9 @@ components:
             - ADAPTER_TYPE
             - EVENT
             - USER
+            - DATA_COMBINING
```
vanch3d, line 2097: I assumed this one is kept for backward compatibility

codepitbull: yes, it's currently not used but was there from the beginning so we keep it for now.

codepitbull: And you were completely right ... that change ended up in the wrong file.

PR #938, opened 2025-04-24
fix(31583): Add metadata to the schema abstraction

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/31583/details/

The PR fixes the content of the abstraction of `schemas` in Edge. They are represented as a "flat" list of properties, ready for presentation and manipulation (e.g. drag-and-drop for mappings). 

The current implementation was restricted to a pragmatic set of attributes of properties (title, description, examples). 

The PR corrects the situation and ensure that every supported metadata are included. It includes  `required`, `min` and `max` for number, `maxitem` for arrays, etc. 

### Out-of-scope
- the metadata are exported when the schemas are transformed but they are not yet accessible to end-users. This will be done in a subsequent ticket
- The introduction of properties metadata to the mapping opens the possibility for much wider validation (e.g. ensuring min and max range overlap). This will be dealt with in subsequent tickets

### After


hivemq-edge/src/frontend/src/modules/Mappings/combiner/DataCombiningEditorField.tsx
```diff
@@ -51,195 +50,193 @@ export const DataCombiningEditorField: FC<FieldProps<DataCombining, RJSFSchema,
...
+              const tag: string[] = []
+              const filter: string[] = []
+
+              let isPrimary = false
+              values.forEach((entity) => {
+                if (entity.type === DataIdentifierReference.type.TAG) tag.push(entity.value)
+                if (entity.type === DataIdentifierReference.type.TOPIC_FILTER) filter.push(entity.value)
+                if (formData.sources.primary?.type === entity.type && formData.sources.primary?.id === entity.value)
+                  isPrimary = true
+              })
```
oli-hivemq, line 108: Better alternative (clearer intent, building arrays instead of pushing into pre-declared ones with `forEach`):
```ts
const tags = values
  .filter(entity => entity.type === DataIdentifierReference.type.TAG)
  .map(entity => entity.value)

const filters = values
  .filter(entity => entity.type === DataIdentifierReference.type.TOPIC_FILTER)
  .map(entity => entity.value)

const isPrimary = values.some(entity =>
  formData.sources.primary?.type === entity.type &&
  formData.sources.primary?.id === entity.value
)
```

oli-hivemq (approved):
Left a suggestion to avoid using `forEach`

PR #948, opened 2025-04-28
DataCombinerManager was incorrectly logging a deletiong error

codepitbull:
**Motivation**

Resolves #32561

**Changes**
- Fix DataCombinerManager incorrectly logging a failure during deletion
- Add tests for the DataCombinerManager

### DIFF
--- a/hivemq-edge/src/test/java/com/hivemq/combining/DataCombinerManagerTest.java
+++ b/hivemq-edge/src/test/java/com/hivemq/combining/DataCombinerManagerTest.java
@@ -44,20 +44,7 @@ public class DataCombinerManagerTest {
     @Test
     public void test_add() {
         var now = System.currentTimeMillis();
-        var eventService = new EventServiceDelegateImpl(new EventStore() {
-
-            private final @NotNull List<Event> events = new ArrayList<>();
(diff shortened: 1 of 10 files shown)


hivemq-edge/src/test/java/com/hivemq/combining/DataCombinerManagerTest.java
```diff
@@ -1,112 +1,119 @@
-/*
```
caoccao, line 1: Why is the copyright removed?

caoccao (approved):
It looks good. But I just hold one concern.

hivemq-edge/src/test/java/com/hivemq/combining/DataCombinerManagerTest.java
codepitbull: Argh, that's a new test, I replaced the old one because the naming was misleading and forgot to add the header.

codepitbull: Fixed

caoccao (approved):
LGTM

PR #950, opened 2025-04-29
Fix/30358 remove legacyconfigsupport

codepitbull:
**Motivation**

Resolves #30358

**Changes**
Remove support for Legacy Configurations.


caoccao (changes requested):
The removal looks good to me. But the `frontend/dist` should be excluded, right?

caoccao (approved):
LGTM

PR #951, opened 2025-04-30
Fix/frontend location

codepitbull:
**Motivation**

Separate frontend/backend

**Changes**

- Move frontend into a dedicated top-level directory
- Fix git actions

**card**: https://hivemq.kanbanize.com/ctrl_board/57/cards/32593/details/

### DIFF
--- a/hivemq-edge-frontend/README.md
+++ b/hivemq-edge-frontend/README.md
@@ -140,4 +140,3 @@ Together. they contain - and enforce - parts of the testing pyramid for frontend
 ### Percy - Visual testing
 
 ### Lighthouse - Performance testing
-


.github/workflows/check.yml
```diff
@@ -15,7 +13,21 @@ jobs:
...
-      - name: Checkout Extension SDK
+        with:
+          fetch-depth: 0
+          
+      - name: Check for frontend changes
+        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.1
+        id: changes
+        with:
+          filters: |
+            edge:
```
vanch3d, line 24: I think I understand the syntax. 
Wouldn't it be better to have exclusion rather than inclusion, e.g.
```
-  !hivemq-edge-frontend/**
```

vanch3d (approved):
small  question
LGTM

caoccao (approved):
LGTM

.github/workflows/check.yml
codepitbull: Not supported by that action :(

PR #964, opened 2025-05-09
Exclude missing tasks for frontend and add idea plugin

codepitbull:
**Motivation**

We had issues with missing tasks and classes not being built correctly when run in idea.

**Changes**
- Exclude frontend from task generation where requried
- Add idea module


caoccao (approved):
LGTM

PR #971, opened 2025-05-13
chore(30540): Improving TSConfig for Cypress and Linting

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30540/details/

The PR fixes some of the problems with Cypress + Typescript configuration (missing code completions) and with the upgrade of Eslint. 

### Out-of-scope
- The Cypress custom commands are still not recognised by the IDE outside the `./cypress/` directory. The globals are not recognised.


hivemq-edge-frontend/eslint.config.mjs
```diff
@@ -9,16 +10,21 @@ import pluginQuery from '@tanstack/eslint-plugin-query'
 
 export default tsEslint.config(
   { ignores: ['dist', '**/__generated__/*'] },
-  eslintConfigPrettier,
+  reactPlugin.configs.flat.recommended, // This is not a plugin object, but a shareable config object
+  reactPlugin.configs.flat['jsx-runtime'], // Add this if you are using React 17+
```
oli-hivemq, line 14: nit: the way the comment is phrased indicates that some AI tool helped there, which is not a problem, but I'd just update it to `// needed when using React 17+` or similar

hivemq-edge-frontend/cypress/support/component.ts
```diff
@@ -0,0 +1,28 @@
...
+// This is a great place to put global configuration and
+// behavior that modifies Cypress.
+//
+// You can change the location of this file or turn off
+// automatically serving support files with the
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
```
oli-hivemq, line 14: nit: I would rephrase this to make it your own.
Example:
```
// ***********************************************************
// This component.ts runs before any Cypress component tests.
// Global settings and custom behaviors for Cypress go here.
// By default, Cypress loads this file automatically before test files.
// To use a different support file or disable it, adjust the `supportFile` option in cypress.config.
// For full configuration details, see https://on.cypress.io/configuration
// ***********************************************************
```

hivemq-edge-frontend/src/types/cypress.d.ts
```diff
@@ -0,0 +1,27 @@
+/// <reference types="cypress" />
```
oli-hivemq, line 1: Why is there `hivemq-edge-frontend/src/types/cypress.d.ts` and `hivemq-edge-frontend/cypress/cypress.d.ts`?
If there's a valid reason, I would document it for your future self so you don't have to remember the long forgotten reasons.

oli-hivemq (approved):
Left a few comments and a question for you

hivemq-edge-frontend/eslint.config.mjs
vanch3d: It's a copy and paste from their doc. But same effect

hivemq-edge-frontend/src/types/cypress.d.ts
vanch3d: That's the crux of the problem. I cannot get Typescript to use the types from Cypress. 
And then the types in the src are not found either.
Or there is another problem: the commands are still not found by the IDE
giving up at that stage

PR #973, opened 2025-05-14
feat(32892): Confirm the deletion of all versions of a script or schema

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/32892/details/

The PR refactors the delete modal for `Script` and `Schema`, highlighting the fact that EVERY version of the selected resource will be deleted 

### Out-of-scope
- The display of the resources, as a "bundle" of several versions, will be dealt with in a subsequent ticket. See https://hivemq.kanbanize.com/ctrl_board/57/cards/32894/details/
- The confirmation modal needs refactoring to handle different types of content

### Before
HiveMQ-Edge-05-14-2025_15_36 (1)

### After 
HiveMQ-Edge-05-14-2025_15_36

### DIFF
--- a/hivemq-edge-frontend/src/extensions/datahub/components/pages/DataHubListings.tsx
+++ b/hivemq-edge-frontend/src/extensions/datahub/components/pages/DataHubListings.tsx
@@ -94,8 +94,9 @@ const DataHubListings: FC = () => {
         isOpen={isConfirmDeleteOpen}
         onClose={handleConfirmOnClose}
         onSubmit={handleConfirmOnSubmit}
-        message={t('Listings.modal.delete.message')}
-        header={t('Listings.modal.delete.header')}
+        header={t('Listings.modal.delete.header', { context: deleteItem?.type })}
+        message={t('Listings.modal.delete.message', { context: deleteItem?.type })}
+        prompt={t('Listings.modal.delete.prompt')}
       />
     </Tabs>
   )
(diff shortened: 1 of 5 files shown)


hivemq-edge-frontend/src/extensions/datahub/components/pages/DataHubListings.tsx
```diff
@@ -94,8 +94,9 @@ const DataHubListings: FC = () => {
         isOpen={isConfirmDeleteOpen}
         onClose={handleConfirmOnClose}
         onSubmit={handleConfirmOnSubmit}
-        message={t('Listings.modal.delete.message')}
-        header={t('Listings.modal.delete.header')}
+        header={t('Listings.modal.delete.header', { context: deleteItem?.type })}
+        message={t('Listings.modal.delete.message', { context: deleteItem?.type })}
+        prompt={t('Listings.modal.delete.prompt')}
```
oli-hivemq, line 99: This seems to me a bit overcomplicated and not very flexible if you have to always pass all these parameters.
By looking at the output, I would have assumed you would have passed an array of strings and used a for loop.
e.g.
```ts
const lines = [
  'Schemas are internally stored as a bundle; every version of the schema will be deleted.',
  "You can't undo this action afterward. Are you sure?"
]

{ lines.map((line: string) => (
  <Box key={line}>
    <Text>{line}</Text>
  </Box>
)}
```

I'm actually doing something similar in Cloud for a custom banner that can be enabled/disabled and customised via LaunchDarkly. The content of the banner (`description`) can be a string or an array of (Markdown) strings.

Your code reminded me of that for some reason.
CleanShot 2025-05-14 at 22 41 15

oli-hivemq (approved):
Left a comment (that can be ignored). Otherwise LGTM 🦣

hivemq-edge-frontend/src/extensions/datahub/components/pages/DataHubListings.tsx
vanch3d: That's a perfect idea. Better than the result of incremental changes bloating the code

PR #974, opened 2025-05-15
Added caching for gradle builds

codepitbull:
**Motivation**

I tested caching with the metrics service and it provided some solid improvements.

**Changes**

Replaced the generic gradle action with one that includes caching (burrunan/gradle-cache-action)


caoccao (approved):
LGTM

PR #976, opened 2025-05-16
feat(32894): Handle list of Resource versions as a "bundle"

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/32894/details/

The PR addresses an issue with resources (`Script` and `Schemas`) being handled as a single item with a list of continuous version associated with it.

The PR adds expandable rows for multi-version items:
- The main items will indicate how many version it includes 
- It had an extra CTA in the actions, allowing versions to be expanded or collapsed
- On expanding, individual version will be listed below the main item, with all the same information (and an extra indentation to mark status)
- Main item can only be deleted while individual versions can only be downloaded 

Single version resources remain displayed as before. 

### Design 
- The primary UX remains a table and behave like a table in every behaviour and per accessibility rules 
- In particular, versions are not tree elements but sub rows 
- They remain attached to the top-level row but are otherwise affected by every UX element of the table: filtering, ordering, paginating, 

The PR also fixes a bug with the use of `cy.clock` and `cy.intercept`

### After

### DIFF
--- a/hivemq-edge-frontend/src/components/PaginatedTable/PaginatedTable.tsx
+++ b/hivemq-edge-frontend/src/components/PaginatedTable/PaginatedTable.tsx
@@ -1,10 +1,11 @@
 import type { CSSProperties } from 'react'
 import { useState } from 'react'
 import { useTranslation } from 'react-i18next'
-import type { ColumnDef, ColumnFiltersState, Row } from '@tanstack/react-table'
+import type { ColumnDef, ColumnFiltersState, ExpandedState, Row } from '@tanstack/react-table'
 import {
   flexRender,
   getCoreRowModel,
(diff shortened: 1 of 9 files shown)


oli-hivemq:
CleanShot 2025-05-21 at 12 42 44

oli-hivemq (changes requested):
No comment on the code, it looks good. But I left a comment on the user experience.
Expanding or collapsing a row doesn't feel like a standard action. Hence I do not believe it should be a button at the end of the row. It shouldn't look and feel like the other actions you have there.
I made a suggestion. Whether you decide to go with it or not, I still feel that the collapse/expand icon doesn't belong in a button.

PR #977, opened 2025-05-16
fix: Exclude frontend from task check

caoccao:
https://hivemq.kanbanize.com/ctrl_board/57/cards/33183/details/


marregui (approved):
LGTM

PR #988, opened 2025-05-23
refactor(33305): Redesign the openAPI specs for the datahub functions

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/33305/details/

The PR proposes a different implementation of the API endpoints, delivering the list of functions to the Datahub Designer. The API was designed to replace the hard-coded elements in the frontend, but was never utilised.

The existing APi, `GET   /api/v1/data-hub/functions` returns a complete JSON-Schema of every function in the system but suffers from a few flaws:
- It includes the user-defined transformation functions
- As a JSON-Schema block, it is expected to be used as such (e.g. in a form with RJSF) but cannot be reused in other elements (e.g. in a selector)
- It contains the metadata as an undocumented property of JSON-Schema
- It cannot be filtered out easily 

The PR deprecates the old endpoint to add a new dedicated API, `GET   /api/v1/data-hub/function-specs`
- The endpoint returns a list of  `FunctionSpecs`, with the usual items-based pattern
- The `FunctionSpecs` contains the full JSON-Schema, an optional UI Schema and the metadata as a type element
- The `FunctionMetadata` summarises the possible metadata of a function
- The `BehaviorPolicyTransitionEvent` enums the list of supported transition events

### Out-of-scope
- The old endpoint should not be used by end-user, as its relevance is questionable. It has been marked as `deprecated` and must be removed from the API at the earliest opportunity
- frontend and backend implementation have their tickets:
  - https://hivemq.kanbanize.com/ctrl_board/57/cards/29945/details/
  - https://hivemq.kanbanize.com/ctrl_board/57/cards/33331/details/

### DIFF
--- a/ext/hivemq-edge-openapi-2025.9-SNAPSHOT.yaml
+++ b/ext/hivemq-edge-openapi-2025.9-SNAPSHOT.yaml
@@ -14,7 +14,7 @@ info:
     ## OpenAPI
     HiveMQ's REST API provides an OpenAPI 3.0 schema definition that can imported into popular API tooling (e.g. Postman) or can be used to generate client-code for multiple programming languages.
   title: HiveMQ Edge REST API
-  version: 2025.4-SNAPSHOT
+  version: 2025.9-SNAPSHOT
   x-logo:
     url: https://www.hivemq.com/img/svg/hivemq-bee.svg
 tags:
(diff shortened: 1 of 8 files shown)


oli-hivemq (approved):
I like the approach of proposing OpenAPI changes - the contract between backend and frontend.
My approval shouldn't mean anything here, the backend team should agree with the changes. I'm approving anyway ;)

vanch3d:
> I like the approach of proposing OpenAPI changes - the contract between backend and frontend. My approval shouldn't mean anything here, the backend team should agree with the changes. I'm approving anyway ;)

I still need your review, mostly because I cannot review myself :-)

busybeetree (approved):

PR #998, opened 2025-05-29
fix(24800): Add the selector widget for topics in Delivery.redirectTo

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/24800/details/

The PR fixes the rendering of the `topic` property of the `Delivery.redirectTo` function by associating it with the custom format `mqqtt-topic` and select-based custom widget

### Out-of-scope
- The `interpolation` format associated with the topic filters is not supported yet. It will be dealt with in a subsequent ticket, see https://hivemq.kanbanize.com/ctrl_board/57/cards/33325/details/
- The fetching of existing topics for the selector options is not prtoperly implemented, see  https://hivemq.kanbanize.com/ctrl_board/57/cards/31942/details/


oli-hivemq (approved):
LGTM 🦣

PR #1004, opened 2025-06-10
fix(29953): Fix bugs when publishing behaviour policies

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/29953/details/

The PR fixes a series of problems with Behaviour Policies and the check/publishing/loading workflow
- The functions are now dynamically fetched from the backend and filtered from the policy and transition event that the Operation node is connected to
- The fetching process has been refactored to allow the filtering to happen outside of a React hook, ensuring the check/publish process is now fully using custom node validations
- The  `Datahub.Transform` function, a frontend-only placeholder for combining serialiser, deserialiser and scripts, was fixed to reflect its availability in Data policies only. This oversight was allowing end-users to create unsupported behaviour policies. 

The PR also fixes a few code duplication sniffed out by SonarQube

### Out-of-scope
- The loading of a policy and subsequent use as a draft is still buggy; it will be fixed in a subsequent ticket

### Before 
HiveMQ-Edge-06-11-2025_12_11

HiveMQ-Edge-06-11-2025_12_07

### After
HiveMQ-Edge-06-11-2025_12_06

HiveMQ-Edge-06-11-2025_12_06 (1)


oli-hivemq (approved):
LGTM 🦣

PR #1007, opened 2025-06-11
fix(33816): Fix the visualisation of the validation process`

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/33816/details/

The PR fixes a bug that prevented nodes to be visually updated when the validation is progressing through the graph.

The recent migration to `React Flow` changed the object mutation of an updated node. 

### After


oli-hivemq (approved):
LGTM 🦣

PR #1009, opened 2025-06-13
fix(29734): Fix a series of bugs when publishing policies with multiple resources 

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/29734/details/

The PR fixes a series of different issues that prevented policies, both Data and Behaviour, from being properly published. Most of the errors resolve around the use of multiple resources at a given point, e.g. multiple schemas with the same (or different) name (or version) in a validator. 

- The loaded elements of a policy have a proper UUID assigned to their node (with a type-based prefix)
- The determination of connected nodes now correctly handles version, source and target  
- Function name and version are now properly handled, from `1` to `latest`
- The detection of a script or schema version at load time is now properly identifying multiple versions of the same resource

### Design decision
The changes introduce a strong identification of the nodes in the `Designer`, mapping individual elements of the policy payload. In particular, different instances of the same resource used across the policy (e.g. in different `Datahub.Transform` operations) are materialised with different nodes, even if they were created from the same node.

### Out-of-scope
- Concerning the uniqueness of resource nodes, we cannot anticipate the end-users' preferred usage. A choice to recombine duplicated resources into a single node will have to be offered. It will be done in a subsequent ticket, see https://hivemq.kanbanize.com/ctrl_board/57/cards/33959/details/

### Aim
HiveMQ-Edge-06-13-2025_11_15

### Before
HiveMQ-Edge-06-13-2025_11_20

### After
HiveMQ-Edge-06-13-2025_11_16

### DIFF
--- a/hivemq-edge-frontend/src/__test-utils__/hooks/WrapperAccessibleDraggableProvider.tsx
+++ b/hivemq-edge-frontend/src/__test-utils__/hooks/WrapperAccessibleDraggableProvider.tsx
@@ -1,0 +1,10 @@
+import { AccessibleDraggableProvider } from '@/hooks/useAccessibleDraggable'
+import type { FC, PropsWithChildren } from 'react'
+
+export const getAccessibleDraggableProvider = () => {
+  const WrapperEdgeProvider: FC<PropsWithChildren> = ({ children }) => {
+    return <AccessibleDraggableProvider>{children}</AccessibleDraggableProvider>
+  }
+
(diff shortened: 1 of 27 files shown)


oli-hivemq:
I've tried to see the differences between aim, before and after images. Next time, for the benefit of the reviewer, could you highlight the areas that are different?

oli-hivemq (approved):
Left a few comments, but LGTM 🦣

PR #1016, opened 2025-06-16
feat(33975): Add the openAPI specs for the interpolation service

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/33975/details/

The PR adds the OpenAPI specs required to implemented the endpoint that returns the list of interpolation variables used in the Data Hub instance

### Out-of-scope
- The backend endpoints will be implemented in a separate ticket, see https://hivemq.kanbanize.com/ctrl_board/57/cards/33977/details/
- The variables are currently hardcoded (and wrongly defined) in the frontend; they will be properly implemented in a separated ticket, see https://hivemq.kanbanize.com/ctrl_board/57/cards/33325/details/

### DIFF
--- a/ext/openAPI/components/schemas/InterpolationVariable.yaml
+++ b/ext/openAPI/components/schemas/InterpolationVariable.yaml
@@ -1,0 +1,31 @@
+type: object
+required:
+  - variable
+  - type
+  - description
+  - isDataPolicy
+  - isBehaviorPolicy
+properties:
+  variable:
+    type: string
+    description: The unique variable name
+  type:
+    type: string
+    enum: ["string", "long"]
+  description:
+    type: string
+    description: The description of the variable name
+  isDataPolicy:
+    type: boolean
+    description: Indicate whether the variable can be used in a Data Policy
+    default: true
+  isBehaviorPolicy:
+    type: boolean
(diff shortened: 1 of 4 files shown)


oli-hivemq (approved):
LGTM (with copilot's typo fix) 🦣

ext/openAPI/components/schemas/InterpolationVariable.yaml
```diff
@@ -0,0 +1,31 @@
...
+  variable:
+    type: string
+    description: The unique variable name
+  type:
+    type: string
+    enum: ["string", "long"]
+  description:
+    type: string
+    description: The description of the variable name
+  isDataPolicy:
```
caoccao, line 18: Would `enum policyType` be a more appropriate?

caoccao (changes requested):
Just a small question.

ext/openAPI/components/schemas/InterpolationVariable.yaml
vanch3d: Sorry, strike the comment above, there is a not-so-bad solution: 
```
    policyType:
            type: array
            uniqueItems: true
            items:
              $ref: "#/components/schemas/PolicyType"
```

PR #1020, opened 2025-06-18
fix(33990): Confirm the replacement of a new draft

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/33990/details/

The PR adds a confirmation modal to the "Create a new draft", when an existing draft is already in editing. 

This is to prevent any important edit from being wiped out by an impromptu click on the (primary) button on the Data Hub landing page. 

The modal offers the usual "cancel" choice, as well as "replace" (default, will navigate to the `Designer` in an empty draft) and "show" (will navigate to the `Designer` without clearing the existing draft)

### Before 
HiveMQ-Edge-06-18-2025_13_47

### After 
HiveMQ-Edge-06-18-2025_13_45

### DIFF
--- a/hivemq-edge-frontend/src/extensions/datahub/components/helpers/DraftCTA.spec.cy.tsx
+++ b/hivemq-edge-frontend/src/extensions/datahub/components/helpers/DraftCTA.spec.cy.tsx
@@ -41,14 +41,14 @@ describe('DraftCTA', () => {
       cy.get('[role="alertdialog"]').should('not.exist')
       cy.get('button').click()
       cy.get('section[role="alertdialog"]').within(() => {
-        cy.get('header').should('contain.text', 'Do you want to replace your draft?')
+        cy.get('header').should('contain.text', 'You already have an active draft')
         cy.getByTestId('confirmation-message').should(
-          'contain.text',
-          'You already have an active draft in the Designer.'
+          'have.text',
+          'If you create a new one, the content of your current draft will be cleared from the canvas. This action cannot be reversed.'
         )
         cy.get('footer button').eq(0).should('have.text', 'Cancel')
-        cy.get('footer button').eq(1).should('have.text', 'Show the draft')
-        cy.get('footer button').eq(2).should('have.text', 'Replace the draft')
+        cy.get('footer button').eq(1).should('have.text', 'Open existing draft')
+        cy.get('footer button').eq(2).should('have.text', 'Create new empty draft')
 
         cy.get('footer button').eq(0).click()
       })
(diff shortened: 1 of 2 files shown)


oli-hivemq (approved):
Added a long comment, and pre-approving for when you're ready, not to be blocked

vanch3d:
I went for 1. "cancel" (rightly or wrongly) is used in every modal to close it without action.
Needs a holistic review of the translations

oli-hivemq:
additional info from Gemini:

Of the three options, I would recommend Option 2 (More Descriptive - focusing on the user's goal) for the following reasons:
1. **Focus on User Intent**:
    * **Headline: "Continue your draft or start fresh?"** This immediately frames the choice around the user's primary goals (continuing work or starting anew) rather than just the system's action ("replace draft"). This is good UX.
    * It anticipates the user's likely thought process when encountering this situation.
2. **Clearer Explanation of Consequences**:
    * **Body**: "It looks like you have an active draft in the Designer. You can continue working on it, or start a new, empty draft. Starting a new draft will clear the canvas, and your current draft cannot be recovered."
    * This copy explicitly states both options (continue or start new) and clearly explains the consequence of starting new, including the irreversibility, but in a more conversational and less alarming tone than just "The action cannot be reversed" on its own. "Cannot be recovered" is also very precise about data loss.
3. **Improved Button Clarity**:
    * **"Stay here" or "Don't create new"**: These options are far clearer than "Cancel." They explicitly state what the user is not doing (creating a new draft) and imply they will remain in their current context or go back to where they were. "Stay here" is perhaps the most user-friendly.
    * **"Continue working on existing draft"**: This is an excellent improvement over "Show the draft." It clearly communicates that clicking this button will load the existing work for continued editing, which is almost certainly the user's intent if they choose this path.
    * **"Start new, empty draft"**: This is more precise than "Replace the draft" because it emphasizes that the new draft will be blank. The suggested addition of "Perhaps with a secondary confirmation step if the risk of data loss is high" is also a great UX design consideration.

**Why not the others?**
* **Option 1 (Minor Tweaks)**: While an improvement, it doesn't go far enough in reframing the interaction from a user-goal perspective. "Create new empty draft" is better than "Replace," but "Open existing draft" is still a bit passive compared to "Continue working on existing draft."
* **Option 3 (Concise and Action-Oriented)**: While conciseness is good, the headline "Discard current draft and start new?" might be a bit too aggressive or leading for the initial prompt, especially if the user hasn't fully decided to discard. "Keep current draft" is good, but "Open current draft" still lacks the explicit "working on" intent of Option 2.

Therefore, **Option 2** balances clarity, user-centric language, and effective communication of consequences most effectively.

PR #1022, opened 2025-06-19
fix(30991): Fix the update of policy status

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/30991/details/

The PR fixes a long-standing bug which, in practice, resulted in data policy configuration (especially the `id`) not being persisted when changed while a validation check was still "active"

The PR also fixes a bug in the loading of the data policy panel, with an untimely "Cannot load the existing policies for checking ids" error message when checking the IDs

### Before 
HiveMQ-Edge-06-19-2025_16_50

### After
HiveMQ-Edge-06-19-2025_16_51


oli-hivemq (approved):
LGTM 🦣

PR #1032, opened 2025-06-24
fix(34191): Fix Topic Filter handles and deletion of edges

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/34214/details/
See https://hivemq.kanbanize.com/ctrl_board/57/cards/34192/details/ 

The PR fixes two annoyingly clever bugs
- The handles of a topic filter used to be identified with the name of the filter. When the name is changed by the user, the handle id changed but NOT the edges connected to it. The IDs are not solely based on their index
- The deletion of an edge in the `Designer` was prevented by guards, even in `DRAFT` mode. This is now properly asserted. 

### Out-of-scope
- Changing the identification of the handle means that the connections are purely based on the order of the topics, if there are several. This is not a supported feature but it will need to be fixed in the future, to prevent connections to be added to the wrong modified topic


oli-hivemq (approved):
LGTM 🦣

PR #1037, opened 2025-06-30
Seelct the correct endpoint instead of picking the 1st

codepitbull:
**Motivation**

Resolves #34450

**Changes**

Select an endpoint based on the configured security policy instead of just picking the first one.


marregui (approved):
LGTM

PR #1042, opened 2025-07-09
Feature/key value store

codepitbull:
**Motivation**

Resolves #34454

**Changes**
Cleaning up the messy state of asnyc code.
This is only step one and covers the ProtocolAdapterWrapper and some parts of the OpcUaAdapter.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
+++ b/hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
@@ -40,13 +40,13 @@ import com.hivemq.edge.modules.api.adapters.ProtocolAdapterPollingService;
 import com.hivemq.persistence.domain.DomainTag;
 import com.hivemq.persistence.domain.DomainTagAddResult;
 import com.hivemq.protocols.northbound.NorthboundConsumerFactory;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.VisibleForTesting;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import jakarta.inject.Inject;
-import jakarta.inject.Singleton;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -56,6 +56,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -78,11 +79,11 @@ public class ProtocolAdapterManager {
     private final @NotNull ProtocolAdapterPollingService protocolAdapterPollingService;
     private final @NotNull ProtocolAdapterMetrics protocolAdapterMetrics;
     private final @NotNull InternalProtocolAdapterWritingService protocolAdapterWritingService;
-    private final @NotNull ExecutorService executorService;
     private final @NotNull ProtocolAdapterFactoryManager protocolAdapterFactoryManager;
     private final @NotNull NorthboundConsumerFactory northboundConsumerFactory;
     private final @NotNull TagManager tagManager;
     private final @NotNull ProtocolAdapterExtractor protocolAdapterConfig;
+    private final @NotNull AtomicBoolean refreshing = new AtomicBoolean(false);
 
     @Inject
     public ProtocolAdapterManager(
@@ -96,7 +97,6 @@ public class ProtocolAdapterManager {
             final @NotNull ProtocolAdapterMetrics protocolAdapterMetrics,
             final @NotNull InternalProtocolAdapterWritingService protocolAdapterWritingService,
             final @NotNull ProtocolAdapterFactoryManager protocolAdapterFactoryManager,
-            final @NotNull ExecutorService executorService,
             final @NotNull NorthboundConsumerFactory northboundConsumerFactory,
             final @NotNull TagManager tagManager,
             final @NotNull ProtocolAdapterExtractor protocolAdapterConfig) {
@@ -109,7 +109,6 @@ public class ProtocolAdapterManager {
         this.protocolAdapterPollingService = protocolAdapterPollingService;
         this.protocolAdapterMetrics = protocolAdapterMetrics;
         this.protocolAdapterWritingService = protocolAdapterWritingService;
-        this.executorService = executorService;
         this.protocolAdapterFactoryManager = protocolAdapterFactoryManager;
         this.northboundConsumerFactory = northboundConsumerFactory;
         this.tagManager = tagManager;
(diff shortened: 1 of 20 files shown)


hivemq-edge/src/main/java/com/hivemq/bootstrap/ioc/Injector.java
```diff
@@ -84,7 +84,7 @@
         AdapterModule.class,
         CombiningModule.class})
 @Singleton
-public interface Injector {
+public  interface Injector {
```
caoccao, line 87: This change could be reverted.

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
```diff
@@ -118,15 +117,13 @@ public ProtocolAdapterManager(
                 protocolAdapterWritingService.writingEnabled()));
     }
 
-    private synchronized @NotNull ProtocolAdapterWrapper createAdapterInternal(
+    private @NotNull ProtocolAdapterWrapper createAdapterInternal(
```
caoccao, line 120: Assuming 2 threads are calling this method at the same time with the same adapter ID,

- T1 calls `protocolAdapters.get(config.getAdapterId())` and gets `null`.
- T2 calls `protocolAdapters.get(config.getAdapterId())` and gets `null`.
- T1 calls `protocolAdapterFactory.createAdapter()`.
- T2 calls `protocolAdapterFactory.createAdapter()`.
- T1 calls `protocolAdapters.put()`.
- T2 calls `protocolAdapters.put()`.

Will that cause any problems?

caoccao (changes requested):
I hold a few concerns. Overall it looks good to me.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/listeners/OpcUaServiceFaultListener.java
```diff
@@ -0,0 +1,56 @@
...
+            @NotNull final String adapterId) {
+        this.protocolAdapterMetricsService = protocolAdapterMetricsService;
+        this.eventService = eventService;
+        this.adapterId = adapterId;
+    }
+
+    @Override
+    public void onServiceFault(final ServiceFault serviceFault) {
+        protocolAdapterMetricsService.increment(Constants.METRIC_SUBSCRIPTION_SERVICE_FAULT_COUNT);
+        log.info("OPC UA client of protocol adapter '{}' detected a service fault: {}", adapterId, serviceFault);
```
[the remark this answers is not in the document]
codepitbull: We agreed to use error only for actual errors inside edge not when something external goes wrong.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaClientConnection.java
```diff
@@ -79,289 +64,306 @@ class OpcUaClientConnection {
...
-                client.removeSessionActivityListener(activityListener);
-            } catch (final Throwable e) {
-                log.warn("Failed to remove session activity listener {}: {}", activityListener, e.getMessage());
-            }
+
+        final var subscriptionOptional = subscribe(client, subscriptionIdOptional);
+
+        if(subscriptionOptional.isEmpty()) {
+            log.error("Failed to create or transfer OPC UA subscription. Closing client connection.");
+            quietlyCloseClient(client, false,null, null);
```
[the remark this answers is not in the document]
codepitbull: The listeners shouldn't be cleaned up.
For this instance we actually want them to report anything they see.

hivemq-edge/src/main/java/com/hivemq/bootstrap/ioc/Injector.java
codepitbull: Done

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterWrapper.java
```diff
@@ -88,40 +92,87 @@ public ProtocolAdapterWrapper(
...
+                                        return false;
+                                    });
+                        }
+                    })
+                    .thenCompose(Function.identity())
+                    .whenComplete((ignored, throwable) -> stateChangeOngoing.set(false));
+        } else {
+            log.warn("State change for adapter with id '{}' is already ongoing, ignoring start request", getId());
+        }
+        return CompletableFuture.completedFuture(true);
```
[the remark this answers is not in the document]
codepitbull: Moved to a different approach.

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterWrapper.java
```diff
@@ -88,40 +92,87 @@ public ProtocolAdapterWrapper(
...
+                            return true;
+                        } else {
+                            log.error("Error stopping adapter with id {}", adapter.getId(), throwable);
+                            return false;
+                        }
+                    });
+        } else {
+            log.warn("State change for adapter with id '{}' is already ongoing, ignoring start request", getId());
+        }
+        return CompletableFuture.completedFuture(true);
```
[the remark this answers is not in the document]
codepitbull: Moved to a different approach.

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
codepitbull: You are right. Wanted to fix this in a separate ticket but in the end it bothered me too much :)
Solved.

hivemq-edge/src/test/java/com/hivemq/protocols/ProtocolAdapterManagerTest.java
```diff
@@ -187,7 +176,7 @@ void test_startWriting_adapterFailedStart_resourcesCleanedUp() {
                 northboundConsumerFactory,
                 tagManager);
 
-        assertThrows(ExecutionException.class, () -> protocolAdapterManager.start(adapterWrapper).get());
+        protocolAdapterManager.start(adapterWrapper).get();
```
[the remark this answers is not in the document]
codepitbull: Not true. The failure is indicated by STOPPED value checked further down.

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterWrapper.java
```diff
@@ -88,40 +94,110 @@ public ProtocolAdapterWrapper(
...
+                        .thenCompose(Function.identity())
+                        .handle((ignored, t) -> {
+                            if(t == null) {
+                                createAndSubscribeTagConsumer();
+                                startPolling(protocolAdapterPollingService, input.moduleServices().eventService());
+                                return startWriting(writingEnabled, protocolAdapterWritingService)
+                                        .thenApply(v -> {
+                                            log.info("Successfully started adapter with id {}", adapter.getId());
+                                            setRuntimeStatus(ProtocolAdapterState.RuntimeStatus.STARTED);
+                                            startFutureRef.get().complete(true);
```
caoccao, line 128: I wonder this is really safe because it is out of `synchronized (startFutureRef)`. If `startAsync()` was called twice at the same time, what would happen?

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterWrapper.java
```diff
@@ -88,40 +94,110 @@ public ProtocolAdapterWrapper(
...
+                                            return true;
+                                        });
+                            } else {
+                                log.error("Error starting protocol adapter", t);
+                                stopPolling(protocolAdapterPollingService);
+                                return stopWriting(protocolAdapterWritingService)
+                                        .thenApply(v -> {
+                                            log.error("Error starting adapter with id {}", adapter.getId(), t);
+                                            setRuntimeStatus(ProtocolAdapterState.RuntimeStatus.STOPPED);
+                                            startFutureRef.get().complete(false);
```
caoccao, line 139: I wonder this is really safe because it is out of `synchronized (startFutureRef)`. If `startAsync()` was called twice at the same time, what would happen?

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterWrapper.java
```diff
@@ -88,40 +94,110 @@ public ProtocolAdapterWrapper(
...
+                            return output.getOutputFuture();
+                        }).thenCompose(Function.identity()).handle((ignored, throwable) -> {
+                            if (destroy) {
+                                log.info("Destroying adapter with id '{}'", getId());
+                                adapter.destroy();
+                            }
+                            if (throwable == null) {
+                                setRuntimeStatus(ProtocolAdapterState.RuntimeStatus.STOPPED);
+                                log.info("Stopped adapter with id {}", adapter.getId());
+                                stopFutureRef.get().complete(true);
```
caoccao, line 182: Same here. I wonder this is really safe.

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterWrapper.java
```diff
@@ -88,40 +94,110 @@ public ProtocolAdapterWrapper(
...
+                                stopFutureRef.set(null);
+                                return true;
+                            } else {
+                                log.error("Error stopping adapter with id {}", adapter.getId(), throwable);
+                                stopFutureRef.get().complete(false);
```
caoccao, line 187: Same here. I wonder this is really safe.

caoccao (changes requested):
I have some new concerns.

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
```diff
@@ -118,213 +117,195 @@ public ProtocolAdapterManager(
...
-                        .fire();
+    }
 
-            }
-        } else {
-            log.error("Tried removing non existing adapter '{}'", id);
-        }
-        return false;
+    private void deleteAdapterInternal(final @NotNull String adapterId) {
+        final ProtocolAdapterWrapper adapterWrapper = protocolAdapters.remove(adapterId);
```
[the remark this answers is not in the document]
codepitbull: That's exactly what I have verified over the past few days.
Just to be save I added a nullcheck and a warn log in case we ever should get here.

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterWrapper.java
codepitbull: That's why I use doublecheck logging.
The double check (check atomic -> sync on atomic -> check atomic again) prevents a second future from being created for the stop and the start case.
I first check the content of startFutureRef, then do the sync, then check again.
Only then the running future is created.

Since the assign of the AtomicReference happens in the sync block it is ensured no other Completablefuture will ever be created for starting or stopping.

Do see any way this assertion could be broken?

PR #1045, opened 2025-07-14
Rework bridge handling to fix a bug

codepitbull:
**Motivation**

Resolves #34787

**Changes**
Editing a bridge via the UI would lead to a broken bridge.
Breakage happened because the client wasn't correctly stopped.

This PR reworks the bridge restart code and fixes the issue by correctly stopping the existing bridge and then restarzing it.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/bridge/BridgeService.java
+++ b/hivemq-edge/src/main/java/com/hivemq/bridge/BridgeService.java
@@ -35,6 +35,7 @@ import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -55,8 +56,8 @@ public class BridgeService {
 
     private final @NotNull Map<String, Throwable> bridgeNameToLastError = new ConcurrentHashMap<>(0);
 
-    private final @NotNull Map<String, MqttBridgeAndClient> activeBridgeNamesToClient = new ConcurrentHashMap<>();
-    private final @NotNull Map<String, MqttBridge> allKnownBridgeConfigs = new ConcurrentHashMap<>();
+    private final @NotNull Map<String, MqttBridgeAndClient> activeBridgeNamesToClient = new HashMap<>();
+    private final @NotNull Map<String, MqttBridge> allKnownBridgeConfigs = new HashMap<>();
 
     @Inject
     public BridgeService(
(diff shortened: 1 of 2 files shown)


hivemq-edge/src/main/java/com/hivemq/bridge/BridgeService.java
```diff
@@ -108,36 +118,36 @@ public synchronized void updateBridges(final @NotNull List<MqttBridge> bridges)
...
+            final var newBridge = bridgeIdToConfig.get(bridgeId);
             if(active != null) {
-                if(active.bridge().equals(bridge)) {
-                    log.debug("Not restarting bridge {} because config is unchanged", bridgeId);
+                if(active.bridge().equals(newBridge)) {
+                    log.debug("Not restarting bridge {} because config  is unchanged", bridgeId);
                 } else {
                     log.info("Restarting bridge {} because config has changed", bridgeId);
+                    allKnownBridgeConfigs.remove(bridgeId);
+                    allKnownBridgeConfigs.put(bridgeId, newBridge);
```
caoccao, line 130: These 2 lines are not atomic. There is a possibility that the bridge is not found by another thread. Is this expected? Why not call put directly without remove?

hivemq-edge/src/main/java/com/hivemq/bridge/BridgeService.java
```diff
@@ -108,36 +118,36 @@ public synchronized void updateBridges(final @NotNull List<MqttBridge> bridges)
...
 
-        newMapOfKnownBridges.forEach((bridgeId, bridge) -> {
+        toUpdate.forEach(bridgeId -> {
             final var active = activeBridgeNamesToClient.get(bridgeId);
+            final var newBridge = bridgeIdToConfig.get(bridgeId);
             if(active != null) {
-                if(active.bridge().equals(bridge)) {
-                    log.debug("Not restarting bridge {} because config is unchanged", bridgeId);
+                if(active.bridge().equals(newBridge)) {
+                    log.debug("Not restarting bridge {} because config  is unchanged", bridgeId);
```
caoccao, line 126: Extra space.

hivemq-edge/src/main/java/com/hivemq/bridge/BridgeService.java
```diff
@@ -108,36 +118,36 @@ public synchronized void updateBridges(final @NotNull List<MqttBridge> bridges)
...
-                log.info("Adding bridge {}", bridgeId);
-                newMapOfKActiveBridges.put(
-                        bridge.getId(),
-                        new MqttBridgeAndClient(bridge, internalStartBridge(bridge)));
-            }
+        toAdd.forEach(bridgeId -> {
+            final var newBridge = bridgeIdToConfig.get(bridgeId);
+            log.info("Adding bridge {}", bridgeId);
+            allKnownBridgeConfigs.put(bridgeId, newBridge);
+            activeBridgeNamesToClient.put(
```
caoccao, line 143: Same question. These 2 put calls are not atomic. Is this expected?

caoccao (changes requested):
I hold some concerns on the thread-safety.

hivemq-edge/src/main/java/com/hivemq/bridge/BridgeService.java
codepitbull: Fixed

caoccao (approved):
LGTM

PR #1056, opened 2025-07-22
fix(35003): Update bridge status in the landing page

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/35003/details/

The PR fixes a bug, whereas the list of bridges displayed on the landing page does not update its status, unlike the live adapter page. 

### Out-of-scope
- The status of an adapter is still incorrect in many cases. It will be fixed in further tickets

### After
HiveMQ-Edge-07-22-2025_14_19


oli-hivemq (approved):
agree with Copilot's suggestions.
Looks good otherwise

PR #1062, opened 2025-07-30
feat(34981): Create the OpenAPI specs for the Pulse integration

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/34981/details/ 

The PR creates the basic OpenAPI specs for the integration of Pulse into Edge:
- the activation of the Pulse Client
- the CRUD operations on the managed assets

### DIFF
--- a/ext/openAPI/components/schemas/Asset.yaml
+++ b/ext/openAPI/components/schemas/Asset.yaml
@@ -1,0 +1,26 @@
+type: object
+description: A blah blah bleh managed by the Pulse Broker
+properties:
+  id:
+    type: string
+    format: uuid
+    description: The unique id of the asset
+    readOnly: true
+  name:
+    type: string
+    description: The user-facing name of the asset
+    readOnly: true
+  topic:
+    type: string
+    format: mqtt-topic
+    readOnly: true
+  schema:
+    type: string
+    format: data-url
+    description: The json schema for this topic filter in the data uri format.
+    readOnly: true
+required:
+  - id
+  - name
+  - schema
+  - topic
--- a/ext/openAPI/components/schemas/Asset.yaml
+++ b/ext/openAPI/components/schemas/Asset.yaml
@@ -1,24 +1,28 @@
 type: object
-description: A blah blah bleh managed by the Pulse Broker
+description: The definition of an asset as sourced from the Pulse Broker
 properties:
   id:
     type: string
     format: uuid
     description: The unique id of the asset
+    example: "123e4567-e89b-12d3-a456-426614174000"
     readOnly: true
   name:
     type: string
     description: The user-facing name of the asset
     readOnly: true
+    example: "The fancy asset"
   topic:
     type: string
     format: mqtt-topic
+    description: The topic associated with the asset
+    example: "a/valid/mqtt/topic"
     readOnly: true
   schema:
     type: string
     format: data-url
-    description: The json schema for this topic filter in the data uri format.
-    readOnly: true
+    description: The schema associated with the asset, in a JSON Schema and data uri format.
+    example: "data:application/schema+json;base64,eyJ0eXBlIjoiT2JqZWN0IiwiaWQiOiJhc3NldC1zY2hlbWEifQ=="
 required:
   - id
   - name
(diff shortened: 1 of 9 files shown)


ext/openAPI/components/schemas/Asset.yaml
```diff
@@ -0,0 +1,30 @@
...
+  name:
+    type: string
+    description: The user-facing name of the asset
+    readOnly: true
+    example: "The fancy asset"
+  topic:
+    type: string
+    format: mqtt-topic
+    description: The topic associated with the asset
+    example: "a/valid/mqtt/topic"
```
sfrehse, line 19: Probably hard to model in the OpenAPI schema, but there can only be one Asset per topic. 

Pulse must not repeat a topic to different Assets.

ext/openAPI/components/schemas/Asset.yaml
```diff
@@ -0,0 +1,30 @@
+type: object
+description: The definition of an asset as sourced from the Pulse Broker
+properties:
```
sfrehse, line 3: An asset also contains a description. It might be helpful for the user.

ext/openAPI/components/schemas/Asset.yaml
```diff
@@ -0,0 +1,30 @@
...
+    example: "The fancy asset"
+  topic:
+    type: string
+    format: mqtt-topic
+    description: The topic associated with the asset
+    example: "a/valid/mqtt/topic"
+    readOnly: true
+  schema:
+    type: string
+    format: data-url
```
sfrehse, line 23: We also strongly consider providing a description to each property in the schema, so you may want to show them.

ext/openAPI/paths/pulse/activation_token.yaml
```diff
@@ -0,0 +1,37 @@
...
+  description: Send an activation token to the Pulse Client
+  summary: Update an activation token
+  operationId: update-pulse-activation-token
+  tags:
+    - Pulse
+  requestBody:
+    content:
+      application/json:
+        schema:
+          $ref: ../../components/schemas/PulseValidationToken.yaml
```
sfrehse, line 26: I couldn't find the `PulseValidationToken.yaml`. It is basically is a JWT encoded as base64 as normal. 

Do you consider showing a status of the connection of the agent to the Pulse server?

vanch3d: Yes, there will be a status polling

ext/openAPI/components/schemas/Asset.yaml
vanch3d: Yes, both `title` and `description` should be used in a JSON-Schema and, when existing, have been systematically used in Edge. They cannot be made required, however

ext/openAPI/components/schemas/Asset.yaml
```diff
@@ -0,0 +1,34 @@
...
+  id:
+    type: string
+    format: uuid
+    description: The unique id of the asset
+    example: "123e4567-e89b-12d3-a456-426614174000"
+    readOnly: true
+  name:
+    type: string
+    description: The user-facing name of the asset
+    readOnly: true
```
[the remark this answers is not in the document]
codepitbull: Waiting for reply form the pulse team.

ext/openAPI/components/schemas/PulseStatus.yaml
```diff
@@ -0,0 +1,17 @@
+type: object
+description: Information on the activation status of the pulse agent and its connection to the platform.
+properties:
```
vanch3d, line 3: Should the status come with an error message as well, to document the `ERROR` enum on both activation and runtime? A single message at root level (similar to the `Status` for both bridge and adapter: 
```
  message:
    type: string
    description: A message associated with the state of a connection
```
But I would recommend to be a ProblemDetail

```
  message:
    $ref: "./ProblemDetails.yaml"
```

vanch3d (commented):
Status looks good, just a question about an error message in the payload

ext/openAPI/components/schemas/Asset.yaml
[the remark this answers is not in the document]
codepitbull: No length limit from their side.

PR #1064, opened 2025-07-31
Feature/35256 streamline datahub

DC2-DanielKrueger:
https://hivemq.kanbanize.com/ctrl_board/57/cards/35256/details/


hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/impl/ProtocolAdapterInterceptorHandler.java
```diff
@@ -273,7 +248,7 @@ public void run() {
             final ProtocolAdapterPublishInboundOutputImpl output = outputHolder.get();
             if (output.isPreventDelivery()) {
                 dropMessage(output);
-                resultFuture.set(PublishReturnCode.FAILED);
+                resultFuture.set(PublishingResult.failed(null, null));
```
codepitbull, line 251: Why is there no reason code here?

hivemq-edge/src/main/java/com/hivemq/mqtt/handler/publish/PublishingResult.java
```diff
@@ -0,0 +1,72 @@
...
+    private PublishingResult(
+            final @NotNull PublishReturnCode publishReturnCode,
+            final @Nullable String reasonString,
+            final @Nullable AckReasonCode ackReasonCode) {
+        this.publishReturnCode = publishReturnCode;
+        this.reasonString = reasonString;
+        this.ackReasonCode = ackReasonCode;
+    }
+
+    public static @NotNull PublishingResult failed(final @Nullable String reasonString, final @Nullable AckReasonCode ackReasonCode) {
```
codepitbull, line 46: Shpuldn't reasonString and ackReasonCode be @NotNull in this case?
Internally it's ok but this method should always enforce both to be provided.

DC2-DanielKrueger: reason code should be NotNull. Reason String might make sense to be optional as we sometimes have no good msg for the client.

PR #1070, opened 2025-08-05
Add DelegatingCompletableFuture

caoccao:
Add `DelegatingCompletableFuture` to simplify the protocol adapter start and stop process.

One assumption of the current design is: calling `startAsync()` or `stopAsync()` N times will fire N times events.

However, that's not necessary. This change assumes: calling `startAsync()` or `stopAsync()` N times will fire 1 event.

It introduces `DelegatingCompletableFuture` that is a subclass of `CompletableFuture`.

- It has a property `future` that is a `CompletableFuture`.
- If `future` is null, calls to all methods will get `null`, `false`, or empty `CompletableFuture`.
- If `future` is not null, calls to all methods will be redirected to that `future`.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/protocols/DelegatingCompletableFuture.java
+++ b/hivemq-edge/src/main/java/com/hivemq/protocols/DelegatingCompletableFuture.java
@@ -1,0 +1,636 @@
+/*
+ *  Copyright 2019-present HiveMQ GmbH
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.hivemq.protocols;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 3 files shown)


hivemq-edge/src/main/java/com/hivemq/protocols/DelegatingCompletableFuture.java
```diff
@@ -0,0 +1,699 @@
...
+    public void setFuture(@Nullable final CompletableFuture<T> future) {
+        this.future = future;
+    }
+
+    @Override
+    public T get(final long timeout, final @NotNull TimeUnit unit)
+            throws InterruptedException, ExecutionException, TimeoutException {
+        final CompletableFuture<T> future = this.future;
+        if (future == null) {
+            return null;
```
codepitbull, line 60: I am worried about all the return null within the future. All of these methods are actually treated as not ever returning a null in the original implementation.

codepitbull (commented):
The `return null` statements in the new Future feel quite dangerous.

hivemq-edge/src/main/java/com/hivemq/protocols/DelegatingCompletableFuture.java
caoccao: You are right. I think there are 2 options:

1. Throw an `ExecutionException`.
2. Sleep a while and wait for `future` to be not null periodically.

(1) doesn't seem to be friendly to the consumers. (2) seems to be a good solution. What do you think? Or you have any other solutions?

caoccao: I suppose the current solution and my proposed solution are eventually the same, as there is a spin lock waiting for the future to be set. It's more on where the complexity lies. I'm fine with keeping the current solution.

codepitbull: Ok, then I will go back to fix the remaining issue.

PR #1071, opened 2025-08-05
fix(35265): Fix bug with interpolation in the Bridge's subscription destinations

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/35265/details/

The PR fixes a bug in the Bridge subscriptions, preventing end-users from utilising interpolation strings (such as `{#}`). Note that the "fix" consists of removing the custom validation of MQTT topics, therefore allowing users to type incorrect topics. 

### Out-of-scope
- The fix removes the `mqtt-topic` format for the string, disabling the validation of an MQTT topic (hence the problem with the wildcard `'#`). This weakens the validation of a bridge and will need to be reintroduced in a subsequent ticket.
- The interpolation patterns are not fully documented. The re-implementation of a properly validated topic + interpolation (e.g. `mqtt-topic-interpolation`) will need to be designed and implemented properly for both backend and frontend

### Before 
HiveMQ-Edge-08-05-2025_10_43

### After
HiveMQ-Edge-08-05-2025_10_42

### DIFF
--- a/hivemq-edge-frontend/src/api/schemas/bridge.json-schema.ts
+++ b/hivemq-edge-frontend/src/api/schemas/bridge.json-schema.ts
@@ -36,7 +36,8 @@ export const bridgeSchema: JSONSchema7 = {
         destination: {
           type: 'string',
           description: `The destination topic for this filter set.`,
-          format: 'mqtt-topic',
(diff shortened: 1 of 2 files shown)


caoccao (approved):
LGTM

PR #1081, opened 2025-08-13
refactor: Update openapi spec for delete/update a token

caoccao:
https://hivemq.kanbanize.com/ctrl_board/57/cards/35468/details/


vanch3d (approved):
one question left unanswered
LGTM

PR #1088, opened 2025-08-19
Missed an error state in modbus

codepitbull:
**Changes**

Modbus missed setting the connection state to error in one place.


caoccao (approved):
LGTM

PR #1095, opened 2025-08-21
refactor(35534): Refactor the Managed Assets

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/35534/details/

The PR changes the way Managed Assets are handled in Edge. 
- `Combiner` are used for the primary definition of asset mappings, as it currently does for more general user-centred solutions
- `PULSE_AGENT` has been added to the list of valid entities. When added to the sources of a combiner, it becomes an Asset Mapper
- `destination` has an extra parameter, assetId, used when mapping with an asset is done
- `ManagedAsset` is refactored to only have a reference to the mapping (`mappingId`) and the `status`


caoccao (approved):
LGTM

PR #1096, opened 2025-08-21
chore(35266): Add a CI check workflow for the openAPI specs

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/35266/details/

The PR adds a simple job in the `check` pipeline, triggering the `Redocly lint` operation on the OpenAPI specs. 

The job will fail if errors are found in the specs, not with warnings. It will then fail the whole check pipeline.


caoccao (approved):
LGTM

PR #1120, opened 2025-09-03
feat(35995): Fix bug with the fetching of combined entities 

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/35995/details/

The PR fixes a bug with the aggregation of requests required to extract tags and topic filters from their sources.

It also fixes two small accessibility issues and a few tests


oli-hivemq (approved):
LGTM 🦣

PR #1136, opened 2025-09-08
refactor(36081): Add filters to the search params of asset-based routes

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/36081/details/

The PR improves the configuration of the `Asset List` table by linking the initial state of column filters to option search params in the router. 

This is used in particular in the `Onboarding` page, where individual "to-do" items can now link to the `Asset list` with a specific status prefilled in the filter. 

### Out-of-scope
- The binding is one-directional, the route defines the initial state of the filters and the table. Changing the filters manually doesn't change the current route. This behaviour might be addressed in a future ticket 

### Before
HiveMQ-Edge-09-08-2025_12_12_PM

HiveMQ-Edge-09-08-2025_12_11_PM (1)

### After
HiveMQ-Edge-09-08-2025_12_11_PM

### DIFF
--- a/hivemq-edge-frontend/src/modules/Pulse/components/assets/AssetsTable.spec.cy.tsx
+++ b/hivemq-edge-frontend/src/modules/Pulse/components/assets/AssetsTable.spec.cy.tsx
@@ -189,7 +189,7 @@ describe('AssetsTable', () => {
       })
   })
 
-  it.only('should handle navbigation', () => {
+  it.only('should handle navigation', () => {
     cy.intercept('/api/v1/management/pulse/managed-assets', MOCK_PULSE_ASSET_LIST).as('getStatus')
 
     cy.mountWithProviders(<AssetsTable />, {


hivemq-edge-frontend/src/components/PaginatedTable/PaginatedTable.spec.cy.tsx
```diff
@@ -114,45 +114,83 @@ describe('PaginatedTable', () => {
...
+          columns={MOCK_COLUMN}
+          pageSizes={[5, 10, 20]}
+          enableColumnFilters
+          aria-label="table"
+        />
+      )
+
+      cy.getByAriaLabel('Clear selected options').should('not.exist')
+
+      cy.get('th').eq(0).find('div#react-select-column1-placeholder').should('have.text', 'Search... (4)')
```
oli-hivemq, line 131: Not related to this PR but I'm realising it now: I'm not sure to understand the meaning of the number between parentheses. Is it the number of items currently shown?
Is it needed? 🤔

PR #1179, opened 2025-09-23
Port PR 583 from community edition (max packet size evaluation)

marregui:
**card**: https://hivemq.kanbanize.com/ctrl_board/57/cards/34090/details/

This PR moves the packet size check earlier in the decode pipeline — before the decoder waits for the full message body to arrive in the buffer. Previously, an oversized packet would block the decoder waiting for bytes that would never be processed anyway.                                                   
                                                                                                                                                                                                                                                                                                                    
  This means:                                                                                                                                                                                                                                                                                                       
  - Oversized packets are rejected immediately based on the fixed header + remaining length, without buffering the entire payload first
  - CONNECT packets too large get a proper CONNACK PACKET_TOO_LARGE (after reading just enough bytes to determine the MQTT protocol version)                                                                                                                                                                        
  - Non-CONNECT packets too large get a DISCONNECT PACKET_TOO_LARGE (with force-close for MQTT 3.x)

  In short: faster rejection of oversized messages, less memory wasted buffering data that will be thrown away.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/codec/decoder/MQTTMessageDecoder.java
+++ b/hivemq-edge/src/main/java/com/hivemq/codec/decoder/MQTTMessageDecoder.java
@@ -205,12 +205,10 @@ public class MQTTMessageDecoder extends ByteToMessageDecoder {
         // connack with PACKET_TOO_LARGE for Mqtt5
         if (messageType == MessageType.CONNECT) {
             // Theoretically, remaining length could be too short to read the protocol version.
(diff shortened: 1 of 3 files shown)


marregui:
> It seems the majority of the changes are reformatting. What are the actual changes?

the changes come from https://github.com/hivemq/hivemq-community-edition/pull/583/files mainly MQTTMessageDecoder calculates max packet size prior to reading the message

codepitbull (approved):

codepitbull (approved):

PR #1189, opened 2025-09-26
fix: At most one destination per asset

caoccao:
https://hivemq.kanbanize.com/ctrl_board/57/cards/36687/details/


codepitbull (approved):
One question, nothing critical.

PR #1190, opened 2025-09-26
Fix/36265 pa fsm

codepitbull:
**Motivation**

Resolves [#36265](https://hivemq.kanbanize.com/ctrl_board/57/cards/36265/details/)

**Changes**

Fix flakyness by introducing a proper FSM for protocoladapters.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/fsm/ProtocolAdapterFSM.java
+++ b/hivemq-edge/src/main/java/com/hivemq/fsm/ProtocolAdapterFSM.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2019-present HiveMQ GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.hivemq.fsm;
 
 import com.hivemq.adapter.sdk.api.state.ProtocolAdapterState;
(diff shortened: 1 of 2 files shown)


hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
```diff
@@ -68,7 +69,7 @@
 public class ProtocolAdapterManager {
     private static final Logger log = LoggerFactory.getLogger(ProtocolAdapterManager.class);
 
-    private final @NotNull Map<String, ProtocolAdapterWrapper> protocolAdapters = new ConcurrentHashMap<>();
+    private final @NotNull Map<String, ProtocolAdapterWrapper> protocolAdapters;
```
caoccao: I suggest we rename it to `protocolAdapterMap`.

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
```diff
@@ -150,21 +176,32 @@ public void refresh(final @NotNull List<ProtocolAdapterEntity> configs) {
...
             adaptersToBeCreated.forEach(name -> {
                 try {
-                    log.debug("Creating adapter '{}'", name);
-                    startAsync(createAdapterInternal(protocolAdapterConfigs.get(name), versionProvider.getVersion())).get();
-                } catch (final InterruptedException | ExecutionException e) {
+                    if (log.isDebugEnabled()) {
+                        log.debug("Creating adapter '{}'", name);
+                    }
+                    startAsync(createAdapterInternal(protocolAdapterConfigs.get(name),
+                            versionProvider.getVersion())).get();
```
caoccao: This makes the adapter creations in sequence. Do we want to run this in parallel? If YES, we could call `map()` to get a list of `Future`s, then call `get()` per future. It'll be much faster.

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
```diff
@@ -120,17 +122,41 @@ public ProtocolAdapterManager(
 
     // API, must be threadsafe
 
+    private static void syncFuture(final @NotNull Future<?> future) {
```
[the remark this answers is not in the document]
marregui: will consider it when performing the integration. I will favour readability indeed.

PR #1198, opened 2025-09-30
fix: Add notification when assets are changed

caoccao:
https://hivemq.kanbanize.com/ctrl_board/57/cards/36520/details/

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/pulse/asset/PulseAgentAssetChangedListener.java
+++ b/hivemq-edge/src/main/java/com/hivemq/pulse/asset/PulseAgentAssetChangedListener.java
@@ -16,6 +16,7 @@
 
 package com.hivemq.pulse.asset;
 
+import com.hivemq.configuration.reader.AssetMappingExtractor;
 import com.hivemq.configuration.reader.PulseExtractor;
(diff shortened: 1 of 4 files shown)


hivemq-edge/src/main/java/com/hivemq/pulse/asset/PulseAgentAssetChangedListener.java
```diff
@@ -16,26 +16,34 @@
...
+    public @NotNull AssetMappingExtractor getAssetMappingExtractor() {
+        return assetMappingExtractor;
+    }
+
     public @NotNull PulseExtractor getPulseExtractor() {
         return pulseExtractor;
     }
 
     @Override
     public void onAssetsChanged(final @NotNull List<Asset> remoteAssets) {
```
codepitbull, line 46: I went down into the code invoked by this and there are blocking calls (replaceConfigsAndTriggerWrite) there.
Is it fine for the invoker to be blocked?

caoccao: Yes. I believe these changes have to be atomic, and there is a global lock ensuring this is atomic.

PR #1200, opened 2025-09-30
fix(36931): Fix missing mappingId

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/36931/details/

The PR fixes a bug when creating a new asset mapper (or updating an existing one), where the `id` of the new mapping associated with the asset is not properly sent to the `updateManagedAsset` API


caoccao (approved):
LGTM

PR #1202, opened 2025-10-01
fix: Allow mapping ID to be set to null in updateManagedAsset

caoccao:
https://hivemq.kanbanize.com/ctrl_board/57/cards/36520/details/


marregui (approved):
LGTM

PR #1224, opened 2025-10-16
feat(37430) Add support for quick filters

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/37430/details/

The PR adds full support for quick filters, i.e. when manually defined criteria are saved in a configuration document and then directly reused 

- Quick filters can be activated or deactivated
- When activated, their filter criteria are applied to the canvas, along any other criteria defined in the manual section. The same `OR` operator is applied 
- Activated filters are NEVER saved in any new quick filter. Only the manual criteria are.

The PR also improves the search toolbar by making it collapsible, helping to keep the workspace as open as possible

### Before 
HiveMQ-Edge-10-14-2025_10_27_AM

### After 
HiveMQ-Edge-10-16-2025_01_43_PM
HiveMQ-Edge-10-16-2025_01_43_PM (1)
HiveMQ-Edge-10-16-2025_01_44_PM

### DIFF
--- a/hivemq-edge-frontend/src/modules/Workspace/components/filters/filters.utils.ts
+++ b/hivemq-edge-frontend/src/modules/Workspace/components/filters/filters.utils.ts
@@ -67,7 +67,6 @@ export const applyStatusFilter = (node: Node, criteria?: ActiveFilter<MultiValue
 export const applyQuickFilters = (node: Node, quickFilters: FilterConfigurationOption[]) => {
   const coverage = quickFilters.reduce<boolean[]>((acc, quickFilter) => {
     if (!quickFilter.isActive) {
-      // acc.push(false)
       return acc
     }
     const isHidden = hideNodeWithFilters(node, quickFilter.filter)


hivemq-edge-frontend/src/modules/Workspace/components/controls/CanvasToolbar.tsx
```diff
@@ -1,24 +1,88 @@
...
-      <HStack>
-        <SearchEntities />
-        <DrawerFilterToolbox />
-      </HStack>
+      <Box
+        m="1.2px"
+        display="flex"
+        alignItems="center"
+        transition="max-width 0.4s cubic-bezier(0.4,0,0.2,1)"
+        maxWidth={expanded ? '1280px' : '80px'}
```
oli-hivemq: Might be a good idea to have these sizes configurable. I would expect these variables to be part of the design system at some point, but for now I'd have them defined somewhere (in CSS or as const somewhere) so you can at least reuse them

PR #1236, opened 2025-10-22
test(37074) Optimisation of Visual Regressions tests in Percy

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/37074/details/

The PR optimises the usage of Percy visual regressions tests in the Edge frontend, as we have a limited number of runs available avery months (and used to significantly exceed that limit)

The PR removes EVERY test that were originial created on the Cypress Component test suites. Instead, it organised the Percy snpashots on selected (possibly critical) paths in the Cypress E2E test suites. 

The outcomes of this first refactoring tasks is a focus on 20 snapshots of whole "pages" at different points of interaction, compared to 90 snapshots of components in isolation


oli-hivemq (approved):
LGTM 🦣

PR #1247, opened 2025-10-28
Support by filtering endpoints against both, MessageSecurityMode and SecurityPolicyUri

codepitbull:
**Motivation**

Resolves #36664

**Changes**
Updated the EndpointFilter to optionally check for combinations if MessageSecurityMode and SecurityPolicyUri.


caoccao (approved):
LGTM

PR #1318, opened 2025-12-01
Chore/generate xsd

codepitbull:
**Motivation**

Replace the outdated XSD with a generated one.

The added code shows just how problematic the original XSD is.
All methods in GenSchemaMain deal with specialities that were introduced by handcrafting the file.
The handcrafting lead to the use of constructs which aren't supported by JAXB.

So this ensures the file is regenerated on each build and therefore tested together with every test loading an XML config.


caoccao (approved):
I suggest we keep the generated config.xsd.

1. It's easy to review for external users.
2. It's easy to review the change log, diff, blame using git commands.

Also, could we do the same to the protocol adapters to generate tag definition xsd and config  xsd?

PR #1325, opened 2025-12-08
Fix/38363 bridge stall

codepitbull:
**Motivation**

Resolves #38363

**Changes**

Under certai nconditions a bridge would not resume after losing its connection.


hivemq-edge/src/main/java/com/hivemq/bridge/mqtt/BridgeMqttClient.java
```diff
@@ -400,13 +399,31 @@ public boolean isConnected() {
             }
             log.info("Bridge '{}' connected to {}:{}", bridge.getId(), bridge.getHost(), bridge.getPort());
             connected.set(true);
+
+            // Check if this is a reconnection (not the initial connection)
+            // On initial connection, we only flush buffered messages without resetting persistence state.
+            // On reconnection, we need to drain all stale state and re-poll from persistence.
+            final boolean isReconnection = !everConnected.compareAndSet(false, true);
```
caoccao: - When is `everConnected` reset to `false`?
- What about the 2nd, 3rd, ... reconnect? `isReconnection` is `false`.

hivemq-edge/src/main/java/com/hivemq/bridge/mqtt/RemoteMqttForwarder.java
```diff
@@ -253,11 +258,27 @@ private void finishProcessing(
...
+    /**
+     * Called when a publish fails - resets the inflight marker so the message can be retried
+     * instead of being removed from persistence. This prevents message loss on transient failures.
+     */
+    private void finishProcessingWithRetry(
+            final @NotNull QoS originalQoS,
+            final @NotNull String uniqueId,
+            final @NotNull String queueId) {
+        inflightCounter.decrementAndGet();
+        // Reset inflight marker instead of removing the message, allowing retry
```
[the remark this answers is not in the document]
caoccao: Agree with Copilot.

caoccao (commented):
I hold one concern.

hivemq-edge/src/main/java/com/hivemq/bridge/mqtt/BridgeMqttClient.java
marregui: - when we start, everConnected is false and becomes true, so isReconnection is false.
- on reconnection, everConnected is already true, stays true, and isReconnection is true.

on reconnection (isReconnection==true) we drain the queue (resets inflight markers in persistence), then we call onReconnect which does the checkBuffers (re-polls messages).

on initial connection (isReconnection==false) we flush in-memory buffered messages, BUT we do not reset inflight markers (to avoid duplicate delivery on restart).

scenario:
  1. Edge service is running, bridge is forwarding messages
  2. Connection drops mid-delivery - some messages have inflight markers set in persistence
  3. Edge service crashes or is killed before it can reconnect and reset the markers
  4. Edge service restarts
  5. everConnected is now false (fresh JVM, fresh object)
  6. Bridge connects → this is treated as "initial connection", not "reconnection"
  7. flushBufferedMessages() is called, but not drainQueue() / removeAllInFlightMarkers()
  8. Messages with stale inflight markers from step 2 remain stuck forever

should inflight markers be reset always on initialization, or should we always view connections as reconnections?, even for the very first connection?

hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
```diff
@@ -171,6 +171,54 @@ public void addForwarder(final @NotNull MqttForwarder mqttForwarder) {
...
+            // and don't hold any locks that could cause deadlock.
+            final Set<String> forwarderQueueIds = queueIdsForForwarder.get(fwdId);
+            if (forwarderQueueIds != null && !forwarderQueueIds.isEmpty()) {
+                if (log.isDebugEnabled()) {
+                    log.debug("Resetting inflight markers for forwarder '{}', {} queue(s)",
+                            fwdId, forwarderQueueIds.size());
+                }
+                final ImmutableList.Builder<ListenableFuture<Void>> futuresBuilder = ImmutableList.builder();
+                for (final String queueIdToReset : forwarderQueueIds) {
+                    futuresBuilder.add(queuePersistence.get().removeAllInFlightMarkers(queueIdToReset));
```
caoccao: What if `removeAllInFlightMarkers()` is interrupted by `Futures.allAsList(futuresBuilder.build()).get(30, TimeUnit.SECONDS)`? Is there a rollback logic?

PR #1353, opened 2026-01-18
EDG-70 Various cleanups of the auth test code

mschoenert-hivemq:
**Motivation**

All I wanted to do was to cleanup the double curly braces initializations in the test files.
They are very strange, because the register one object in the inner class init,
but then more objects are registered by explicit register statements externally.

But then I got carried away and made a ton more changes.

**Resolves**

Only my OCD - but no known issues with the code or the tests.

**Changes**

The test files with some refactoring.

**Note**

I think the cleanup is easier to observe when viewing the new files instead of the diffs.
Great care was taken to not change what is tested and how.


codepitbull (approved):
LGTM

PR #1354, opened 2026-01-19
PROTOBUF > streamline and update version

DC2-DanielKrueger:


hivemq-edge/build.gradle.kts
```diff
@@ -216,7 +216,7 @@ dependencies {
 
     //FIXME: should be in module instead
     // we need better module isolation for that as the modules pick up Netty from the app class loader
-    implementation("com.google.protobuf:protobuf-java:4.33.2")
+    implementation("com.google.protobuf:protobuf-java:4.33.4")
```
codepitbull: Can you use the versions from lib.versions.toml here please?

PR #1385, opened 2026-02-12
fix: Subscribe primary if it is not in instructions

caoccao:
https://hivemq.kanbanize.com/ctrl_board/57/cards/38959/details/

# Behavior Changes

## Eager consumer start

Each consumer starts immediately in `subscribe()` rather than batch-starting at the end, but this is benign since non-primary consumers only store data without triggering publish.

## Primary may not be stored

|                Scenario      | primary   | storeDataPoints |
| --- | --- | --- |
| Primary in instructions      | true      | true            |
| Non-primary in instructions  | false     | true            |
| Primary not in instructions  | true      | false           |

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/combining/runtime/DataCombiningRuntime.java
+++ b/hivemq-edge/src/main/java/com/hivemq/combining/runtime/DataCombiningRuntime.java
@@ -127,8 +127,8 @@ public class DataCombiningRuntime {
 
     public void stop() {
         consumers.forEach(InternalConsumer::close);
-        dataCombiningTransformationService.removeScriptForDataCombining(combining);
         consumers.clear();
+        dataCombiningTransformationService.removeScriptForDataCombining(combining);
     }
 
     public void triggerPublish(final @NotNull DataCombining dataCombining) {


hivemq-edge/src/main/java/com/hivemq/combining/runtime/DataCombiningRuntime.java
```diff
@@ -88,58 +88,36 @@ public void start() {
...
-                                    ref.scope(),
-                                    combining.sources().primaryReference().scope());
-                    return new InternalTagConsumer(ref.id(), ref.scope(), combining, isPrimary);
-                })
-                .forEach(consumer -> {
-                    tagManager.addConsumer(consumer);
-                    consumers.add(consumer);
-                });
+                .forEach(this::subscribeNonPrimary);
+        subscribe(primaryRef, true);
```
mschoenert-hivemq: I like the idea of `subscribeNonPrimary`.  But then you should also have `subscribePrimary`.

Additionally note that for `subscribePrimary` you also want to pass a flag that says whether the subscriber must store values for this tag/topicFilter in the hash map. Because - if the primary is not referenced in any instruction, then it would be wasteful to store its value in the hashmap and serialize it into the JSON object.

hivemq-edge/src/test/java/com/hivemq/combining/runtime/DataCombiningRuntimeTest.java
```diff
@@ -0,0 +1,788 @@
...
+        verify(tagManager, times(2)).addConsumer(captor.capture());
+        final List<TagConsumer> consumers = captor.getAllValues();
+        // First is the instruction tag (non-primary), second (last) is the primary
+        assertThat(consumers.get(0).getTagName()).isEqualTo("tag2");
+        assertThat(consumers.get(0).getScope()).isEqualTo("adapter2");
+        assertThat(consumers.get(1).getTagName()).isEqualTo("tag1");
+        assertThat(consumers.get(1).getScope()).isEqualTo("adapter1");
+
+        // Non-primary consumer should NOT trigger a publish
+        consumers.get(0).accept(List.of(new DataPointImpl("tag2", "{\"v\":1}")));
```
mschoenert-hivemq: I feel like this can't work. The instruction above is referencing $.value - and here you are constructing a datapoint that has no "value" property - but only a "v" property.

mschoenert-hivemq: Maybe this is the cause for the problems that you see in the transformation service?

hivemq-edge/src/main/java/com/hivemq/combining/runtime/DataCombiningRuntime.java
```diff
@@ -88,58 +88,36 @@ public void start() {
...
+        internalSubscriptions.stream().map(InternalSubscription::queueConsumer).forEach(QueueConsumer::start);
+    }
```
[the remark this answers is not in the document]
caoccao: I have one reason not to do that early: if there are any exceptions in that forEach call, no subscription is started.

The current design which is the original design implied all subscriptions are set up successfully, then they will be started. I'd like to keep this.

hivemq-edge/src/main/java/com/hivemq/combining/vanilla/VanillaDataCombiningTransformationService.java
```diff
@@ -73,11 +74,23 @@ public VanillaDataCombiningTransformationService(
                         parsingException = e;
                     }
                     if (parsingException != null) {
-                        LOGGER.warn("Source json path {} does not exist", sourceJsonPath, parsingException);
-                    } else if (value == null) {
-                        LOGGER.warn("No data found for source json path {}", sourceJsonPath);
+                        if (instruction.dataIdentifierReference().type() == DataIdentifierReference.Type.TAG
```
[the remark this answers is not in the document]
caoccao: Yes, that's very bad.

hivemq-edge/src/main/java/com/hivemq/combining/runtime/DataCombiningRuntime.java
```diff
@@ -88,58 +88,36 @@ public void start() {
...
 
-        internalSubscriptions.forEach(
-                internalSubscription -> internalSubscription.queueConsumer().start());
+    private void subscribe(final @NotNull DataIdentifierReference ref, final boolean isPrimary) {
+        switch (ref.type()) {
+            case TAG -> {
+                log.debug("Starting tag consumer for tag {} with scope {}", ref.id(), ref.scope());
+                final InternalTagConsumer consumer =
+                        new InternalTagConsumer(ref.id(), ref.scope(), combining, isPrimary);
```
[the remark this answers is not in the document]
caoccao: I refactored the code similar to what you described.

hivemq-edge/src/main/java/com/hivemq/combining/runtime/DataCombiningRuntime.java
```diff
@@ -88,97 +86,49 @@ public void start() {
...
-                })
-                .forEach(consumer -> {
-                    tagManager.addConsumer(consumer);
-                    consumers.add(consumer);
-                });
-
-        // Topic filter subscriptions - derive from instructions()
-        // instead of from sources().topicFilters()
+        final DataIdentifierReference primaryRef = combining.sources().primaryReference();
+        final AtomicBoolean primaryProcessed = new AtomicBoolean();
```
mschoenert-hivemq: OK. This is a stylistic comment - you may disregard it. But here goes:
* I always try to avoid side effects (of method calls or - in this case - a fluid chain)
* and Atomics are tricky beasts - I always try to avoid them unless there is no way around it
 
Here you have both together (the side effect of setting the atomic primaryProcess).
I would really rewrite that:
* the first pass that simply skips over the primary and subscribes all other sources
* a second pass that checks whether primary is referenced in the instructions (and that can stop early, by using 'anyMatch')
* and finally the subscription of the primary

Performance doesn't matter here at all: we are just setting up a combiner mapping. It happens much more when we assemble and publish. And those lists of instructions will very seldom be very long. That said, I would even assume that your one-pass solution, where the final for-each have to contend for the atomic, is actually slower. ;-)

But in the end - your choice.

hivemq-edge/src/main/java/com/hivemq/combining/runtime/DataCombiningRuntime.java
```diff
@@ -88,97 +86,49 @@ public void start() {
         log.debug("Starting data combining {}", combining.id());
         // prepare the script for the data combining
         dataCombiningTransformationService.addScriptForDataCombining(combining);
```
mschoenert-hivemq: Do you actually know what this does? 
I thought that the call from `triggerPublish` to `dataCombiningPublishService.publish` does all the magic. 
Why must we add this script here? What would happen if we didn't?

hivemq-edge/src/main/java/com/hivemq/combining/runtime/DataCombiningRuntime.java
```diff
@@ -88,97 +86,49 @@ public void start() {
...
+            }
+            case TOPIC_FILTER -> {
+                log.debug("Starting mqtt consumer for filter {}", ref.id());
+                consumer = new InternalTopicFilterConsumer(ref, primary, storeDataPoints);
+            }
+            default -> log.warn("Unsupported data identifier reference type: {}", ref.type());
+        }
+        if (consumer != null) {
+            consumers.add(consumer);
+            consumer.start();
```
mschoenert-hivemq: I would move the start() even further in (in the creation of the consumers). 
Then each branch of the switch would ONLY do
````
consumers.add(new InternalXXXConsumer(YYY)
````
But again - I'm not super passionate about that.

mschoenert-hivemq (commented):
That looks pretty good to me. 
Some stylistic changes.
Can we get to an agreement about the handling of PULSE_ASSETS?

mschoenert-hivemq (approved):
LGTM

PR #1387, opened 2026-02-12
Reenable telemetry, which was disabled due to a whitespace in the destination url

marregui:
card: https://hivemq.kanbanize.com/ctrl_board/57/cards/38974/details/


caoccao (approved):
LGTM

PR #1404, opened 2026-02-24
refactor(EDG-34) Source Ownership Visibility in Mapping Instructions 

vanch3d:
**Kanban Ticket:** https://linear.app/hivemq/issue/EDG-34/display-of-sources-in-the-destination

## Description

This PR solves a silent ambiguity in the mapping editor: when two different adapters
expose a property with the same name (e.g. `value`), the drop zone showed identical
cards with no way to tell which data source was actually mapped.

The enhancement introduces:

- **Source identity row**: An ownership chip now appears above each mapped drop zone,
  showing exactly which adapter or topic filter the value comes from
- **Scoped TAG display**: Adapter tags render as `scope :: id/path` using the existing
  PLCTag component, with full path formatting
- **Topic filter display**: Topic filter sources render using the existing TopicFilter
  component
- **Backward compatibility**: Instructions without ownership information render exactly
  as before — no ownership row appears

### User Experience Improvements

**What users gain:**

- **Instant source identification**: No more guessing which adapter a `value` field
  comes from — the source is visible at a glance
- **Consistent visual language**: The ownership chip reuses the same PLCTag and
  TopicFilter components already present in the source panel — no new visual vocabulary
  to learn
- **Clutter-free when not needed**: Instructions without a `sourceRef` (legacy or
  unmapped) show no extra row — the layout is only enriched when the information exists

### Technical Summary

**Implementation highlights:**

- `formatOwnershipString(ref)` added to `topic-utils.ts` — returns `scope :: id` when
  scoped, `id` only otherwise
- `resolveSourceOwner(ref)` helper extracted outside the component — clean
  `if / if / return null` pattern, no nested JSX conditionals
- `scope` now correctly propagated in both mouse-drop and keyboard-drop paths (was
  previously omitted)

---

### Previous Behavior — Source ownership invisible

When multiple adapters contributed properties with the same name, the mapping cards were
visually identical. Users had no way to verify which source was mapped to which
destination.

**Limitations:**

- Source identity completely absent from the mapping card
- Identical-looking cards for different adapters — no way to distinguish at a glance
- Silent misconfiguration risk: a mapping could point to the wrong adapter's data

##### Before — three identical instructions, ownership invisible
image
---

### New Behavior — Source identity visible on every mapped instruction

An ownership row now appears directly above the drop zone, using the existing entity tag
components so the visual language is immediately familiar.

##### Ambiguity resolved — three sources, three clearly distinct owners

image

**Key Visual Elements:**

- **PLCTag chip** (TAG source): displays `adapter-name :: path / to / tag` with path
  segments separated by thin-space slashes — identical to the source panel
- **TopicFilter chip** (TOPIC_FILTER source): displays the full filter string

**User Benefits:**

At a glance, users can now verify that each destination field is wired to the intended
source — even when all source properties share the same name.

---

### Comprehensive Testing

- **29 tests total, all passing ✅**
- **Unit tests (13)**: `formatTopicString` pass-through with ownership prefix;
  `formatOwnershipString` across all source types and scope variants
- **Component tests (16)**: ownership row presence/absence for all source types; drop
  zone structurally unchanged; responsive at 504 px; accessibility with axe-core

### Visual Regression

- Cypress component screenshots at 800×900, 504×900, and 1400×900 viewports
- Investigation test suite provides before/after visual baseline for all scenarios

---

## Breaking Changes

None. Legacy instructions without `sourceRef` render identically to before.

- ✅ Existing mappings load and display correctly
- ✅ No API contract changes

---

## Performance Impact

No impact. The ownership row is computed synchronously from an already-available prop.

---

## Accessibility

- ✅ Validated with axe-core across TAG, TOPIC_FILTER, readonly, and no-instruction
  variants
- ✅ PLCTag and TopicFilter carry their own ARIA attributes
- ✅ Ownership row is purely informational — no interactive elements introduced

---

## Reviewer Notes

**Focus areas for review:**

1. **Ownership chip layout**: Does the chip integrate visually at both full-screen and
   narrow-panel widths?
2. **Legacy backward compatibility**: Instructions without `sourceRef` must not show an
   ownership row
3. **PULSE_ASSET suppression**: Intentionally produces no ownership row — confirm this
   is the desired behaviour before the type is supported
4. **`formatOwnershipString` contract**: Returns a raw string; `EntityTag` applies
   `formatTopicString` — verify no double-formatting occurs

**Manual testing suggestions:**

1. Open the combiner mapping editor with two adapters sharing a tag name
2. Map both to different destination fields
3. Observe: each card shows a distinct ownership chip above the drop zone ✅
4. Open a legacy combiner (no `sourceRef` on instructions)
5. Observe: no ownership row — cards render as before ✅

**Quick test commands:**

```bash
# Unit tests
pnpm vitest run src/components/MQTT/topic-utils.spec.ts

# Component tests
pnpm cypress:run:component --spec "src/components/rjsf/MqttTransformation/components/mapping/MappingInstruction.spec.cy.tsx"
```

### DIFF
--- a/hivemq-edge-frontend/src/components/rjsf/MqttTransformation/components/mapping/MappingInstruction.tsx
+++ b/hivemq-edge-frontend/src/components/rjsf/MqttTransformation/components/mapping/MappingInstruction.tsx
@@ -94,11 +94,7 @@ const MappingInstruction: FC<MappingInstructionProps> = ({
         const sourceRef: DataIdentifierReference | undefined = target.dataReference
           ? { id: target.dataReference.id, type: target.dataReference.type, scope: target.dataReference.scope ?? null }
           : undefined
-        onChange?.(
-          [...target.path, target.key].join('.') as string,
-          [...property.path, property.key].join('.') as string,
-          sourceRef
-        )
+        onChange?.([...target.path, target.key].join('.'), [...property.path, property.key].join('.'), sourceRef)
       },
     })
   }, [onChange, property])
@@ -185,8 +181,8 @@ const MappingInstruction: FC<MappingInstructionProps> = ({
                       }
                     : undefined
                   onChange?.(
-                    [...source.property.path, source.property.key].join('.') as string,
-                    [...property.path, property.key].join('.') as string,
+                    [...source.property.path, source.property.key].join('.'),
+                    [...property.path, property.key].join('.'),
                     sourceRef
                   )
 
@@ -213,7 +209,7 @@ const MappingInstruction: FC<MappingInstructionProps> = ({
                 aria-label={t('rjsf.MqttTransformationField.instructions.actions.clear.aria-label')}
                 icon={<RiDeleteBin2Fill />}
                 onClick={onHandleClear}
-                isDisabled={Boolean(!instruction?.source)}
+                isDisabled={!instruction?.source}
               />
             </ButtonGroup>
             {instruction?.source ? (


antpaw (approved):
LGTM :shark:

PR #1405, opened 2026-02-24
Chore/edg 93/errorprone fixes

codepitbull:
**Motivation**

Resolves #EDG-93

**Changes**
Addresses all errorprone findings of level ERROR and additional findings which would be considered BUGS.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/configuration/entity/pulse/PulseAssetEntity.java
+++ b/hivemq-edge/src/main/java/com/hivemq/configuration/entity/pulse/PulseAssetEntity.java
@@ -90,7 +90,7 @@ public class PulseAssetEntity implements EntityValidatable {
         return new Builder();
     }
 
-    public boolean matchesAsset(final @NotNull Asset remoteAsset) {
+    public boolean equals(final @NotNull Asset remoteAsset) {
         if (!Objects.equals(id.toString(), remoteAsset.id())) {
             return false;
         }
--- a/hivemq-edge/src/main/java/com/hivemq/pulse/utils/PulseAgentAssetUtils.java
+++ b/hivemq-edge/src/main/java/com/hivemq/pulse/utils/PulseAgentAssetUtils.java
@@ -83,7 +83,7 @@ public class PulseAgentAssetUtils {
                     }
                 } else {
                     // Asset is found remotely. We check if it has changed.
-                    if (localAsset.matchesAsset(remoteAsset)) {
+                    if (localAsset.equals(remoteAsset)) {
                         newLocalAssets.add(localAsset.withMapping(localAsset
                                 .getMapping()
                                 .withStatus(


hivemq-edge/src/main/java/com/hivemq/api/auth/jwt/JwtAuthenticationProvider.java
```diff
@@ -202,16 +202,12 @@ public Optional<Long> getExpiryTimeMillis(final @NotNull String tokenValue) {
...
+        } catch (final MalformedClaimException e) {
             log.warn("jwt parse failed, reason {}", e.getMessage());
-        } catch (InvalidJwtException e) {
-            if (!e.hasExpired()) {
-                log.trace("jwt validation failed, reason {}", e.getMessage());
-            } else {
-                log.trace("jwt validation failed, reason {}", e.getMessage());
-            }
+        } catch (final InvalidJwtException e) {
+            log.trace("jwt validation failed, reason {}", e.getMessage());
```
mschoenert-hivemq: feels like we are throwing away valuable information?

```suggestion
        } catch (final InvalidJwtException e) {
            if (e.hasExpired()) {
                log.trace("jwt validation failed, token expired, reason {}", e.getMessage());
            } else {
                log.trace("jwt validation failed, reason {}", e.getMessage());
            }
```

hivemq-edge/src/main/java/com/hivemq/api/json/CustomConfigSchemaGenerator.java
```diff
@@ -254,11 +254,11 @@ protected String stringPattern(final @NotNull MemberScope<?, ?> memberScope) {
         protected @Nullable ModuleConfigField getModuleFieldInfo(MemberScope<?, ?> member) {
             ModuleConfigField annotation = member.getAnnotation(ModuleConfigField.class);
             if (annotation == null) {
-                MemberScope<?, ?> source;
-                if (member instanceof FieldScope) {
-                    source = ((FieldScope) member).findGetter();
-                } else if (member instanceof MethodScope) {
-                    source = ((MethodScope) member).findGetterField();
+                final MemberScope<?, ?> source;
+                if (member instanceof FieldScope fieldScope) {
```
mschoenert-hivemq: May I use this opportunity to say that I find `X instanceof Y Z` one of the ugliest constructs in Java.

hivemq-edge/src/main/java/com/hivemq/persistence/InMemoryProducerQueues.java
```diff
@@ -67,18 +66,12 @@ public InMemoryProducerQueues(final int persistenceBucketCount, final int amount
         bucketsPerQueue = persistenceBucketCount / amountOfQueues;
         shutdownGracePeriod = InternalConfigurations.PERSISTENCE_SHUTDOWN_GRACE_PERIOD_MSEC.get();
 
-        final ImmutableList.Builder<AtomicLong> counterBuilder = ImmutableList.builder();
```
mschoenert-hivemq: What was this ever supposed to do?

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/data/TagManager.java
```diff
@@ -56,8 +56,8 @@ public TagManager(final @NotNull MetricsHolder metricsHolder) {
     @Override
     public void feed(final @NotNull String tagName, final @NotNull List<DataPoint> dataPoints) {
         lastValueForTag.put(tagName, dataPoints);
+        readWriteLock.readLock().lock();
```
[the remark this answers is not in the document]
mschoenert-hivemq: OK. Either I didn't get across. Or you.

My point was that in some files we first assign the lock to a local variable and then lock it in a second statement - and then in the final we just call `lock.unlock();`.

Here we directly lock, without assigning to a local variable - and then in the final repeat `readWriteLock.readLock().unlock()`.

I wonder whether we should always to it consistently in all files. And if so, then I would vote for the idiom with the extra local variable.

hivemq-edge/src/main/java/com/hivemq/persistence/InMemoryProducerQueues.java
codepitbull: I have no idea. Now it's dead code.

mschoenert-hivemq:
> Things are separated exactly that by commits. You could have simply clicked throug the commits :)

Wow. That is a lot of eggs in my face this morning.

PR #1407, opened 2026-02-25
refactor(EDG-35): Tag Ownership Rendering Across the UI

vanch3d:
**Linear Ticket:** [EDG-35 — Visual Rendering of Tags in Data Combinings (Mappings)](https://linear.app/hivemq/issue/EDG-35/visual-rendering-of-tags-in-data-combinings-mappings)

## Description

This PR extends ownership-aware rendering across every location in the combiner UI where a
PLC tag is displayed. Previously, tag names were shown in isolation — when two adapters both
exported a tag with the same name, the UI offered no way to tell them apart. Now, any tag
that belongs to an adapter is shown as `{adapter-id} :: {tag-name}` wherever the context is
ambiguous.

The enhancement introduces:

- **Scoped tag chips in the source selector**: Selected tags from different adapters are now
  visually distinct even when they share the same tag name.
- **Adapter attribution in the dropdown option list**: Each option row shows its owning adapter
  as secondary muted text, and options are sorted so same-named tags from different adapters
  appear adjacent.
- **Scoped headings in the schema browser**: The panel header identifying a source schema now
  shows the full ownership string when scope is set.
- **Scoped tag names in the combiner summary table**: Tag chips in the Sources column now
  recover scope from the mapping instructions so the table is readable without cross-referencing.
- **Consistent chip widgets in the workspace filter**: The topic/tag/filter filter now renders
  the same EntityTag components used everywhere else (visual consistency fix).

### User Experience Improvements

**What users gain:**

- **No more identical chips**: Selecting `temperature` from two different adapters now produces
  two clearly labelled chips instead of two indistinguishable `[🏷 temperature]` tags.
- **Confident selection from the dropdown**: Same-named tags from different adapters now appear
  side-by-side with an adapter label, so the user always picks the right source.
- **Readable combiner summaries**: The mapping table shows which adapter each tag came from
  without requiring the user to open each mapping to investigate.
- **Graceful fallback preserved**: Single-adapter contexts continue to show plain tag names —
  the scope prefix is added only where it resolves real ambiguity.

### Technical Summary

- The `formatOwnershipString(ref: DataIdentifierReference)` utility (introduced in EDG-34)
  is now the single source of truth for all ownership string formatting across the codebase.
- Ambiguity was assessed per location: scope data availability and multi-adapter co-occurrence
  risk were evaluated before changing any rendering path.
- The combiner summary table uses a `getScopeForTag()` helper to recover scope from
  `sources.primary` and `instructions[].sourceRef` as the deprecated `sources.tags` string
  array does not carry scope directly.
- The dropdown option list keeps `label = tag.name` for natural alphabetical ordering
  (same-named tags group together), adds adapter attribution as a separate column, and
  extends `filterOption` so typing an adapter name also narrows the list.

### Ambiguity Analysis

The key design decision in this PR was **where ownership context is truly needed** versus
where the plain tag name is unambiguous and adding a prefix would add noise.

Every `PLCTag` / `TopicFilter` / `AssetTag` rendering location was audited against two questions:

1. **Is scope data available at the call site?**
2. **Can multiple adapters produce same-named tags in this view simultaneously?**

| Location | Ambiguity risk | Scope available | Action |
| --- | --- | --- | --- |
| Source selector chips (`CombinedEntitySelect`) | High | Yes — `ref.scope` | Show ownership string |
| Source selector option list (`CombinedEntitySelect`) | High | Yes — `entity.id` | Show adapter as secondary column; sort by tag name |
| Schema browser heading (`JsonSchemaBrowser`) | High | Yes — `dataReference.scope` | Show ownership string |
| Schema browser error fallback (`CombinedSchemaLoader`) | High | Yes — `dataReference.scope` | Show ownership string |
| Combiner summary table (`DataCombiningTableField`) | Medium | Reconstructed from `instructions` | Show ownership string with fallback |
| Workspace filter chips (`FilterTopics`) | Low | Not available (data layer gap) | Widget consistency only; ownership deferred |
| Southbound instruction drop zone (`MappingInstruction`) | Unambiguous | — | Done in EDG-34; no change |
| Per-adapter tag table (`TagTableField`) | Unambiguous | — | Single-adapter scope; no change |

The workspace filter (`FilterTopics`) is the only location where scope data is not yet
available. Ownership display there requires `useGetDomainOntology` to expose adapter IDs per
tag — this is out of EDG-35 scope and tracked as a follow-up.

### Previous Behavior — Ambiguous Tag Display

The UI showed plain tag names everywhere, with no ownership context. This was harmless in
single-adapter workflows, but as soon as multiple adapters exported the same tag name the
display became ambiguous:

**Limitations:**

- Selecting the same-named tag from two different adapters produced two identical
  `[🏷 temperature ×]` chips in the source selector — indistinguishable at a glance.
- The dropdown option list showed duplicate entries for the same tag name, relying on the
  description sub-text (which can be absent or also identical) to differentiate them.
- The schema panel heading showed only the tag ID with no adapter context.
- The combiner summary table listed raw tag IDs, making it impossible to tell which adapter
  was the source without opening each mapping.

### 1. Selected chips — ownership string resolves the ambiguity

When the same tag name appears in more than one selected adapter, each chip now shows its
owning adapter as a prefix.

**Before:**

image

**After:**

image

**Key Visual Elements:**

- **Ownership prefix**: `{adapter-id} :: ` is prepended to the tag name inside the chip.
- **Fallback preserved**: A tag from a single-adapter context still shows the plain tag name
  without any prefix — the scope is only injected when set on the reference.

**User Benefits:**

At a glance, users can confirm they have selected the correct temperature sensor from the
correct adapter, without needing to hover, expand, or cross-reference another panel.

### 2. Schema browser headings — ambiguity across two source panels

When two adapters are combined and both export a tag with the same name, the source schema
browser renders one panel per source. Previously both panels carried an identical tag heading,
making it impossible to know which schema belonged to which adapter.

**Before:**

image

**After:**

image

**Key Visual Elements:**

- **Unique headings**: Each schema panel is now headed by `{adapter} :: {tag-name}`, making it
  immediately clear which schema tree belongs to which adapter.
- **Same fix in error state**: If a schema fails to load, the error fallback heading
  (`CombinedSchemaLoader`) uses the same ownership string — so even a failed source is
  unambiguously identified.

**User Benefits:**

When drag-mapping fields from multiple source schemas onto a target, users can see at a glance
which panel is which — no need to collapse one tree and re-read a sidebar to orient
themselves.

## Test Coverage

- **20+ component tests, all passing** across five changed components
- **Cypress component tests** for each location:
    - Scope present → ownership string visible
    - Scope absent → plain tag name, no `::` separator
    - Accessibility (`cy.checkAccessibility()`) passes for every changed component
- **Existing tests unmodified and passing**: All pre-existing scenarios continue to work
  unchanged; the fallback to plain tag name means no regression in single-adapter setups.
- **Investigation files** created alongside each component to capture before/after scenarios
  for visual documentation.

## Breaking Changes

None. All changes are backward compatible:

- ✅ When `scope` is null, the display falls back to the plain tag name — existing single-adapter
  behaviour is preserved exactly.
- ✅ No API or data-model changes — scope is read from fields already present on
  `DataIdentifierReference`.
- ✅ `formatOwnershipString` is a pure function with a null-safe fallback; callers with no
  scope pass through unchanged.

## Performance Impact

No impact. All changes are in render functions; no new queries or heavy computations are added.
The `getScopeForTag` helper in `DataCombiningTableField` is an inline array scan over the
already-loaded `instructions` array.

## Accessibility

- ✅ `cy.checkAccessibility()` passes for all five modified components.
- ✅ The ownership prefix is part of the visible label text, so screen readers announce the full
  `{adapter} :: {tag}` string — no additional ARIA markup required.
- ✅ The muted adapter column in the dropdown option list uses sufficient colour contrast
  (Chakra `gray.500` on white).

## Documentation

- `.tasks/EDG-35-tag-ownership-rendering/TASK_BRIEF.md` — Objective, requirements, success criteria
- `.tasks/EDG-35-tag-ownership-rendering/DESIGN_ANALYSIS.md` — Per-location analysis, approach
  trade-offs, implementation sequence, and test strategy

## Future Enhancements

- [ ] Extend `useGetDomainOntology` to expose adapter IDs per tag, enabling ownership display
  in the workspace `FilterTopics` component (currently deferred — data layer gap).
- [ ] Migrate `DataCombiningTableField` sources column away from the deprecated `sources.tags`
  string array to the instructions-based model, eliminating the need for scope reconstruction.

## Reviewer Notes

**Focus areas for review:**

1. **Ambiguity threshold**: The `FilterTopics` chip intentionally shows plain tag names (scope
   not available). Confirm this is acceptable as a deferred item.
2. **`getScopeForTag` fallback**: The helper returns `null` when scope cannot be recovered from
   either `sources.primary` or `instructions`. Confirm the plain-name fallback is correct for
   combiner rows that pre-date the scope introduction.
3. **Sort order in `CombinedEntitySelect`**: Options are now sorted `[tag.name, adapterId]`.
   Verify this is the expected UX for the dropdown (same-named tags from different adapters
   appear adjacent).
4. **`filterOption` override**: Typing an adapter name now filters the option list. Verify
   this does not interfere with existing tag-name search behaviour.

**Manual testing suggestions:**

1. Open the data combiner editor with two adapters that both export a tag named `temperature`.
2. Open the source dropdown and observe two `temperature` rows, each with a muted adapter label,
   sorted adjacent.
3. Select both — verify the two chips read `{adapter-1} :: temperature` and
   `{adapter-2} :: temperature`.
4. With a single-adapter setup, verify the chip still shows `temperature` with no prefix.
5. Open a mapping with a primary tag that has scope set — verify the schema browser heading
   reads `{adapter} :: {tag}`.

### DIFF
--- a/hivemq-edge-frontend/src/modules/Mappings/CombinerMappingManager.spec.cy.tsx
+++ b/hivemq-edge-frontend/src/modules/Mappings/CombinerMappingManager.spec.cy.tsx
@@ -10,7 +10,8 @@ import { MOCK_DEVICE_TAGS, mockAdapter, mockProtocolAdapter_OPCUA } from '@/api/
 import { MOCK_TOPIC_FILTER } from '@/api/hooks/useTopicFilters/__handlers__'
 import { mockEmptyCombiner } from '@/api/hooks/useCombiners/__handlers__'
 
-import { NodeTypes } from '@/modules/Workspace/types'
+import { EntityType } from '@/api/__generated__'
+import { IdStubs, NodeTypes } from '@/modules/Workspace/types'
 import CombinerMappingManager from './CombinerMappingManager'
 
 const INITIAL_ENTRY = '/workspace'
@@ -173,6 +174,7 @@ describe('CombinerMappingManager', () => {
     cy.intercept('/api/v1/management/protocol-adapters/adapters/my-other-adapter/tags', {
       items: MOCK_DEVICE_TAGS('my-other-adapter', MockAdapterType.OPC_UA),
     })
+    cy.intercept('/api/v1/management/protocol-adapters/writing-schema/**', { body: {} })
     cy.intercept('/api/v1/management/topic-filters', {
       items: [
         MOCK_TOPIC_FILTER,
@@ -183,7 +185,7 @@ describe('CombinerMappingManager', () => {
       ],
     })
     cy.intercept('/api/v1/management/protocol-adapters/types', { items: [mockProtocolAdapter_OPCUA] })
-    cy.intercept('api/v1/management/protocol-adapters/adapters', {
+    cy.intercept('/api/v1/management/protocol-adapters/adapters', {
       items: [
         {
           ...mockAdapter,
@@ -197,7 +199,7 @@ describe('CombinerMappingManager', () => {
         },
       ],
     })
-    cy.intercept('PUT', 'api/v1/management/combiners/**', { updated: 'the combiner' }).as('update')
+    cy.intercept('PUT', '/api/v1/management/combiners/**', { updated: 'the combiner' }).as('update')
 
     cy.mountWithProviders(<CombinerMappingManager />, {
       routerProps: { initialEntries: [INITIAL_ENTRY] },
@@ -205,7 +207,15 @@ describe('CombinerMappingManager', () => {
         {
           id: 'idCombiner',
           type: NodeTypes.COMBINER_NODE,
-          data: mockEmptyCombiner,
+          data: {
+            ...mockEmptyCombiner,
+            sources: {
+              items: [
+                { type: EntityType.ADAPTER, id: 'my-adapter' },
+                { id: IdStubs.EDGE_NODE, type: EntityType.EDGE_BROKER },
+              ],
+            },
+          },
           ...MOCK_DEFAULT_NODE,
           position: { x: 0, y: 0 },
         },


hivemq-edge-frontend/src/modules/Mappings/combiner/CombinedSchemaLoader.tsx
```diff
@@ -7,6 +7,7 @@ import { DataIdentifierReference } from '@/api/__generated__'
 import { useGetCombinedDataSchemas } from '@/api/hooks/useDomainModel/useGetCombinedDataSchemas'
 import ErrorMessage from '@/components/ErrorMessage'
 import { PLCTag, TopicFilter } from '@/components/MQTT/EntityTag'
+import { formatOwnershipString } from '@/components/MQTT/topic-utils'
```
oli-hivemq: I'm not seeing a new file topic-utils or a new function `formatOwnershipString` in this PR. It probably means it already exists, or something was missed. I assume it's the former, but I'm just highlighting that.

PR #1431, opened 2026-03-04
fix(EDG-173) Fix AJV "schema already exists" error when uploading destination schema

vanch3d:
**Linear:** https://linear.app/hivemq/issue/EDG-173

## Description

This PR fixes an intermittent error that appeared when uploading a destination schema in the mapping editor:

> `Error: schema with key or id "..." already exists`

The error was thrown by AJV whenever the uploaded schema contained a `$id` property — a common and valid JSON Schema field present in schemas generated by standard tooling or written following best practices.

**Root cause**: `decodeDataUriJsonSchema()` called `validator.ajv.compile(json)` purely to validate schema structure (the return value was discarded). AJV's `compile()` registers schemas by `$id` in its **global singleton** instance. A second call with the same `$id` — caused by a re-render, re-opening the drawer, or a second upload — threw the collision error.

The fix extracts a `cleanSchemaForValidation()` helper that strips `$id` and `$schema` before any `validator.ajv.compile()` call, and applies it consistently across both affected sites. This mirrors the existing pattern RJSF uses internally.

## BEFORE

- Uploading a schema with a `$id` field worked the first time.
- Closing and re-opening the mapping editor drawer, or uploading the same schema again, threw an unhandled AJV error.
- `ChakraRJSForm` already stripped `$schema` before compile but not `$id`.

## AFTER

- Schemas with `$id` can be uploaded, replaced, and re-decoded without errors.
- `cleanSchemaForValidation()` is the single canonical place to sanitise schemas before AJV structural validation.
- Both `topic-filter.schema.ts` and `ChakraRJSForm.tsx` use the same helper — consistent behaviour across the codebase.

## Test Coverage

- **15 Vitest unit tests, all passing ✅** (`topic-filter.schema.spec.ts`)
- **New regression test**: calls `decodeDataUriJsonSchema` twice with a schema that has `$id: 'urn:test:edg-173-duplicate-id'`. Would fail before the fix; passes after.

## Breaking Changes

None. The schema stored in `formData` retains its `$id` unchanged. Only the schema copy passed to AJV compile is sanitised.

## Reviewer Notes

**Focus area:** The fix is contained to `validation.utils.ts` (new helper), `topic-filter.schema.ts` (one call site), and `ChakraRJSForm.tsx` (one call site). No behaviour change for schemas without `$id`.

**To reproduce the original bug before this PR:**

1. Prepare a schema file with `"$id": "urn:any:value"` at the root.
2. Open the mapping editor, set a destination topic, click Upload Schema.
3. Upload the schema — succeeds.
4. Close and re-open the drawer, or upload the same file again.
5. Observe the AJV error (console + error boundary).

**Test command:**

```bash
pnpm test src/modules/TopicFilters/utils/topic-filter.schema.spec.ts
```

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/protocols/northbound/NorthboundTagConsumer.java
+++ b/hivemq-edge/src/main/java/com/hivemq/protocols/northbound/NorthboundTagConsumer.java
@@ -32,11 +32,6 @@ import com.hivemq.edge.modules.adapters.data.DataPointImpl;
 import com.hivemq.edge.modules.adapters.impl.ProtocolAdapterPublishServiceImpl;
 import com.hivemq.edge.modules.api.events.model.EventImpl;
 import com.hivemq.protocols.ProtocolAdapterWrapper;
-import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;


hivemq-edge-frontend/src/modules/TopicFilters/utils/topic-filter.schema.ts
```diff
@@ -44,7 +45,9 @@ export const decodeDataUriJsonSchema = (dataUrl: string) => {
     const json: RJSFSchema = JSON.parse(decoded)
 
     // This will take care of some of the basic json error but not of a valid JSONSchema
-    validator.ajv.compile(json)
+    // $id and $schema are stripped so AJV does not register this schema in its global singleton cache
```
mschoenert-hivemq: ⏺ That's actually not correct — AJV does cache schemas without a $id, using the schema object reference as the cache  
  key. That's precisely why removeSchema(schemaObject) works for anonymous schemas.                                  
                                                                                                                      
  You can verify this behavior:
                                                                                                                      
  const schema = { type: 'object', properties: { name: { type: 'string' } } }

  const v1 = ajv.compile(schema)  // compiles + caches by object reference
  const v2 = ajv.compile(schema)  // returns cached result, no recompile

  console.log(v1 === v2)  // true — same cached function

  But with a structurally identical but different object:

  const schema2 = { type: 'object', properties: { name: { type: 'string' } } }
  const v3 = ajv.compile(schema2)  // new object reference → recompiles

  console.log(v1 === v3)  // false

  So in the Edge code, every call to decodeDataUriJsonSchema with a freshly JSON.parse()-d object creates a new
  reference, meaning each upload compiles and caches a new entry. That's actually a mild concern — repeatedly
  uploading schemas would accumulate entries in the cache with no cleanup. Adding a removeSchema(json) call after the
  compile-for-validation step would address that.

hivemq-edge-frontend/src/modules/TopicFilters/utils/topic-filter.schema.ts
```diff
@@ -44,7 +45,9 @@ export const decodeDataUriJsonSchema = (dataUrl: string) => {
...
+    // (compile is called only for its throw-on-invalid side effect; the return value is discarded)
+    validator.ajv.compile(cleanSchemaForValidation(json))
```
mschoenert-hivemq: Please change this to: 
```suggestion
    const jsonSchemaWithoutId = cleanSchemaForValidation(json)
    validator.ajv.compile(jsonSchemaWithoutId)
    validator.ajv.removeSchema(jsonSchemaWithoutId)
```

to ensure that the schema cache is not polluted.

PR #1432, opened 2026-03-04
fix(EDG-164) Fix mapping integration points deriving from deprecated sources array

vanch3d:
**Linear:** https://linear.app/hivemq/issue/EDG-164

## Description

This PR fixes two related bugs in the mapping editor where selected integration points were read from the deprecated `sources.tags` / `sources.topicFilters` arrays instead of the authoritative `instructions[].sourceRef` records.

**Bug 1 — Wrong chips displayed**: When opening an existing mapping, an unscoped tag chip (`temperature`) could appear alongside scoped ones (`adapter-a :: temperature`), because `CombinedEntitySelect` fell back to the deprecated flat array during the initial render before the context was fully hydrated.

**Bug 2 — Missing schema panel**: When two adapters share a tag name (e.g. both have a `temperature` tag), only one schema panel loaded. The reconstruction logic used `Array.find()` on tag name, so both instructions resolved to the first adapter's scope — deduplicated down to one schema request.

The fix removes all frontend reads and writes of `sources.tags` / `sources.topicFilters` and makes `instructions[].sourceRef` the single source of truth throughout.

The fix also updates the computation of the status of an adapter

## BEFORE

- Opening a mapping with same-named tags from two adapters showed one chip and loaded one schema instead of two.
- A bare unscoped chip could appear in the multi-select alongside the correct scoped chips.
- Old data without instructions (deprecated format) incorrectly showed chips from the flat tag array.

## AFTER

- Integration points always derive from `instructions[].sourceRef`, which carries full scope (`adapterId`).
- Two adapters sharing `temperature` produce two distinct chips (`o1 :: temperature`, `o2 :: temperature`) and load two separate schemas.
- Old data without instructions shows an empty selection — the correct and safe behaviour.

## Test Coverage

- **25 tests, all passing ✅**
- **12 Cypress component tests** (`DataCombiningEditorField.spec.cy.tsx`): 7 EDG-164 scenario tests (S1–S7) covering fresh mapping, scoped instructions, same-named tags across adapters, deprecated-only data, partial data, topic filters, and mixed sources.
- **9 Cypress component tests** (`CombinedEntitySelect.spec.cy.tsx`): chip display, ownership strings, duplicate adapter names, accessibility.
- **4 Cypress component tests** (`SchemaMerger.spec.cy.tsx`): no regression.

## Breaking Changes

None. The deprecated `sources.tags` / `sources.topicFilters` fields remain in the TypeScript type (they are backend-generated). Only frontend reads and writes are removed.

## Reviewer Notes

**Focus areas:**

1. `reconstructSelectedSources` now reads exclusively from `instructions[].sourceRef` — verify S3 scenario (same tag name, two adapters) works end-to-end.
2. `getFilteredDataReferences` no longer has a Phase 1 fallback — only `formContext.selectedSources` is used.
3. Old mappings without instructions (S4) now show empty selection, which is the correct migration behaviour.

**Manual testing:**

1. Open the mapping editor on a combiner with two adapters that share a tag name.
2. Add an instruction for `temperature` on adapter-1 and another for `temperature` on adapter-2.
3. Save, re-open: verify two scoped chips appear and two schema panels load.

**Test command:**

```bash
pnpm cypress:run:component --spec "src/modules/Mappings/combiner/DataCombiningEditorField.spec.cy.tsx"
```

### DIFF
--- a/hivemq-edge-frontend/src/modules/Mappings/utils/combining.utils.ts
+++ b/hivemq-edge-frontend/src/modules/Mappings/utils/combining.utils.ts
@@ -1,4 +1,7 @@
+import debug from 'debug'
 import type { UseQueryResult } from '@tanstack/react-query'
+
+const log = debug('Mappings:combining:utils')
 
 import {
   DataIdentifierReference,
@@ -260,6 +263,9 @@ export const reconstructSelectedSources = (
   // Also prepend scoped primary if not already represented
   const primary = formData.sources.primary
   const all: DataIdentifierReference[] = []
+  if (primary?.type === DataIdentifierReference.type.TAG && !primary.scope) {
+    log('TAG primary has no scope — skipped during selectedSources reconstruction %o', primary)
+  }
   if (primary?.scope) all.push(primary)
   all.push(...fromInstructions)
 


mschoenert-hivemq:
This involves the interaction with the backend and must be reviewed by the backend.

hivemq-edge-frontend/src/modules/Mappings/combiner/SchemaMerger.tsx
```diff
@@ -51,8 +51,16 @@ const SchemaMerger: FC<SchemaMergerProps> = ({ formData, formContext, onClose, o
   const properties = useMemo(() => {
     const displayedSchemas = getSchemasFromReferences(references, schemaQueries)
 
-    const tagIndexMap = new Map(formData?.sources?.tags?.map((tag, index) => [tag, index]))
-    const topicFilterIndexMap = new Map(formData?.sources?.topicFilters?.map((filter, index) => [filter, index]))
+    const tagIndexMap = new Map(
+      references
+        .filter((r) => r.type === DataIdentifierReference.type.TAG)
+        .map((ref, index) => [`${ref.id}@${ref.scope ?? ''}`, index])
```
caoccao: What if `ref.scope` is `undefined`?

caoccao (commented):
I left 2 comments.

hivemq-edge-frontend/src/modules/Mappings/utils/combining.utils.ts
```diff
@@ -252,75 +238,39 @@ export const getAdapterIdForTag = (tagId: string, formContext?: CombinerContext)
...
+
+  // Also prepend scoped primary if not already represented
+  const primary = formData.sources.primary
+  const all: DataIdentifierReference[] = []
+  if (primary?.scope) all.push(primary)
+  all.push(...fromInstructions)
+
+  // Deduplicate by id + type + scope
+  const unique = all.reduce<DataIdentifierReference[]>((acc, ref) => {
+    const exists = acc.find((r) => r.id === ref.id && r.type === ref.type && r.scope === ref.scope)
```
oli-hivemq: This dedup relies entirely on `scope` being populated to distinguish same-named tags from different adapters. If `scope` is ever `undefined` (legacy data, API not sending it), two distinct adapter refs collapse to one - silently re-introducing Bug 2. Should we add a defensive guard or at minimum a `console.warn` when a TAG-type sourceRef has no scope?

vanch3d: Custom validation in the RJSF should already alert the users about a misconfiguration. But I think it's worth "de-silencing"  the defensive behavior

PR #1461, opened 2026-03-17
chore: reduce trigger job build retention from 100 to 10

marregui:
## Summary
- Reduce `numToKeepStr` from `100` to `10` in trigger job `CompositeJenkinsfile` and `modules/hivemq-edge-module-opcua/CompositeJenkinsfile`
- Trigger jobs are lightweight (just kick off the composite build) — 100 builds of history per branch is excessive
- Aligns with the platform standard of 10 builds for trigger jobs

Linear: EDG-285


caoccao (approved):
LGTM

PR #1466, opened 2026-03-18
Chore/edg 185/integrate nullaway

codepitbull:
**Motivation**

Resolves #EDG-185

**Changes**
- Add a convention plugin to integrate NullAway
- Fix all reported issues


hivemq-edge/src/main/java/com/hivemq/api/auth/ApiPrincipal.java
```diff
@@ -20,14 +20,15 @@
...
 
 /**
  * @author Simon L Johnson
  */
 public class ApiPrincipal implements Principal {
 
-    private @NotNull String name;
-    private Set<String> roles;
+    private @Nullable String name;
+    private @Nullable Set<String> roles;
```
caoccao: I don't think this is the right fix. The setter should be fixed.

Image

PR #1469, opened 2026-03-20
fix: Migrate primary TAG reference scope in DataCombiningScopeMigrator

caoccao:
DataCombiningScopeMigrator.migrateDataCombining() only migrated TAG
  references found in instructions but skipped the primaryReference in
  DataCombining.sources(). This meant that legacy configs with an unscoped
  TAG as the primary trigger were never backfilled with the correct adapter
  scope, causing the runtime to fail to resolve the primary subscription.

  Extract the scope-backfill logic into a reusable migrateTagReference()
  method and apply it to both the primaryReference and each instruction's
  dataIdentifierReference. When the primary is migrated, a new
  DataCombiningSources is constructed preserving the existing tags and
  topicFilters lists.

  Add DataCombiningScopeMigratorTest with 11 tests covering:
  - Primary TAG without scope: single adapter (migrated), multiple
    adapters (warning, not migrated), tag not found (warning, not migrated)
  - Primary TAG already scoped or TOPIC_FILTER type (no migration)
  - Combined scenarios: both primary and instruction unscoped, only primary
    unscoped, only instruction unscoped
  - Sources metadata (tags/topicFilters lists) preserved after migration
  - Instruction with null ref unchanged
  - AssetMappingExtractor receives migrations alongside DataCombiningExtractor


hivemq-edge/src/main/java/com/hivemq/configuration/migration/DataCombiningScopeMigrator.java
```diff
@@ -158,15 +148,25 @@ private void migrateExtractor(
...
     private @NotNull DataCombining migrateDataCombining(
             final @NotNull DataCombiner combiner,
             final @NotNull DataCombining dataCombining,
             final @NotNull Map<String, List<String>> tagToAdapters) {
 
         boolean anyChanges = false;
-        final List<Instruction> migratedInstructions = new ArrayList<>();
 
+        // Migrate the primary reference
+        final DataIdentifierReference primaryRef = dataCombining.sources().primaryReference();
```
marregui: is there any way that an old config may return null for primaryReference ?

PR #1489, opened 2026-04-02
EDG-217: add frontend UI for Houston device tag browse/import

marregui:
Adds a "Device Tags" tab to the OPC-UA adapter configuration drawer. The tab contains two sections:
- Browse & Download: calls POST /device-tags/browse and saves the result as a CSV/JSON/YAML file
- Import: uploads a file to POST /device-tags/import with configurable conflict-resolution mode

The tab is shown only when editing an existing OPC-UA adapter (hardcoded adapter type check; can be replaced with a capability flag later).

ObjectFieldTemplate is extended to accept extraTabs from formContext, allowing arbitrary React content to be injected as additional tabs alongside the schema-driven ones.

**Motivation**

Resolves #<issue>

**Changes**


marregui:
cherry-picked the UI

PR #1493, opened 2026-04-09
EDG-263 Edge - Implementation of standardized schema generation

codepitbull:
**Motivation**

Resolves #EDG-263

**Changes**
Implementation of the new schema handling


hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java
```diff
@@ -141,6 +142,7 @@ public class ConfigFileReaderWriter {
                     .toArray(new Class<?>[0]));
         } catch (final Throwable e) {
             log.error("Cannot create the jaxb context:", e);
+            log.error("argh {}", ((IllegalAnnotationsException) e).getErrors());
```
caoccao: What's `argh`?

caoccao (approved):
Only one small question. LGTM

hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java
codepitbull: A statement of my desperation because of a pulse change ...

I just turned it into a proper error output to help identifying config reading issues.

caoccao: I thought about [https://en.wiktionary.org/wiki/argh](<https://en.wiktionary.org/wiki/argh>) but wasn't sure. 🤣

> In imitation of a cry. Used since at least the 18th century.\[1\] Compare ah, of which it could be a lengthened form.

PR #1495, opened 2026-04-14
EDG-463 - fix validation of Edge Broker in Data Combiner creation

mschoenert-hivemq:
## Summary

- Fixes a regression introduced by the tag scoping epic (PR #1434) that made it impossible to create a Data Combiner through the "Create New > Combiner" wizard
- The validation in `useValidateCombiner.ts` was checking `formData.sources.items` for the presence of EDGE_BROKER, which relied on a mutation side-effect that PR #1434 correctly removed
- Fix: check `entities` (the hook parameter) instead, which always contains EDGE_BROKER regardless of how the form was opened

## Root cause

Before PR #1434, `entities.push(EDGE_BROKER)` mutated `selectedNode.data.sources.items` in-place, so `formData.sources.items` accidentally contained EDGE_BROKER. After the fix (new array returned), the mutation was gone and validation broke.

## Test plan

- [ ] Open the frontend and click "Create New > Combiner" — select one OPC-UA adapter, proceed through the wizard, verify the form submits successfully (no "Edge broker must be connected" error)
- [ ] Verify the canvas multi-select flow (select adapter + Edge Broker node, click combiner icon) still works as before

### DIFF
--- a/hivemq-edge-frontend/src/modules/Mappings/hooks/useValidateCombiner.spec.ts
+++ b/hivemq-edge-frontend/src/modules/Mappings/hooks/useValidateCombiner.spec.ts
@@ -155,25 +155,20 @@ describe('useValidateCombiner', () => {
 
     it('should fail to validate an adapter without combine but validate the Edge', async () => {
       server.use(...failCapabilityHandlers)
-      const errors = await renderValidateHook({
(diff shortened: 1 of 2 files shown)


caoccao (approved):
LGTM

PR #1496, opened 2026-04-14
EDG-471 Improve Data Combiner source entity selection UX

mschoenert-hivemq:
## Summary

Implements EDG-471: four UX improvements to Data Combiner source entity selection.

- **Allow single-source combiners** — `minNodes` changed from 2 to 1 in wizard metadata; combining tags from a single adapter is valid.
- **Edge Broker selectable everywhere** — removed three separate hard-exclusions (wizard `allowedNodeTypes`/`customFilter`, `ReactFlowWrapper` click handler, `toolbar.utils` candidate filter) so the Edge Broker can be explicitly added as a source in both the wizard and the canvas toolbar.
- **Edge Broker is optional** — removed the mandatory `hasEdge.length !== 1` validation from `useValidateCombiner`; the Edge Broker is only needed when the user wants topic filters as inputs.
- **"Add source" dropdown in Sources tab** — `EntityReferenceTableWidget` now shows a `Select` dropdown that lets users add additional source entities to an existing combiner without a REST call; entities are persisted on Submit.

### Notable implementation detail — RJSF controlled form

The hardest part was preventing sources from disappearing after being added. Root cause: RJSF's `getSnapshotBeforeUpdate` fires on **any** prop change (including `formContext`). It calls `getStateFromProps(this.props.formData)` and if the result differs from the current internal state, it resets the form. Since `formContext` legitimately changes when entity queries resolve after a source is added, the form was silently resetting every time.

Fix: controlled form pattern in `CombinerMappingManager` — `currentFormData` state is kept in sync with RJSF's internal state via `onChange`. When RJSF recomputes `nextState` from the prop, it matches `prevState`, so `shouldUpdate = false` and no reset occurs.

The `onSourcesChange` callback remains the dedicated path for updating `liveSources` (and thus `entityQueries`/`formContext`), kept separate from the full-form-data tracking.

### Out of scope

Point 5 (load-time entity sync) is deferred to EDG-473.

## Test plan

- [ ] Wizard: can select only Edge Broker → combiner created with Edge Broker as sole source
- [ ] Wizard: can select a single adapter (no Edge Broker) → combiner created, no validation error about missing Edge Broker
- [ ] Canvas toolbar: selecting only Edge Broker enables the combiner button
- [ ] Canvas toolbar: clicking combiner button with Edge Broker selected creates a combiner with Edge Broker as source
- [ ] Sources tab: "Add source" dropdown shows only entities not already in the sources list
- [ ] Sources tab: adding a source persists it in the table; it does not disappear after a second
- [ ] Sources tab: removing a source re-adds it to the dropdown
- [ ] Mapping tab: after adding a source in the Sources tab, the integration point selectors reflect the new entity's tags/topic filters
- [ ] Existing combiners: open an existing combiner → no spurious validation error about Edge Broker
- [ ] Submit: saving a combiner correctly persists the updated sources list


hivemq-edge-frontend/src/modules/Mappings/combiner/EntityReferenceTableWidget.tsx
```diff
@@ -29,33 +55,47 @@ export const EntityReferenceTableWidget = (props: WidgetProps<WidgetProps<Array<
...
                 />
               )}
             </ButtonGroup>
           )
         },
       },
     ]
-  }, [t])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [t, props.value])
```
caoccao: Is this change necessary?

hivemq-edge-frontend/src/modules/Mappings/CombinerMappingManager.tsx
```diff
@@ -140,24 +140,55 @@ const CombinerMappingManager: FC<CombinerMappingManagerProps> = ({ wizardContext
...
-    if (!isBridgeIn) {
-      return [...sourceItems, { id: IdStubs.EDGE_NODE, type: EntityType.EDGE_BROKER }]
-    }
-    return sourceItems
-  }, [selectedNode.data.sources.items])
+  // Track live sources so that adding/removing a source in the Sources tab
+  // immediately updates entity queries and the Mapping tab's integration point selectors.
+  // This state is updated ONLY via onSourcesChange (called explicitly by EntityReferenceTableWidget)
+  // rather than via form onChange, which would fire on every keystroke and risk resetting the form.
+  const [liveSources, setLiveSources] = useState<EntityReference[]>(selectedNode.data.sources.items || [])
```
caoccao: Better make a copy of `selectedNode.data.sources.items`.

caoccao (approved):

hivemq-edge-frontend/src/modules/Mappings/combiner/EntityReferenceTableWidget.tsx
mschoenert-hivemq: fair point — the eslint-disable comment is a code smell. Here's what's happening:                          
                    
displayColumns captures handleDelete in its closure (line 65: onClick={() => handleDelete(...)}. handleDelete itself uses props.value to filter the sources array. If props.value isn't in the useMemo deps, the delete callback would close over stale props.value after the first render — clicking delete on a source added after mount would silently fail.                       
                                                                                                                                
So props.value is necessary. But the cleaner fix is to move handleDelete inside the useMemo, so ESLint can verify the deps automatically and the comment isn't needed:

hivemq-edge-frontend/src/modules/Mappings/CombinerMappingManager.tsx
mschoenert-hivemq: Good catch.

hivemq-edge-frontend/src/modules/Mappings/combiner/EntityReferenceTableWidget.spec.cy.tsx
```diff
@@ -65,7 +65,7 @@ describe('EntityReferenceTableWidget', () => {
   })
 
   it('should render permanent ', () => {
-    const v = { ...MOCK_ENTITY_PROPS }
+    const v = { ...MOCK_ENTITY_PROPS, value: [...(MOCK_ENTITY_PROPS.value as Array<EntityReference>)] }
```
caoccao, line 61: That smells bad. `structuredCopy` would work.

caoccao (approved):
One more comment.

hivemq-edge-frontend/src/modules/Mappings/combiner/EntityReferenceTableWidget.tsx
caoccao: Yes, I prefer a cleaner solution.

hivemq-edge-frontend/src/modules/Mappings/combiner/EntityReferenceTableWidget.spec.cy.tsx
mschoenert-hivemq: I feel I'm learning more typescript than I ever wanted to know. Thanks. ;-)

mschoenert-hivemq: But then testing it revealed: 

structuredClone cannot clone function values (onChange). Spread only the
value array to avoid mutating the shared mock, and add a comment explaining
why structuredClone is not viable here.

caoccao: Oh, that is even worse. It's supposed to be a pure data object. Now it carries functions. That's a weird design. But, we are not going to fix that. Let's keep this smelly fix.

hivemq-edge-frontend/src/modules/Mappings/combiner/EntityReferenceTableWidget.tsx
mschoenert-hivemq: This is now in the code (with handleDelete inside the useMemo).

hivemq-edge-frontend/src/modules/Mappings/combiner/EntityReferenceTableWidget.spec.cy.tsx
mschoenert-hivemq: Indeed.

hivemq-edge-frontend/src/modules/Mappings/CombinerMappingManager.tsx
mschoenert-hivemq: Done.

PR #1497, opened 2026-04-14
feat(EDG-471): improve Data Combiner source entity selection UX

mschoenert-hivemq:
## Summary

Implements EDG-471: four UX improvements to Data Combiner source entity selection.

- **Allow single-source combiners** — `minNodes` changed from 2 to 1 in wizard metadata; combining tags from a single adapter is valid.
- **Edge Broker selectable everywhere** — removed three separate hard-exclusions (wizard `allowedNodeTypes`/`customFilter`, `ReactFlowWrapper` click handler, `toolbar.utils` candidate filter) so the Edge Broker can be explicitly added as a source in both the wizard and the canvas toolbar.
- **Edge Broker is optional** — removed the mandatory `hasEdge.length !== 1` validation from `useValidateCombiner`; the Edge Broker is only needed when the user wants topic filters as inputs.
- **"Add source" dropdown in Sources tab** — `EntityReferenceTableWidget` now shows a `Select` dropdown that lets users add additional source entities to an existing combiner without a REST call; entities are persisted on Submit.

### Notable implementation detail — RJSF controlled form

The hardest part was preventing sources from disappearing after being added. Root cause: RJSF's `getSnapshotBeforeUpdate` fires on **any** prop change (including `formContext`). It calls `getStateFromProps(this.props.formData)` and if the result differs from the current internal state, it resets the form. Since `formContext` legitimately changes when entity queries resolve after a source is added, the form was silently resetting every time.

Fix: controlled form pattern in `CombinerMappingManager` — `currentFormData` state is kept in sync with RJSF's internal state via `onChange`. When RJSF recomputes `nextState` from the prop, it matches `prevState`, so `shouldUpdate = false` and no reset occurs.

The `onSourcesChange` callback remains the dedicated path for updating `liveSources` (and thus `entityQueries`/`formContext`), kept separate from the full-form-data tracking.

### Out of scope

Point 5 (load-time entity sync) is deferred to EDG-473.

## Test plan

- [ ] Wizard: can select only Edge Broker → combiner created with Edge Broker as sole source
- [ ] Wizard: can select a single adapter (no Edge Broker) → combiner created, no validation error about missing Edge Broker
- [ ] Canvas toolbar: selecting only Edge Broker enables the combiner button
- [ ] Canvas toolbar: clicking combiner button with Edge Broker selected creates a combiner with Edge Broker as source
- [ ] Sources tab: "Add source" dropdown shows only entities not already in the sources list
- [ ] Sources tab: adding a source persists it in the table; it does not disappear after a second
- [ ] Sources tab: removing a source re-adds it to the dropdown
- [ ] Mapping tab: after adding a source in the Sources tab, the integration point selectors reflect the new entity's tags/topic filters
- [ ] Existing combiners: open an existing combiner → no spurious validation error about Edge Broker
- [ ] Submit: saving a combiner correctly persists the updated sources list

### DIFF
--- a/hivemq-edge-frontend/src/locales/en/translation.json
+++ b/hivemq-edge-frontend/src/locales/en/translation.json
@@ -1769,6 +1769,7 @@
       },
       "sources": {
         "description": "The list of data sources available for this combiner",
+        "addSource": "Add a source...",
         "table": {
           "action": "Actions",
           "delete": "Delete the source",
--- a/hivemq-edge-frontend/src/modules/Mappings/CombinerMappingManager.tsx
+++ b/hivemq-edge-frontend/src/modules/Mappings/CombinerMappingManager.tsx
@@ -152,10 +152,17 @@ const CombinerMappingManager: FC<CombinerMappingManagerProps> = ({ wizardContext
 
   const entities = useMemo(() => liveSources, [liveSources])
 
-  // Stable formData reference: keyed to the combiner ID so RJSF's internal state is
-  // never reset by re-renders (e.g. React Flow recreating node objects for position updates).
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  const initialFormData = useMemo(() => selectedNode.data, [combinerId])
+  // Controlled form data: kept in sync with RJSF's internal state via onChange.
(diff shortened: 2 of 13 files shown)


mschoenert-hivemq:
Duplicate of #1496 — closing.

PR #1498, opened 2026-04-16
EDG-486 Houston follow-up: browse/import throughput + browse stability fixes

marregui:
## Summary

Houston follow-up on the OPC UA browse/import path: four server-safe throughput changes plus two browse-stability fixes found while load-testing this branch against Prosys. The slow-but-correct defaults established by [EDG-465](https://linear.app/hivemq/issue/EDG-465) (continuation-point fix on Siemens S7-1500) — `MAX_CONCURRENT_BROWSES = 1`, `READ_BATCH_SIZE = 100`, full-Phase-1-before-Phase-2 — are preserved. Nothing here adds concurrent server requests, widens batch sizes, or auto-tunes against advertised server capabilities.

Parent ticket: [EDG-486 — Tag Browsing and Discovery OPC-UA - Houston - Follow-up](https://linear.app/hivemq/issue/EDG-486). Sub-tickets EDG-493 (additive-only classification) and EDG-576/577 landed earlier via #1579; EDG-488 (adapter-level `DataTypeTree` cache) is cancelled — Milo 1.1.6 caches the tree per client and the EDG-577 warm-up pre-builds it on connect. EDG-492 (per-adapter import lock) is dropped after review, see below.

### Throughput (one sub-ticket each)

| Ticket | Change |
|--------|--------|
| [EDG-487](https://linear.app/hivemq/issue/EDG-487) | `BatchAttributeSpliterator` fires `readAsync` for batch N+1 as soon as batch N's response lands, instead of blocking inside `tryAdvance` after the consumer drains batch N. Milo's channel stays serial, so the server sees the same request rate; only client-side idle gaps disappear. `pendingBatchSize` keeps `estimateSize()` exact under the `SIZED` contract. |
| [EDG-489](https://linear.app/hivemq/issue/EDG-489) | CSV streaming output wrapped in a `BufferedWriter`. |
| [EDG-490](https://linear.app/hivemq/issue/EDG-490) | JSON output is compact (`INDENT_OUTPUT` removed); roughly halves wire size, round-trip unchanged. |
| [EDG-491](https://linear.app/hivemq/issue/EDG-491) | `OpcUaNodeBrowser.sanitize()` is a single character-walking pass, no regex. Same observable behaviour, 13 existing + 6 new tests. |

### Stability fixes (found by the load test below; both pre-date this branch)

- **`tag_name_default` collision suffixes were non-deterministic.** Phase 1 sorted discovered variables by path only, so nodes sharing a browse path (Prosys `ValueSimulation` instances, for example) kept the order the async browse callbacks delivered them in, and the `-2/-3/-4` suffixes shuffled between browses (~1 in 30 under load). A CSV exported from one browse could therefore name a different node than the next browse would. Now tie-broken on the `NodeId`. Regression test asserts the same assignment for both delivery orders; verified live with 160 concurrent browses across four adapters producing one identical `nodeId → tag_name_default` map.
- **A read failing mid-stream returned a well-formed JSON document with fewer rows under HTTP 200.** The status is committed at the first flush, and `JsonGenerator.close()` on the exception path auto-terminated the open array and object, so the client could not tell 500 rows from 886. Observed once when an adapter was stopped during Phase 2 (`Bad_SessionClosed`). `AUTO_CLOSE_JSON_CONTENT` is now off for the streaming path, so a truncated document is left unterminated and fails to parse. CSV and YAML remain silently truncatable in this situation — a truncated block sequence is still valid YAML — and are not addressed here.

### Dropped after review: EDG-492 per-adapter import lock

The first cut replaced `synchronized (adapterExtractor)` around the import's read-compute-write cycle with a per-adapter lock so imports on different adapters could overlap. Sam flagged that the lock map was never cleaned up when an adapter was removed. Rather than fix the leak, the change is reverted and `DeviceTagImporter` is byte-identical to master again, because the win it promised does not exist here:

- the expensive part of an import — `updateAdapter`, which swaps the adapter and writes `config.xml` — is itself `synchronized` on the extractor, so imports on different adapters were serialised there anyway; only the millisecond validation phase could overlap;
- the admin API runs on two HTTP threads, so at most two imports are ever in flight;
- the extractor monitor also kept REST writes, config reloads and `updateAllAdapters` from interleaving with an import's cycle, which a per-adapter lock does not.

Measured on the Prosys node, fresh start per run, three runs each: four 60-tag imports on four adapters in parallel complete in **0.08–0.09 s wall with the per-adapter lock and 0.08 s with the extractor monitor**; four on one adapter, 0.05–0.06 s both ways. Per-import 50–90 ms. The synchronisation is not where the time goes.

Coverage added instead: two `DeviceTagImporterTest` cases drive the importer through a *stateful* extractor fake (reads return the last write, reads are slowed to widen the race) — 16 concurrent imports on one adapter must all land, and 8 adapters × 4 imports must each end with exactly their own tags. Both fail with the `synchronized` block removed (2 of 16 tags survive), both pass with it.

## Out of scope (deliberate)

- Increasing `MAX_CONCURRENT_BROWSES` (semaphore stays at 1 — the S7-1500 lesson holds).
- Auto-tuning `READ_BATCH_SIZE` from server-advertised `MaxNodesPerRead` (servers misadvertise).
- Pipelining Phase 1 + Phase 2 streaming (changes ordering contract; needs its own ticket).
- Dropping `CopyOnWriteArrayList`/`ConcurrentHashMap.newKeySet()` for plain `ArrayList`/`HashSet` — async callbacks can race even with semaphore=1, so the existing concurrent collections are correct.
- The admin API's fixed pool of **two** HTTP threads (`InternalConfigurations.HTTP_API_THREAD_COUNT`, not configurable). Verified with thread dumps under load: at most two browses ever execute concurrently whatever the adapter count, and every other REST call queues behind them. That caps cross-adapter parallelism on this path and belongs to the v2 design, not this PR.

## Test plan

- [x] Unit: `hivemq-edge-module-opcua` (1259) and `com.hivemq.edge.adapters.browse.*` (229) green on Temurin 25. All four new tests were mutation-checked: each fails with its fix (or the lock) removed.
- [x] Integration: `DeviceTagBrowseIT`, `DeviceTagImportIT`, `DeviceTagEndToEndIT`, `OpcUaNodeBrowserConcurrencyIT` (56) green from `hivemq-edge-test` master on JDK 21, `--no-build-cache`.
- [x] `spotlessJavaCheck` clean on `hivemq-edge` and `hivemq-edge-module-opcua`.
- [x] Load test against Prosys Simulation Server (886–898 nodes, ~500 KB JSON), branch core + module jars, four adapters:
  - sequential p50 **308 ms vs 338 ms** with the master module jar swapped in (A/B on the same node); the 2× Phase-2 claim needs a server with tens of ms RTT — on localhost Phase 1 dominates.
  - 8 concurrent browses on one adapter: p50 2.3 s, all 200, byte-identical.
  - 16 concurrent across four adapters: p50 3.7–4.0 s, 3.85 browses/s (capped by the two HTTP threads above).
  - chaos, 75 s, 8 workers while two adapters are stopped/started three times, 60-tag imports run in `MERGE_SAFE` and `OVERWRITE`, and an adapter is deleted and recreated: 526 requests → 427×200, 90×503 `AdapterBrowseNotReady` (first 200 within 1 s of START, 1.6 s of CREATE), 7×404 while deleted, 2×409 for browses in flight when their adapter stopped. No 500s, no hangs, no thread leak.
  - parallel imports with the extractor monitor vs the per-adapter lock: no measurable difference (numbers above).
- [ ] QA: smoke-test against live S7-1500 (SUGUS adapter) — full-depth browse should still return ~225 nodes consistently.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/edge/adapters/browse/importer/DeviceTagImporter.java
+++ b/hivemq-edge/src/main/java/com/hivemq/edge/adapters/browse/importer/DeviceTagImporter.java
@@ -62,6 +62,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.stream.Collectors;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -78,6 +80,11 @@ public class DeviceTagImporter {
 
     private final @NotNull ProtocolAdapterExtractor adapterExtractor;
     private final @NotNull DataCombiningExtractor combiningExtractor;
+
+    // Per-adapter locks so concurrent imports on different adapters can proceed in parallel.
+    // Previously the entire import serialized on the shared adapterExtractor monitor, which
+    // was stricter than needed — TOCTOU protection is per-adapter, not global.
+    private final @NotNull ConcurrentMap<String, Object> perAdapterLocks = new ConcurrentHashMap<>();
 
     @Inject
     public DeviceTagImporter(
@@ -116,10 +123,12 @@ public class DeviceTagImporter {
             final @NotNull String adapterId,
             final @Nullable BulkTagBrowser browser)
             throws DeviceTagImporterException {
-        // Synchronize the entire read-compute-write cycle on the same intrinsic lock used by
-        // ProtocolAdapterExtractor's synchronized methods to prevent TOCTOU races between
-        // concurrent imports (e.g., two OVERWRITE operations reading the same stale state).
-        synchronized (adapterExtractor) {
+        // Per-adapter lock: two imports on the same adapter still serialize (preventing TOCTOU
+        // between concurrent OVERWRITE operations on the same adapter), but imports on
+        // different adapters can proceed in parallel. This matters for multi-adapter
+        // scripted-provisioning workflows where a global lock was a needless bottleneck.
+        final Object lock = perAdapterLocks.computeIfAbsent(adapterId, id -> new Object());
+        synchronized (lock) {
             return doImportLocked(rows, mode, adapterId, browser);
         }
     }
@@ -158,6 +167,23 @@ public class DeviceTagImporter {
 
         // if there was no validation error, now is the time to execute the plan
         final ProtocolAdapterEntity fac = finalAdapterConfiguration(adapter, twmsFinal);
+        // Classify the change so future work can skip the adapter restart for additive-only
+        // imports. Today ProtocolAdapterExtractor.updateAdapter always triggers a full-entity
+        // replacement; the registered consumer then restarts the adapter. A lighter-weight
+        // "addTagsAndMappings" path on the extractor (and a corresponding handler on the
+        // consumer) would let additive-only imports avoid disrupting the running data flow.
+        // Tracked as future optimisation — intentionally not gated here.
+        final boolean additiveOnly =
+                stats.tagsDeleted == 0 && stats.tagsUpdated == 0 && stats.nbDeleted == 0 && stats.sbDeleted == 0;
+        if (additiveOnly && log.isDebugEnabled()) {
+            log.debug(
+                    "Import for adapter '{}' is additive-only ({} tags created, {} NB, {} SB) — "
+                            + "eligible for no-restart path once the extractor supports it",
+                    adapterId,
+                    stats.tagsCreated,
+                    stats.nbCreated,
+                    stats.sbCreated);
+        }
         final boolean success = adapterExtractor.updateAdapter(fac);
         if (!success) {
             throw new DeviceTagImporterException(List.of(new ValidationError(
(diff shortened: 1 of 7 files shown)


hivemq-edge/src/main/java/com/hivemq/edge/adapters/browse/importer/DeviceTagImporter.java
```diff
@@ -116,10 +123,12 @@ public DeviceTagImporter(
...
             throws DeviceTagImporterException {
-        // Synchronize the entire read-compute-write cycle on the same intrinsic lock used by
-        // ProtocolAdapterExtractor's synchronized methods to prevent TOCTOU races between
-        // concurrent imports (e.g., two OVERWRITE operations reading the same stale state).
-        synchronized (adapterExtractor) {
+        // Per-adapter lock: two imports on the same adapter still serialize (preventing TOCTOU
+        // between concurrent OVERWRITE operations on the same adapter), but imports on
+        // different adapters can proceed in parallel. This matters for multi-adapter
+        // scripted-provisioning workflows where a global lock was a needless bottleneck.
+        final Object lock = perAdapterLocks.computeIfAbsent(adapterId, id -> new Object());
```
caoccao: There's no lock clean up, if the adapter ID is removed.

caoccao (approved):
A small issue, but it doesn't block the merge.

hivemq-edge/src/main/java/com/hivemq/edge/adapters/browse/importer/DeviceTagImporter.java
marregui: Good catch. Rather than clean the map up I removed the per-adapter lock altogether (f625c6b2) — `updateAdapter` is synchronized on the extractor anyway and the API has two threads, so it never bought parallelism (measured: 4 parallel imports 0.08 s either way). The extractor monitor is back, plus two stateful concurrency tests that fail without it.

PR #1499, opened 2026-04-16
Fix CVE-2025-14813: upgrade Bouncycastle from 1.83 to 1.84

caoccao:
The GOST-R-3413-2015 CTR implementation (G3413CTRBlockCipher) in Bouncycastle versions 1.59 through 1.83 used a single byte as a counter, limiting encryption/decryption to 255 blocks — well below the algorithm specification.


marregui (approved):
LGTM

PR #1506, opened 2026-04-21
EDG-504: Improve internal subscriptions with InternalTopicFilterSubscriber

mschoenert-hivemq:
## Summary

Introduces `InternalTopicFilterSubscriber` — a proper subscription mechanism for internal Edge components that need to consume MQTT messages from the topic tree (bridge, data combiners, asset mappers, …) without being actual MQTT clients.

The current approach forces all such components to use shared subscriptions, a mechanism designed for cluster-wide fan-out that is meaningless overhead in a single-node Edge deployment. This PR replaces that workaround with a dedicated base class and the two minimal platform changes that make it functional.

- **`InternalTopicFilterSubscriber`** (`com.hivemq.persistence.clientqueue`) — abstract base class owning the full subscription lifecycle: topic tree registration, callback registration, SingleWriter-based poll loop via `readNew(..., shared=false, ...)`, message removal, and retry. Subclasses implement `process(PUBLISH)` only.
- **`PublishDistributorImpl`** — adds `INTERNAL_SUBSCRIBER_PREFIX = "internal#"` and wraps the `ClientSession` lookup in `handlePublish()` so internal subscriber client IDs are not rejected with `NOT_CONNECTED`.
- **`ClientQueuePersistenceImpl`** — extends `publishAvailable()` to check `queueidCallbackMap` before the session/channel path, so internal subscribers receive wake-up notifications when messages arrive.

No existing subscribers are changed by this PR. The class is ready to use; TyneBridge and data combiners will be migrated in follow-up issues.

Linear: https://linear.app/hivemq/issue/EDG-504


hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java
```diff
@@ -0,0 +1,178 @@
...
+ * <p>The client ID has the form {@code "internal#<componentPrefix>#<instanceId>"}, e.g.
+ * {@code "internal#tynebridge#my-bridge"} or {@code "internal#combiner#combiner:123/inst:2"}.
+ *
+ * <p>Call {@link #start()} once to subscribe and {@link #stop()} to unsubscribe. Both methods
+ * are safe to call on any thread.
+ *
+ * @see com.hivemq.mqtt.services.PublishDistributorImpl#INTERNAL_SUBSCRIBER_PREFIX
+ */
+@SuppressWarnings({"FutureReturnValueIgnored", "CheckReturnValue"})
+public abstract class InternalTopicFilterSubscriber {
```
caoccao, line 50: Is this classes being used? Without seeing its usage, I cannot tell if the API is appropriate.

caoccao (commented):
I have one question.

hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java
mschoenert-hivemq: This is how it would be used e.g. in DataCombiningRuntime. It simplifies the current subscription **a lot**. You don't have to know about shared subscription, how to construct the `sharedName` and how it must match the `clientId` and `topicFilter`. Reduces the current code by \~30 lines to the essential parts: defining the subscriber (with a unique identity) and providing the processing method/callback. You can - but often don't have to - override the `start()` and `stop()` methods.

```
        dataCombiningPublishService.publish(
                combining.destination(), rootNode.toString().getBytes(StandardCharsets.UTF_8), dataCombining);
    }

    public interface InternalConsumer {
        void start();

        void stop();
    }

    public final class InternalTopicFilterConsumer extends InternalTopicFilterSubscriber implements InternalConsumer {
        private static final String COMPONENT_PREFIX = "combiner";

        private final @NotNull String topicFilter;
        private final boolean primary;
        private final boolean storeDataPoints;

        public InternalTopicFilterConsumer(
                final @NotNull DataIdentifierReference dataIdentifierReference,
                final boolean primary,
                final boolean storeDataPoints) {
            super(
                    COMPONENT_PREFIX,
                    combining.id() + "/" + dataIdentifierReference.id(),
                    dataIdentifierReference.id(),
                    localTopicTree,
                    clientQueuePersistence,
                    singleWriterService);
            this.topicFilter = dataIdentifierReference.id();
            this.primary = primary;
            this.storeDataPoints = storeDataPoints;
        }

        @Override
        public void process(final @NotNull PUBLISH publish) {
            if (storeDataPoints) {
                topicFilterToPublish.put(topicFilter, publish);
            }
            if (primary) {
                triggerPublish(combining);
            }
        }
    }

    public final class InternalTagConsumer implements InternalConsumer, SingleTagConsumer {
        private final @NotNull DataIdentifierReference dataIdentifierReference;
```

caoccao: OK, I see it now. The hierarchy looks a bit odd. Maybe we can further refactor it later.

caoccao (approved):

hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java
mschoenert-hivemq: If you want, feel free to propose an API that you feel is "less odd". ;-)
Please do that in the Linear ticket.

hivemq-edge/src/main/java/com/hivemq/mqtt/services/PublishDistributorImpl.java
```diff
@@ -59,6 +59,12 @@
 @Singleton
 public class PublishDistributorImpl implements PublishDistributor {
 
+    /**
+     * Client ID prefix shared by all {@link com.hivemq.persistence.clientqueue.InternalTopicFilterSubscriber}
+     * instances. Used in {@link #handlePublish} to bypass the session check for internal subscribers.
+     */
+    public static final @NotNull String INTERNAL_SUBSCRIBER_PREFIX = "internal#";
```
codepitbull: I know it's pedantic but we should add further protection of this.
While it is very unlikely but there might be clients using that id.
Either by accident or to mess with us.
I would suggest to add a check where the clientId of an actual mqtt client is checked for this magic string and the client is disconnected with a proper reason code.

mschoenert-hivemq: Great point. Then I would clean it up a bit more. WDYT?

**Current Subscribers**

| Component | Subscriber ID format | Shared name |
|---|---|---|
| MQTT clients | raw client ID | `null` (or share group from `$share/group/topic`) |
| Bridge forwarder | `forwarder#{forwarderId}#{hivemqId}` | `forwarder#{forwarderId}` |
| Sampling service | `Sampler#{topicFilter}` | `Sampler#{topicFilter}` |
| Data combining | `{uuid}#` | `{uuid}` |
| InternalTopicFilterSubscriber | `internal#{component}#{instance}` | `null` |
| Startup recovery | client ID from persisted sessions | restored from persistence |

There is no namespace enforcement — it is entirely convention. An MQTT client registering with ID `forwarder#x#y` or `internal#combiner#foo` would silently collide with an internal component, and the `startsWith(INTERNAL_SUBSCRIBER_PREFIX)` check in `PublishDistributorImpl` would misroute publishes to it.

**Proposed Namespacing**

Use `$`-prefixed namespaces to make the boundaries explicit and machine-checkable:

| Component | New subscriber ID format |
|---|---|
| MQTT clients | raw client ID (unchanged) |
| Bridge forwarder | `$FORWARDER::{forwarderId}` |
| Sampling service | `$SAMPLER::{topicFilter}` |
| Data combining | `$COMBINER::{uuid}` |
| InternalTopicFilterSubscriber | `$INTERNAL::{component}::{instance}` |

**Feasibility**

The subscriber population splits into two fundamentally different cases.

**Internal components (forwarder, sampler, combiner, InternalTopicFilterSubscriber)** are not real MQTT clients. Their IDs are fabricated strings, never from the wire. They have no MQTT sessions, no session persistence records, no connection persistence entries. The only places their IDs appear are:

- their own class (1–2 `addTopic`/`removeSubscriber` call sites each)
- `ClientQueuePersistenceImpl` — 3 `startsWith` routing checks
- `PublishDistributorImpl` — 2 `startsWith` routing checks

Renaming all internal components is a small, contained refactor: update ID construction in each component, update the 5 `startsWith` checks. No data migration, no protocol impact.

**MQTT clients** are a different story. The client ID comes from the CONNECT packet — user-supplied, immutable during the session, and used as the primary key across session persistence, queue persistence, connection persistence, and subscription persistence. Introducing a prefix on the topic tree side while keeping the raw ID everywhere else would require a decoupling layer and a data migration across all four persistence stores. Not worth it.

**Protection for MQTT Clients**

The right fix for the MQTT client collision risk is **CONNECT-time validation**: reject any CONNECT packet whose client ID starts with a reserved prefix (`$FORWARDER::`, `$SAMPLER::`, `$COMBINER::`, `$INTERNAL::`). One addition, one place, collision risk eliminated.

The `$` prefix convention signals "system-reserved" — already established in MQTT for `$SYS` and `$share`. Legitimate MQTT clients almost never use it, and with explicit validation they cannot.

**Change Summary**

| Change | Effort |
|---|---|
| Rename forwarder ID to `$FORWARDER::{id}` | ~5 lines |
| Rename sampler ID to `$SAMPLER::{topic}` | ~5 lines |
| Rename combiner ID to `$COMBINER::{uuid}` | ~5 lines |
| Rename internal subscriber prefix to `$INTERNAL::{component}::{instance}` | ~2 lines |
| Update 5 `startsWith` routing checks in 2 files | small |
| Add CONNECT-time validation rejecting reserved prefixes | ~10 lines |
| MQTT clients | no change |

codepitbull: Looks fine to me!
But this will require a big fat warning for the next release.

mschoenert-hivemq: Hmm. Might be a bit difficult to formulate that warning:
* WARNING: Until now you could use MQTT clientIds that would collide with internal subscriptions and would mess up the bridge, your client subscription, or probably both - now we will reject such a client.
;-)

mschoenert-hivemq: DONE

caoccao (approved):

codepitbull (approved):

PR #1508, opened 2026-04-22
Feature/edg 501/simulatio improvements

codepitbull:
**Motivation**

Resolves #EDG-501

**Changes**
- Add static and random tags
- Add southbound capability
- Add schema generation
- Remove usage of deprecated APIs

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/simulation/SimulationProtocolAdapter.java
+++ b/hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/simulation/SimulationProtocolAdapter.java
@@ -17,7 +17,10 @@ package com.hivemq.edge.modules.adapters.simulation;
 
 import static com.hivemq.adapter.sdk.api.state.ProtocolAdapterState.ConnectionStatus.STATELESS;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.hivemq.adapter.sdk.api.ProtocolAdapterInformation;
+import com.hivemq.adapter.sdk.api.datapoint.DataPointBuilder;
+import com.hivemq.adapter.sdk.api.datapoint.DataPointListBuilder;
 import com.hivemq.adapter.sdk.api.model.ProtocolAdapterInput;
 import com.hivemq.adapter.sdk.api.model.ProtocolAdapterStartInput;
 import com.hivemq.adapter.sdk.api.model.ProtocolAdapterStartOutput;
@@ -32,6 +35,10 @@ import com.hivemq.adapter.sdk.api.schema.SchemaBuilder;
 import com.hivemq.adapter.sdk.api.schema.TagSchemaCreationInput;
 import com.hivemq.adapter.sdk.api.schema.TagSchemaCreationOutput;
 import com.hivemq.adapter.sdk.api.state.ProtocolAdapterState;
+import com.hivemq.adapter.sdk.api.writing.WritingInput;
+import com.hivemq.adapter.sdk.api.writing.WritingOutput;
+import com.hivemq.adapter.sdk.api.writing.WritingPayload;
+import com.hivemq.adapter.sdk.api.writing.WritingProtocolAdapter;
 import com.hivemq.edge.modules.adapters.simulation.config.SimulationSpecificAdapterConfig;
 import com.hivemq.edge.modules.adapters.simulation.tag.RandomValueConfig;
 import com.hivemq.edge.modules.adapters.simulation.tag.SimulationTag;
@@ -41,10 +48,11 @@ import com.hivemq.edge.modules.adapters.simulation.tag.StaticValueConfig;
 import java.util.List;
 import java.util.Objects;
 import java.util.Random;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ThreadLocalRandom;
 import org.jetbrains.annotations.NotNull;
 
-public class SimulationProtocolAdapter implements BatchPollingProtocolAdapter {
+public class SimulationProtocolAdapter implements BatchPollingProtocolAdapter, WritingProtocolAdapter {
 
     private final @NotNull ProtocolAdapterInformation adapterInformation;
     private final @NotNull SimulationSpecificAdapterConfig adapterConfig;
@@ -53,6 +61,13 @@ public class SimulationProtocolAdapter implements BatchPollingProtocolAdapter {
     private static final @NotNull Random RANDOM = new Random();
     private final @NotNull String adapterId;
     private final @NotNull List<SimulationTag> tags;
+    /**
+     * Holds the live value for every STATIC_VALUE tag. Seeded from each tag's configured
+     * {@link StaticValueConfig#getParsedValue()} at {@link #start(ProtocolAdapterStartInput, ProtocolAdapterStartOutput)}
+     * and updated via {@link #write(WritingInput, WritingOutput)}. The configured value on the
+     * tag itself is intentionally never mutated.
+     */
+    private final @NotNull ConcurrentHashMap<String, Object> currentStaticValues = new ConcurrentHashMap<>();
 
     public SimulationProtocolAdapter(
             final @NotNull ProtocolAdapterInformation adapterInformation,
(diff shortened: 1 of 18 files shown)


hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/simulation/SimulationProtocolAdapter.java
```diff
@@ -138,6 +278,68 @@ public int getMaxPollingErrorsBeforeRemoval() {
...
+                        .readable();
+            case LONG ->
+                builder.scalar(ScalarType.LONG)
+                        .minimum((long) min)
+                        .maximum((long) max)
+                        .readable();
+            case DOUBLE ->
+                builder.scalar(ScalarType.DOUBLE).minimum(min).maximum(max).readable();
+            case STRING -> throw new IllegalStateException("RANDOM_NUMBER cannot have STRING valueType");
+        }
```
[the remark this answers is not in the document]
codepitbull: Fixed.

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/simulation/tag/RandomValueConfig.java
```diff
@@ -0,0 +1,100 @@
...
+            enumDisplayValues = {"Integer (32-bit)", "Long (64-bit)", "Double"},
+            required = true)
+    private final @NotNull SimulationValueType valueType;
+
+    @JsonProperty(value = "minValue", required = true)
+    @ModuleConfigField(title = "Minimum Value", description = "Inclusive lower bound.", required = true)
+    private final double minValue;
+
+    @JsonProperty(value = "maxValue", required = true)
+    @ModuleConfigField(title = "Maximum Value", description = "Upper bound.", required = true)
```
marregui: is it inclusive or exclussive?

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/simulation/tag/RandomValueConfig.java
```diff
@@ -0,0 +1,100 @@
...
+        if (valueType == SimulationValueType.STRING) {
+            throw new IllegalArgumentException("randomValue.valueType must be INT, LONG or DOUBLE");
+        }
+        if (minValue >= maxValue) {
+            throw new IllegalArgumentException(
+                    "randomValue requires minValue < maxValue, was: " + minValue + " / " + maxValue);
+        }
+
+        if (valueType == SimulationValueType.INT) {
+            if (minValue < Integer.MIN_VALUE || maxValue > (long) Integer.MAX_VALUE + 1) {
```
marregui: as per here, Integer.MAX_VALUE + 1 is a valid value for upper bound

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/simulation/SimulationProtocolAdapter.java
```diff
@@ -86,43 +114,142 @@ public void stop(final @NotNull ProtocolAdapterStopInput input, final @NotNull P
...
+                b.value(ThreadLocalRandom.current()
+                        .nextDouble(
+                                Math.min(adapterConfig.getMinValue(), adapterConfig.getMaxValue()),
+                                Math.max(adapterConfig.getMinValue() + 1, adapterConfig.getMaxValue())));
+            case RANDOM_NUMBER -> {
+                final RandomValueConfig rv = Objects.requireNonNull(def.getRandomValue());
+                switch (rv.getValueType()) {
+                    case INT ->
+                        b.value(ThreadLocalRandom.current()
+                                .nextInt((int) rv.getMinValue(), (int) rv.getMaxValue() + 1));
```
marregui, line 165: as Integer.MAX_VALUE + 1  is a valid maxValue as per `RandomValueConfig`, this will overflow all over

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterWrapper.java
```diff
@@ -241,7 +241,15 @@ private void cleanUpScheduler() {
...
                 protocolAdapterState.setConnectionStatusListener(status -> {
-                    if (firstCallToStatusListener.compareAndSet(true, false) && status == initialConnectionStatus) {
+                    // setConnectionStatusListener fires the listener once with the current status.
+                    // We want to skip that synthetic "initial" callback only when the adapter has
+                    // not yet reached a ready state — otherwise we'd wait forever for a transition
+                    // that never comes (e.g. STATELESS adapters that are ready the moment they
+                    // are constructed).
+                    if (firstCallToStatusListener.compareAndSet(true, false)
+                            && status == initialConnectionStatus
+                            && status != ProtocolAdapterState.ConnectionStatus.CONNECTED
```
marregui, line 256: in looking deeper into this `attemptStartingConsumers`, in line 262 we call startWritingAsync from ProtocolAdapterWritingServiceImpl . My worry is that startWritingAsync is not idempotent. If the sequence:

```
        // Boot the adapter. No listener is wired yet at this point.
        wrapper.startAsync(true).get(10, TimeUnit.SECONDS);

        // The wrapper installed its listener inside attemptStartingConsumers. Adapter is currently
        // DISCONNECTED (default), so the synthetic replay is skipped (pre-fix and post-fix behavior
        // match here: initial = DISCONNECTED, neither CONNECTED nor STATELESS).
        //
        // --- CONNECTED #1: first genuine transition, triggers startWritingAsync call #1 ---
        adapterState.setConnectionStatus(ProtocolAdapterState.ConnectionStatus.CONNECTED);

        // --- DISCONNECTED: peer drops, listener fires but startWritingAsync is NOT called ---
        adapterState.setConnectionStatus(ProtocolAdapterState.ConnectionStatus.DISCONNECTED);

        // --- CONNECTED #2: reconnect, triggers startWritingAsync call #2 (the bug) ---
        adapterState.setConnectionStatus(ProtocolAdapterState.ConnectionStatus.CONNECTED);
``` 

we will end up with leaked subscriptions

marregui (changes requested):

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/simulation/tag/RandomValueConfig.java
codepitbull: fixed

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/simulation/tag/RandomValueConfig.java
codepitbull: fixed

hivemq-edge/src/main/java/com/hivemq/edge/modules/adapters/simulation/SimulationProtocolAdapter.java
codepitbull: fixed

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterWrapper.java
codepitbull: fixed.

marregui: `writingStarted.compareAndSet(false, true)` flips before `startWritingAsync().get()` runs. If `.get()` throws or the data hub returns false, the flag stays `true` and no future CONNECTED will retry. That's a deliberate-looking trade-off (retrying without first tearing down a partial install would re-leak the very resources this guards), but it isn't stated anywhere. Suggest a one-liner so the next person doesn't "fix" it:

  ```java
  // CAS before startWritingAsync so a concurrent reconnect cannot re-enter.
  // If startWritingAsync fails, writingStarted stays true on purpose — retrying
  // without first tearing down the partial install would re-leak the resources
  // this guard protects. Recovery is stop+start the adapter.
  if (!writingStarted.compareAndSet(false, true)) {
      return;
  }

  If we ever do want auto-retry on transient writing-start failures, it needs (a) stopWriting proven safe against a partial startWritingAsync, and (b) a separate signal to keep the DISCONNECTED early-return honest — out of scope here.

marregui (approved):
LGTM

PR #1509, opened 2026-04-23
Switch to Java 25

codepitbull:
**Motivation**

Resolves #EDG-516

**Changes**
Update build to produce Java 25 bytecode.


caoccao (approved):
LGTM

.github/workflows/etherip-package.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5
         with:
           distribution: 'adopt'
-          java-version: '21'
+          java-version: '25'
```
mschoenert-hivemq, line 24: What is this file? Why is there a "publish EtherIp" workflow in the Edge Repo?

codepitbull: The library is not published to maven central.
We have to build and publish it internally to use it.

PR #1514, opened 2026-04-27
feat(EDG-546): SNMP adapter initial version

Owen-Compton:
Closes EDG-546

## Summary

- New SNMP protocol adapter module under `modules/hivemq-edge-module-snmp/`
- Migrated off deprecated `DataPointFactory` API to the new `DataPointListBuilder` chain
- JDK 25: `StructuredTaskScope` (JEP 505) for concurrent OID fan-out, switch pattern matching for typed value dispatch, virtual-thread-friendly blocking I/O
- SNMP4J 3.7.8 shaded under `com.hivemq.edge.adapters.snmp.shaded.snmp4j` to avoid classpath conflicts
- Supports SNMPv1, v2c, and v3 (auth/priv)
- OID browser via `discoverValues()` with MIB-II folder structure

## Test coverage

- `SnmpProtocolAdapterInformationTest` — adapter metadata contracts
- `SnmpAdapterConfigTest` — JSON deserialization and round-trip for minimal, full V2c, and full V3 configs
- `SnmpProtocolAdapterTest` — lifecycle (start/stop/double-start guard), poll guards, typed value dispatch, partial failure handling
- `SnmpAdapterSmokeTest` — embedded SNMP4J agent exercises real network I/O end-to-end

## Test plan

- [ ] `./gradlew :hivemq-edge-module-snmp:test`
- [ ] Connect to a real SNMP device and verify polling publishes data points to MQTT
- [ ] Test SNMPv3 auth/priv with a compatible agent
- [ ] Verify OID browser in the Edge UI returns the MIB-II folder structure

🤖 Generated with [Claude Code](https://claude.com/claude-code)


modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/config/SnmpSpecificAdapterConfig.java
```diff
@@ -0,0 +1,245 @@
...
+public class SnmpSpecificAdapterConfig implements ProtocolSpecificAdapterConfig {
+
+    public static final int PORT_MIN = 1;
+    public static final int PORT_MAX = 65535;
+    public static final int DEFAULT_PORT = 161;
+    public static final int DEFAULT_TIMEOUT_MILLIS = 3000;
+    public static final int DEFAULT_RETRIES = 1;
+    public static final String DEFAULT_COMMUNITY = "public";
+
+    private static final @NotNull String ID_REGEX = "^([a-zA-Z_0-9-_])*$";
```
caoccao: Bug: `-_` creates a range Inside a character class, - defines a range unless it’s first/last or escaped.
Here the tail -_ is interpreted as a range from - (ASCII 45) to _ (ASCII 95), which includes a lot more characters than intended:

It will allow these extra characters:
`./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`

So it inadvertently allows things like ., /, :, @, [ ], \, ^, etc.

If you meant “hyphen or underscore”, write one of:

`^[a-zA-Z0-9_-]*$` (put - at the end)
`^[a-zA-Z0-9_\\-]*$` (escape the hyphen)
`^[\\w-]*$` (note: `\w` also includes letters/digits/underscore, but not locale-specific unless you use flags)
Probably not desired: empty string is valid Because of *, "" matches. If you want at least 1 character, use +:

`^[a-zA-Z0-9_-]+$`
Redundancy / style

The parentheses are unnecessary: `^([…])*$` can be `^[…]*$`.
`a-zA-Z_0-9` can be reordered more clearly as `a-zA-Z0-9_`.

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/SnmpProtocolAdapter.java
```diff
@@ -0,0 +1,353 @@
...
+            return;
+        }
+
+        if (!started.compareAndSet(false, true)) {
+            output.failStart(new IllegalStateException("Already started"), "Adapter already started");
+            return;
+        }
+
+        try {
+            client = createClient();
```
caoccao: If `testConnection()` returns `false`, `client` should be closed and set to `null`. Otherwise, there will be resource leak.

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/config/SnmpToMqttConfig.java
```diff
@@ -0,0 +1,92 @@
...
+            numberMax = 100,
+            defaultValue = "10")
+    private final int maxPollingErrorsBeforeRemoval;
+
+    @JsonProperty(value = "publishChangedDataOnly")
+    @ModuleConfigField(
+            title = "Publish Changed Data Only",
+            description = "Only publish data points when their values have changed",
+            defaultValue = "false")
+    private final boolean publishChangedDataOnly;
```
caoccao, line 52: `publishChangedDataOnly` is not implemented at all.

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/SnmpProtocolAdapter.java
```diff
@@ -0,0 +1,353 @@
...
+                    .toList();
+        } // close() blocks until all submitted tasks complete
+
+        final var publisher = pollingOutput.dataPointListPublisher();
+        for (int i = 0; i < tags.size(); i++) {
+            final SnmpTag tag = tags.get(i);
+            try {
+                publishTagResult(publisher, futures.get(i).get());
+            } catch (final ExecutionException e) {
+                final Throwable cause = e.getCause();
```
caoccao, line 207: Adapter marks connection as CONNECTED even when tag reads fail

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/config/tag/SnmpTagDefinition.java
```diff
@@ -0,0 +1,83 @@
...
+            required = true,
+            stringPattern = OID_REGEX)
+    private final @NotNull String oid;
+
+    @JsonProperty(value = "dataType")
+    @ModuleConfigField(
+            title = "Data Type",
+            description = "Expected SNMP data type. Use AUTO for automatic detection.",
+            defaultValue = "AUTO")
+    private final @NotNull SnmpDataType dataType;
```
caoccao, line 49: `dataType` hint is exposed but ignored in value conversion

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/SnmpProtocolAdapter.java
```diff
@@ -0,0 +1,353 @@
...
+
+    private static final @NotNull Logger log = LoggerFactory.getLogger(SnmpProtocolAdapter.class);
+
+    private final @NotNull String adapterId;
+    private final @NotNull ProtocolAdapterInformation adapterInformation;
+    private final @NotNull SnmpSpecificAdapterConfig config;
+    private final @NotNull ProtocolAdapterState state;
+    private final @NotNull List<SnmpTag> tags;
+    private final @NotNull AtomicBoolean started = new AtomicBoolean(false);
+    private final @NotNull AtomicBoolean stopRequested = new AtomicBoolean(false);
```
caoccao, line 74: # stop() resets stopRequested to false in the finally block.

This defeats the CAS-based "already stopping" guard. Once stop() returns, a concurrent in-flight poll() that re-checks stopRequested.get() will see false. More importantly, stopRequested is supposed to be a one-way latch ("this adapter has been asked to stop") but is being used as a re-entrancy guard that also acts like a latch — it can't be both. Pick one semantics. If it's a latch, drop the reset; if it's just re-entrancy, rename it.

# Race between start() and stop() can leak the client and leave the adapter in a half-broken state

Trace: thread A enters start() past the stopRequested.get() check; thread B enters stop(), finds client == null (A hasn't assigned yet), sets started=false, returns; A then assigns client, calls testConnection() successfully, calls output.startedSuccessfully() — but started is now false, so subsequent polls fail and the just-created client is leaked. The started / stopRequested / client triple isn't atomic; either guard with a single lock around the start/stop sections or merge into a single state machine reference.

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/SnmpProtocolAdapter.java
```diff
@@ -0,0 +1,353 @@
...
+            final @NotNull ProtocolAdapterDiscoveryInput input, final @NotNull ProtocolAdapterDiscoveryOutput output) {
+        final NodeTree tree = output.getNodeTree();
+
+        if (input.getRootNode() == null) {
+            tree.addNode(
+                    "mib2-system",
+                    "System",
+                    "1.3.6.1.2.1.1",
+                    "System MIB - sysDescr, sysObjectID, sysUpTime, sysContact, sysName, sysLocation, sysServices",
+                    null,
```
caoccao, line 302: Child nodes aren't linked to their parent folder.

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/SnmpClient.java
```diff
@@ -0,0 +1,278 @@
...
+        // Add default security protocols
+        SecurityProtocols.getInstance().addDefaultProtocols();
+
+        // Create USM with local engine ID
+        OctetString localEngineID = new OctetString(MPv3.createLocalEngineID());
+        USM usm = new USM(SecurityProtocols.getInstance(), localEngineID, 0);
+        SecurityModels.getInstance().addSecurityModel(usm);
+
+        // Add user if security name is provided
+        if (config.getSecurityName() != null) {
```
caoccao: `SnmpClient.configureSnmpV3()` skips USM user creation if securityName == null, but createTarget() still builds a UserTarget and proceeds. The first GET will then fail with an obscure SNMP4J error. Validate at config load that V3 => securityName != null and reject early with a clear message.

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/SnmpProtocolAdapter.java
```diff
@@ -0,0 +1,353 @@
...
+        // Fan out all SNMP GETs on virtual threads, then collect results sequentially.
+        final List<Future<TagReadResult>> futures;
+        try (final ExecutorService executor = Executors.newVirtualThreadPerTaskExecutor()) {
+            futures = tags.stream()
+                    .map(tag -> executor.submit(() -> readTag(tag)))
+                    .toList();
+        } // close() blocks until all submitted tasks complete
+
+        final var publisher = pollingOutput.dataPointListPublisher();
+        for (int i = 0; i < tags.size(); i++) {
```
caoccao: Better use `futures` instead of `tags`.

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/SnmpProtocolAdapter.java
```diff
@@ -0,0 +1,353 @@
...
+        state.setConnectionStatus(CONNECTED);
+        publisher.publish();
+    }
+
+    protected @NotNull SnmpClient createClient() throws IOException {
+        return new SnmpClient(config);
+    }
+
+    private @NotNull TagReadResult readTag(final @NotNull SnmpTag tag) throws IOException {
+        final SnmpClient snmpClient = client;
```
caoccao, line 229: There are other code paths that don't apply this defensive copy.

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/config/tag/SnmpTag.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @JsonProperty(value = "description")
+    @ModuleConfigField(title = "Description", description = "A human-readable description of the tag")
+    private final @NotNull String description;
+
+    @JsonProperty(value = "definition", required = true)
+    @ModuleConfigField(title = "Definition", description = "The OID definition for this tag")
+    private final @NotNull SnmpTagDefinition definition;
+
+    public SnmpTag(
```
caoccao, line 48: `@JsonCreator` is missing.

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/SnmpProtocolAdapterInformation.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public @Nullable ProtocolAdapterCategory getCategory() {
+        return ProtocolAdapterCategory.CONNECTIVITY;
+    }
+
+    @Override
+    public List<ProtocolAdapterTag> getTags() {
+        return List.of(ProtocolAdapterTag.UDP, ProtocolAdapterTag.TCP);
```
caoccao: SNMP is UDP only.

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/SnmpProtocolAdapter.java
```diff
@@ -0,0 +1,353 @@
...
+        log.trace(
+                "Read OID {} = {} for tag {}",
+                tagResult.oid(),
+                tagResult.result().getValue(),
+                tagResult.tag().getName());
+    }
+
+    private void setTypedValue(
+            final @NotNull DataPointBuilder<DataPointListBuilder> builder, final @Nullable Object value) {
+        switch (value) {
```
caoccao, line 281: Does this cover all types?

caoccao (changes requested):
Architectural / API

  - Remove the static SnmpClient UDP socket from the constructor: today the Snmp instance is created and listens immediately. If the adapter is constructed but never started, the socket leaks. Move transport.listen() to a separate open() method called from
  start().
  - Inject the SnmpClient factory instead of using a protected createClient() hook for testing. Cleaner DI than subclassing in tests.
  - Make SnmpReadResult a record — it's a value carrier with no behaviour. Five fewer lines of boilerplate.
  - Consider GETBULK for v2c/v3 multi-OID polls instead of one request per tag. For 50+ tags, the current approach saturates the agent with 50 PDUs per poll cycle.
  - Consider returning sized integers from convertVariable: Counter32/Gauge32 always fit in long, but consumers reading metaBuilder.put("rawType", "Counter32") and seeing a Long value may want a hint that it's 32-bit.

  Configuration

  - Validate combinations: authProtocol != NONE ⇒ authPassword != null, same for priv. Surface as ProtocolSpecificAdapterConfig.validate() (or constructor-time IllegalArgumentException).
  - Require V3 ⇒ securityName (see bug #8).
  - Reject community for V3 and auth*/priv*/securityName for V1/V2c, otherwise the UI lets the user fill fields that are silently ignored.
  - Add engineId config for SNMPv3 — some agents require an explicitly-configured engine ID instead of MPv3.createLocalEngineID().

  Testing

  - SnmpAdapterSmokeTest runs a real UDP listener on every test class load. Consider tagging it @Tag("integration") and excluding from the default suite.
  - The smoke test mocks the config instead of constructing a real one — that's a smell. The constructor takes 12 args but a small builder or a public V2C-only convenience constructor would make tests cleaner.
  - No tests for: V3 codepath, walk/discovery, timeout behaviour, malformed OID at runtime, all-tags-fail scenarios. Add table-driven tests for convertVariable covering every SNMP4J Variable subclass.

  Observability

  - The SnmpReadResult.errorStatus and errorIndex are added to metadata even though they're always 0 for successful reads (the unsuccessful path throws before constructing SnmpReadResult). Drop them, or actually capture them on partial successes.
  - Polling logs at TRACE for every successful read; consider DEBUG-level summary at end of poll instead.
  - Discovery log says "Discovery walk from {} found {} OIDs" but doesn't say from which adapter — include adapterId.

Owen-Compton (commented):
All of Sam's comments attended to.

modules/hivemq-edge-module-snmp/src/main/java/com/hivemq/edge/adapters/snmp/SnmpProtocolAdapter.java
```diff
@@ -0,0 +1,420 @@
...
+
+    private final @NotNull String adapterId;
+    private final @NotNull ProtocolAdapterInformation adapterInformation;
+    private final @NotNull SnmpSpecificAdapterConfig config;
+    private final @NotNull ProtocolAdapterState state;
+    private final @NotNull List<SnmpTag> tags;
+    private final @NotNull SnmpClientFactory clientFactory;
+    private final @NotNull AtomicBoolean started = new AtomicBoolean(false);
+    // One-way latch: once stop() is called this adapter instance is permanently retired.
+    private final @NotNull AtomicBoolean stopRequested = new AtomicBoolean(false);
```
caoccao, line 74: - Race between `start()` and `stop()`: a re-check of `stopRequested` was added after the client is constructed (file:115-123) so the obvious window is closed. Remaining concerns:
  - `client` is a plain (non-volatile) reference. The `start` thread's assignment (file:113) and the `stop` thread's read (file:158) have no happens-before guarantee. Mark it `volatile` or guard `start`/`stop` with a single lock.
  - `started` is still flipped independently of `stopRequested` and `client`. The trio is not atomic; it is possible for `start` to call `output.startedSuccessfully()` (file:127) and immediately afterwards have `stop` set `started=false` and close the client. The next `poll` then trips the early-out at file:174 — functionally safe, but the lifecycle still violates the "either started or stopped" invariant the reviewer asked for.
  - **Recommendation**: replace the three flags + nullable client with a single `AtomicReference<State>` (e.g. `INIT/STARTING/RUNNING/STOPPING/STOPPED`) or wrap `start`/`stop` in a private lock. Otherwise document the race as accepted.

caoccao (approved):
The race conditions are not fully removed. But, that shouldn't be a block issue.

I recommend we keep it open till FSM is merged, otherwise, the FSM ticket will have to be reworked.

PR #1515, opened 2026-04-28
PLT-405: Rely on GitHub webhook to trigger composite builds

ymengesha:
## Summary
- Remove `build job` call from `finally` block in `CompositeJenkinsfile` — the GitHub push webhook now handles triggering the composite build
- Delete redundant `CompositeJenkinsfile` from the `hivemq-edge-module-opcua` module

## Test plan
- [ ] Verify that pushing a branch triggers the composite build via the GitHub webhook
- [ ] Confirm the pipeline only creates the remote branch without explicitly triggering a job


caoccao (approved):
LGTM

PR #1516, opened 2026-04-28
Feature/edg 265/new schema endpoint

codepitbull:
**Motivation**

Resolves #EDG-265

**Changes**
getWritiingSchema was a very misleading name and the endpoint is moved to getSchema


hivemq-edge-frontend/cypress/support/__generated__/apiRoutes.ts
```diff
@@ -190,7 +192,7 @@ export const API_ROUTES = {
     /**
      * @see {@link AuthenticationService.validateToken}
      */
-    validateToken: route<void>('POST', '/api/v1/auth/validate-token'),
+    validateToken: route<unknown>('POST', '/api/v1/auth/validate-token'),
```
caoccao, line 195: This change is weird. We generally don't recommend using `unknown`. Any reason for that?

caoccao (approved):
LGTM, just one question

hivemq-edge-frontend/cypress/support/__generated__/apiRoutes.ts
codepitbull: Hmm, that's a weird one.
The code is generated and there are no changes in the openapispec affecting this path.
Maybe related to a lib update?

caoccao: Better let CC analyze that and find the root cause.

PR #1518, opened 2026-04-29
Bring back the thread in simulation adapter because without it the PollingService dies

codepitbull:
**Motivation**

Fix for resource leak revealed aother issue in the PollingService.
Bringing back the old behavior.


estefania-hivemq (approved):
LGTM

caoccao (approved):
LGTM

PR #1524, opened 2026-05-07
EDG-585: presets/axes model for OPC UA certificate validation

marregui:
## The problem

A factory floor usually has no certificate authority: every machine presents a certificate it issued to itself. HiveMQ Edge could not talk to those servers at all, and the setting that looks like it should help — `tlsChecks=NONE` — does not mean what its name suggests. It only switches off the *optional* checks; the certificate must still chain to a trusted authority, which a self-signed certificate never does. That is the failure Miele reported.

## What this changes

Certificate validation is configured through exactly one of two mutually exclusive settings:

- **`tlsChecks`** — a named preset for the common cases. Two new values: `SELF_SIGNED`, which trusts specific certificates by their SHA-256 fingerprint listed in a file you maintain, and `NO_VERIFICATION`, which accepts anything and says so loudly in the log.
- **`tlsChecksFull`** — the six individual checks (trust mode, ApplicationUri, hostname, validity, revocation, key usage), for anything the presets do not cover. Every one is optional and defaults to its strictest value, so a forgotten entry gives you *more* checking, never less.

Setting both is a configuration error. Setting neither behaves exactly as before.

**Nothing changes on upgrade.** All four legacy preset values were verified bit-for-bit against the check sets they selected previously. The settings are read and never rewritten, so a configuration file keeps exactly what its author wrote.

`SELF_SIGNED` is the answer for the factory case, and it is a real trust decision rather than an absence of one: an unexpected certificate is refused, and the log names the fingerprint that was presented so an operator can add it deliberately.

## Generic adapter-core changes — reviewed as owner

Some of this is core code that affects **every adapter type**, not just OPC UA. Martin assigned ownership of `ProtocolAdapterManager` and `ProtocolAdapterWrapper` to me for this PR, and the decision was taken to keep the work here rather than split it. The full record is in the Linear ticket; the substance:

- **Per-adapter conversion isolation.** An adapter whose configuration cannot be read no longer stops every other adapter in the deployment from being reconfigured. Adapter configurations were converted in a single pass, so one unreadable element anywhere aborted the whole reload with a single stack trace and no event.
- **Failure containment is now total and uniform.** All four refresh loops catch `Throwable` and rethrow only `VirtualMachineError` — the four errors that mean the JVM itself is broken. A third-party adapter module whose classloader is broken or whose static initialiser throws is that one adapter's problem, whichever type it throws.
- **At-most-once destroy, and no start after destroy.** `start()` refuses on a destroyed wrapper, and the idle short-circuit destroys only an adapter that actually reached `Working` — an adapter that never started acquired nothing to release, and the SDK does not promise `destroy()` tolerates being called without a preceding `start()`.
- **Log volume is bounded.** A configuration nobody corrects gets a full stack trace on its first failure and a single line on every reload after that.
- **The entity/config ID invariant is pinned by a test** rather than asserted, so the duplicate-ID arithmetic cannot silently develop a hole.

Two consequences were accepted deliberately rather than fixed here, and each has a ticket: an adapter running a configuration that no longer matches the file reports no visible state ([EDG-892](https://linear.app/hivemq/issue/EDG-892/adapter-running-a-stale-configuration-looks-healthy-in-the-ui)), and unknown configuration settings are warned-and-dropped application-wide, which is right for forward compatibility but means security-relevant settings must be refused by each module as OPC UA does ([EDG-893](https://linear.app/hivemq/issue/EDG-893/audit-adapters-for-security-relevant-settings-that-are-silently-warn)).

## Also in this PR

- **A configuration Edge cannot read is refused, not silently defaulted.** The refusal names the element and the mistake. For `tls`, `keystore` and `truststore` it deliberately does *not* repeat the text it found: an XML collapse concatenates everything nested inside the element, so that text is typically a store password — and for `tls`, both stores' passwords at once. This exception is logged with a stack trace, so echoing it would copy credentials into logs and support bundles. `tlsChecksFull` still quotes its text, because those elements hold only validation settings.
- A fingerprint file configured where it will never be used now produces a warning instead of being silently ignored.
- A failed adapter start no longer leaves scheduler threads behind.
- **The frontend TLS form test now renders a generated artifact.** `TlsSchemaArtifactTest` writes the RJSF schema pair from `CustomConfigSchemaGenerator` and the real `opcua-adapter-ui-schema.json`, and fails when the committed copy is stale — the same staleness gate as the committed OpenAPI spec. It replaces a 138-line hand-maintained transcription that had already drifted from what Edge serves.

## Also in this PR: a Monaco readiness fix in the DataHub editor

Three files here are not OPC UA: `src/extensions/datahub/components/forms/CodeEditor.tsx`, its component spec, and the shared Cypress page object `cypress/pages/Monaco/MonacoEditor.ts`. Flagging them so they are reviewed as what they are.

**The defect.** `@monaco-editor/react` creates the editor and subscribes to `onDidChangeModelContent` from two effects of the same commit, in that order. A write landing between `create()` and that subscription is applied to the model and never reaches `onChange` — the editor shows the new text while the form holds the old one. Nothing observable distinguishes that window: the editor is in the DOM, visible, and already returned by `monaco.editor.getEditors()`.

**The change.** `CodeEditor` publishes `data-editor-ready` from its `onMount` callback, whose effect precedes the subscription. The shared page object waits on that attribute instead of the previous workaround — a no-op `editor.trigger('keyboard', 'type', { text: '' })` after `setValue()`, which masked the race by forcing a second change event rather than waiting for the first to be heard.

**Runtime impact: none.** The attribute is rendered as `isEditorMounted || undefined`, so it is absent rather than `"false"`. No user-visible behaviour changes — a person cannot type into an editor that has not mounted. The affected editor is DataHub's, which the OPC UA TLS form does not use.

**Why it is in this PR:** it should have been its own PR. It was found and fixed while working on this branch, and it is called out here rather than folded in quietly. Happy to split it out if you would rather review it separately.

## One more unrelated file: a Workspace E2E test

`cypress/e2e/workspace/wizard/wizard-create-group.spec.cy.ts` re-fits the canvas before clicks made after the wizard's selection panel has grown. The panel is a top-right React Flow `Panel` with `pointerEvents: 'all'` that gains a list entry per selected node, so it can cover the very node the test then clicks.

**This defect is master's, not this branch's.** Reproduced against the instrumented build that CI actually runs (`VITE_COVERAGE=true` build + `vite preview`, not the dev server): master fails the identical two tests. Master passes in CI only on timing margin, which this branch's slightly larger bundle spends. Merging this fixes master too.

The underlying product issue is untouched and deserves its own ticket: a user cannot click a node the panel covers, and the panel's own comment claims it does not block canvas interaction.

## Verification

723 OPC UA module tests and 4796 core tests green locally, with spotless, ErrorProne and NullAway clean. Each fix was checked by disabling it and confirming the intended tests fail — including the credential-redaction assertions, which were proved capable of failing by temporarily re-introducing the interpolation. Verified end to end against a real Prosys OPC UA simulation server: the reported failure reproduces exactly, and `SELF_SIGNED` with the fingerprint from the log connects and streams data.

Companion PRs: hivemq/hivemq-edge-test#345, hivemq/hivemq4-documentation#2675. Design and full owner decision record: [EDG-585](https://linear.app/hivemq/issue/EDG-585/opc-ua-explicit-acceptanyservercertificate-flag-for-self-signed).


hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterUtils.java
```diff
@@ -15,28 +15,128 @@
...
 
+    private static final @NotNull Logger LOGGER = LoggerFactory.getLogger(ProtocolAdapterUtils.class);
+
     public static @NotNull ObjectMapper createProtocolAdapterMapper(final @NotNull ObjectMapper objectMapper) {
         final ObjectMapper copy = objectMapper.copy();
         copy.coercionConfigFor(LogicalType.POJO).setCoercion(CoercionInputShape.EmptyString, CoercionAction.AsNull);
         copy.coercionConfigFor(LogicalType.Collection)
                 .setCoercion(CoercionInputShape.EmptyString, CoercionAction.AsNull);
         copy.configure(MapperFeature.AUTO_DETECT_GETTERS, false);
+        copy.addHandler(new ReportUnknownPropertyHandler());
```
caoccao: `copy` is not a good name. Could we give it a more meaningful name?

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterUtils.java
```diff
@@ -15,28 +15,128 @@
...
+     *
+     * <p>Separate from {@link #createProtocolAdapterMapper} on purpose: that one is used to read adapter
+     * configurations arriving over the REST API and applies coercions this path does not want. Reading
+     * the configuration <em>file</em> goes through {@link ProtocolAdapterConfigConverter}, which holds
+     * the application-wide mapper directly — so a handler added only to the other one never runs on the
+     * path an operator's file actually takes.
+     */
+    public static @NotNull ObjectMapper withUnknownPropertyReporting(final @NotNull ObjectMapper objectMapper) {
+        final ObjectMapper copy = objectMapper.copy();
+        copy.addHandler(new ReportUnknownPropertyHandler());
```
caoccao: `copy` is not a good name. Could we give it a more meaningful name?

caoccao (changes requested):
There are some issues in https://linear.app/hivemq/document/sam-code-review-v01-b492c8c81589

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
```diff
@@ -714,6 +721,99 @@ public void refresh(final @NotNull List<ProtocolAdapterEntity> configs) {
...
+            }
+
+            // Keyed on the converted config's id, exactly as the previous Collectors.toMap was.
+            final String adapterId = config.getAdapterId();
+            if (!configuredAdapterIds.add(adapterId)) {
+                // Two adapters sharing an id: which configuration the operator meant is not knowable, so
+                // neither is applied and that adapter alone is left as it is. Previously this threw
+                // IllegalStateException out of Collectors.toMap and took the whole refresh with it.
+                protocolAdapterConfigs.remove(adapterId);
+                failedAdapterSet.add(adapterId);
```
caoccao, line 807: This is weird. Why is `failedAdapterSet.add(adapterId);` called here, not in the `catch` block? Is this a bug?

caoccao (changes requested):
I found one issue and the agent found something https://linear.app/hivemq/document/sam-code-review-v02-de1add083c87

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
caoccao: My bad, it's called in the catch block.

marregui: not a bug — deliberate. this branch handles duplicate ids, which are a configuration error without an exception: the second entity *converts fine*, so there is nothing to catch. the failure-branch counterpart is inside the catch a few lines up. the round-2 push adds tests for all the mixed orderings (fail+valid, valid+fail, 3-way) if you want to poke at it.

marregui:
round 03 is closed on the branch. head `1e0b87c22` — the six commits of the round, plus two after it.

| v03 finding | where |
| -- | -- |
| 1 High — `messageSecurityMode` degrades to `IGNORED` | `4b00389ca` — the same shared enum parser as the certificate-validation enums, at the Jackson, REST-mapper and config-file layers, plus a reflective sweep over the config package so the next enum cannot reintroduce it quietly |
| 2 Med — blank store path | `24fd280c4` — resolved as reject-blank on both stores, no migration: RJSF strips empty strings, so the form submits `truststore: {}` and never `{path: ""}` |
| 3 Med — core countersign | open, below |
| 4 Low — duplicate-id wording | `7f6a70577`, pinned in full against both an all-new duplicate and a duplicate of a running adapter by `1e0b87c22` |
| 5 Low — stale comments and fixture | `97c43c988` — four spots rather than two |
| 6 Low — no UI-schema regression test | `d40561388` (backend `UiSchemaSurfaceTest`) and `d910c8740` (the RJSF-level frontend test you preferred) |

docs: `f222c32f5` on hivemq/hivemq4-documentation#2675 this morning.

still open:

- **finding 3.** the generic core delta stays in this PR, no split. it needs an approving review from the owner of the protocol-adapter refresh — github's decision here is still `CHANGES_REQUESTED` and i cannot clear that myself.
- **CI.** the aggregate `check` passes and the maven 429 is gone. two cypress jobs are red under `continue-on-error`: `E2E-Workspace` (wizard group selection, unrelated to this branch) and `E2E-Core` (datahub protobuf message type). the second fails at `DatahubPage.ts:167`, the block `ee8958ab2` touched — the listbox assertion above it now passes, so the menu opens and the option list is empty. that moved the failure one line, it did not fix it. mine to chase; neither is a merge gate.

two follow-ups filed rather than folded in: EDG-886 (the UI-schema guard exists only for OPC UA — nine other adapters have the same failure mode and no test parses any of them) and EDG-887 (the frontend `opc-ua.ts` fixture is an Edge 2025.5 snapshot predating the presets/axes model).

caoccao (changes requested):
There is a critical bug.

https://linear.app/hivemq/document/sam-code-review-v04-2fa803f23819

caoccao (approved):

PR #1525, opened 2026-05-08
refactor(protocols): route wrapper start/stopAsync via lifecycle executor

caoccao:
## Summary

- `ProtocolAdapterWrapper.startAsync` / `stopAsync` previously used `CompletableFuture.runAsync(Runnable)` without an explicit executor, silently dispatching blocking adapter-lifecycle work (precheck, OPC UA / PLC4X / BACnet/IP connect, RocksDB writing init, disconnect) onto `ForkJoinPool.commonPool()`. That pool is sized for short CPU-bound tasks, has no backpressure, no named threads, and is not shut down by Edge — so a slow adapter could starve unrelated `parallel()` streams elsewhere in the JVM, and `ProtocolAdapterManager.shutdown()` could time out behind queued common-pool tasks.
- Inject the existing `adapterLifecycleExecutor` (4–10 threads, queue 20, `CallerRunsPolicy`, `protocol-adapter-lifecycle-%d`) into `ProtocolAdapterWrapper` via its constructor, and pass it explicitly to both `runAsync` calls. The manager already used this executor for its own `startAsync` / `stopAsync`; the wrapper now shares it so direct callers (`AbstractSubscriptionSampler.onSamplerError` → `wrapper.stopAsync(false)`, and `ProtocolAdapterManager.shutdown` → `wrapper.stopAsync(true).get(...)`) get the same bounded, named, backpressured semantics.
- Tests pass `Runnable::run` as the executor for deterministic synchronous execution; no test in this repo exercises the async dispatch path through the wrapper, so a synchronous executor is the cleanest substitute.

Tracking: EDG-582

## Test plan

- [ ] `./gradlew :hivemq-edge-build:hivemq-edge:check` (Spotless + ErrorProne + NullAway + unit tests for the `hivemq-edge` core module)
- [ ] `./gradlew :hivemq-edge-build:hivemq-edge-test:test` (composite integration suite that exercises adapter lifecycle end-to-end)
- [ ] Manual smoke: start Edge with a slow OPC UA endpoint, observe `protocol-adapter-lifecycle-*` threads in a thread dump during start/stop (no `commonPool-worker-*` threads doing adapter work).
- [ ] Confirm `manager.shutdown()` completes within the 5-second timeout when adapters are mid-start.

## Out of scope (deliberately deferred)

- `AssetMapperManager.stopAll()` and `DataCombinerManager.stopAll()` — same `runAsync`-without-executor anti-pattern, separate domains. Tracked as follow-ups in `Documents/Diary/2026/202605/fsm-patch-01.md` §7.
- Migrate `Runtime.getRuntime().addShutdownHook(...)` for the manager's executors to the project's `ShutdownHooks` registry — pre-existing inconsistency, would couple this PR to a wider DI rewire.
- Per-protocol connect executors (OPC UA, PLC4X, BACnet/IP) — adapter-internal idioms, separate review.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
+++ b/hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
@@ -538,7 +538,8 @@ public class ProtocolAdapterManager {
                         perModule,
                         tagManager,
                         northboundConsumerFactory,
-                        protocolAdapterWritingService);
+                        protocolAdapterWritingService,
+                        adapterLifecycleExecutor);
                 protocolAdapterMetrics.increaseProtocolAdapterMetric(configProtocolId);
                 return wrapper;
             });
--- a/hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
+++ b/hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
@@ -220,6 +220,17 @@ public class ProtocolAdapterManager {
                 }
             }
         });
+        // Release the manager-owned executor threads so embedded Edge instances (tests) don't
+        // leak parked threads across runs. The JVM-exit shutdown hook is kept as a fallback and
+        // is idempotent.
+        adapterLifecycleExecutor.shutdown();
+        executorService.shutdown();
+        try {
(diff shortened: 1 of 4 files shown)


marregui (approved):
LGTM !

PR #1539, opened 2026-05-21
Change building of license files

codepitbull:
**Motivation**

Resolves #EDG-654

**Changes**
- Make sure licenses are not checked in
- Build license and sbom on the fly
- integrate license into artifacts
- ad a global license check task to composite


caoccao (approved):
LGTM. Is the final layout of those file changed in the release jar? I assume it's not.

PR #1546, opened 2026-06-03
Fix OPC UA tag-schema problems

codepitbull:
**Motivation**

Resolves #EDG-683

## Problem 1:
A struct field typed with a custom non-structure DataType that has no DataTypeDefinition (e.g. a String subtype or enum in the server namespace) made tag-schema generation return 500 "Problem accessing node"
  (RuntimeException "Unsupported type definition: null"). JsonSchemaGenerator guessed builtin-vs-struct from the namespace prefix, so any custom-namespace field was recursed into as a struct and threw.

### Fix
Resolve fields via DataTypeTree.getBuiltinType instead: ExtensionObject -> recurse, scalar-mappable builtin
  -> scalar (String alias -> string), Variant/DataValue/DiagnosticInfo -> render as any. Also degrade the unknown-definition branch to any and guard the top-level collectTypeInfo path likewise.

## Problem 2:
A tag on a structured node expanded VALUE into its fields, but a field that is itself an array (e.g. Int16[10], ValueRank 1) was flattened to a scalar integer. Direct array tags were unaffected because the top-level path already reads the node  array dimensions; the struct field loop hardcoded arrayDimensions=null.

### Fix:
Derive array dimensions from the StructureField (ValueRank >= 1) and pass them into the field FieldInformation so the field renders as an array of its element type. Array-of-struct fields remain a known limitation (the schema builder
  only emits scalar items).

## Problem 3:
Accept JSON int nodes when writing OPC UA Int64/UInt64
extractLong (Int64) and extractUnsignedLong (UInt64) only accepted Jackson LongNodes (isLong()), which is false for values that fit in an int. A normal value like 290 is an IntNode, so writing it to an Int64 node threw "Can not convert '290' of class 'IntNode' to 'Int64'" and the message was dropped, contradicting the documented JSON-integer -> Int64 mapping.

### Fix:
Accept isInt() || isLong() (as extractUInteger already does) and read via longValue().


marregui (approved):
lgtm

PR #1547, opened 2026-06-03
EDG-688: Apply configured publishingInterval on OPC-UA subscription create

mschoenert-hivemq:
**Motivation**

Resolves #EDG-688

The OPC-UA adapter's configured `publishingInterval` was silently ignored — every Subscription ran at Eclipse Milo's 1000 ms default regardless of configuration. The helper called `subscription.create()` *before* `setPublishingInterval(...)`; the post-create setter only queues a `ModifySubscription` diff that the adapter never flushed via `modify()`. Reported by Miele during a POC (100 ms requested, 1000 ms observed on the server-side subscription, despite another OPC-UA client honouring 100 ms against the same PLC).

The `onTransferFailed` recovery path additionally never called `setPublishingInterval` at all — same end state.

**Changes**

Replace the static `createNewSubscription` helper with a fluent builder:

```java
newSubscription(client)
    .publishingInterval(config.getOpcuaToMqttConfig().publishingInterval())
    .create()
```

The builder applies its setters between `new OpcUaSubscription(client)` and `subscription.create()`, so the configured value is sent in the initial `CreateSubscription` request rather than queued as a deferred modification. The staged lifecycle becomes structural — callers cannot configure after creating because there is no setter on the returned `Optional<OpcUaSubscription>`. Each future per-Subscription knob (lifetimeCount, maxKeepAliveCount, priority, …) becomes a new builder setter rather than another positional ordering hazard.

Both call sites (`subscribe` and `onTransferFailed`) now route through the builder. After `create()` returns, the requested-vs-revised publishingInterval is logged so future "interval not honoured" reports can be diagnosed without source-reading: WARN if the server revised by more than 1 ms, INFO otherwise.

`samplingInterval` stays coupled to `publishingInterval` — decoupling is in EDG-605's scope, not this fix.

**Tests**

New `OpcUaClientConnectionTest.configuredPublishingInterval_isAppliedOnTheServer()` drives a real `OpcUaClientConnection` against the embedded Milo test server with a non-default 250 ms config, then asserts the server-side `Subscription.getPublishingInterval()` reads 250 ms. Confirmed the test fails before the fix (sees Milo's 1000 ms default) and passes after.

Full OPC-UA module unit suite: 197/0/0/1.
OPC-UA integration suite (`com.hivemq.edge.modules.opcua.*` in `hivemq-edge-test`): 96/0/0/0, including `OpcUaAdapterLifecycleIT`, `SubscriptionResumeOpcUaIT`, and `OpcUaConnectionRetryIT` which exercise the lifecycle paths this change rewires.

### DIFF
--- a/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/listeners/OpcUaSubscriptionLifecycleHandler.java
+++ b/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/listeners/OpcUaSubscriptionLifecycleHandler.java
@@ -101,32 +101,76 @@ public class OpcUaSubscriptionLifecycleHandler implements OpcUaSubscription.Subs
     }
 
     /**
-     * Creates a new OPC UA subscription.
-     * If the subscription is created successfully, it returns an Optional containing the subscription.
-     * If the subscription creation fails, it returns an empty Optional.
+     * Starts construction of a new OPC UA subscription. Configure with the builder's setters, then
+     * call {@link Builder#create()} to push the subscription to the server.
      *
      * @param client the OPC UA client
-     * @return an Optional containing the created subscription or empty if creation failed
+     * @return a builder
      */
-    public static @NotNull Optional<OpcUaSubscription> createNewSubscription(final @NotNull OpcUaClient client) {
-        log.debug("Creating new OPC UA subscription");
-        final OpcUaSubscription subscription = new OpcUaSubscription(client);
-        try {
-            subscription.create();
-            return subscription
-                    .getSubscriptionId()
-                    .map(subscriptionId -> {
-                        log.trace("New subscription ID: {}", subscriptionId);
-                        return subscription;
-                    })
-                    .or(() -> {
-                        log.error("Subscription not created on the server");
-                        return Optional.empty();
-                    });
-        } catch (final UaException e) {
-            log.error("Failed to create subscription", e);
+    public static @NotNull Builder newSubscription(final @NotNull OpcUaClient client) {
+        return new Builder(client);
+    }
+
+    /** Fluent builder for an {@link OpcUaSubscription}. */
+    public static final class Builder {
+
+        private final @NotNull OpcUaClient client;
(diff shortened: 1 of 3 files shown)


mschoenert-hivemq:
I've added an empty commit in composite to force a CI Jenkins run.
There is one flaky tests, apart from that it looks good
https://jenkins.cicd.pd.hmq.dev/job/HiveMQ4/job/hivemq-edge-composite/job/fix%252FEDG-688%252Fopcua-publish-interval-is-ignored/

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/listeners/OpcUaSubscriptionLifecycleHandler.java
```diff
@@ -101,32 +101,76 @@ public OpcUaSubscriptionLifecycleHandler(
...
+        public @NotNull Optional<OpcUaSubscription> create() {
+            log.debug("Creating new OPC UA subscription with publishingInterval={}ms", publishingIntervalMs);
+            final OpcUaSubscription subscription = new OpcUaSubscription(client);
+            subscription.setPublishingInterval((double) publishingIntervalMs);
+            try {
+                subscription.create();
+                final double revised = subscription.getPublishingInterval();
+                if (Math.abs(revised - publishingIntervalMs) > 1.0) {
+                    log.warn(
+                            "OPC UA server revised publishingInterval: requested={}ms, revised={}ms",
```
codepitbull, line 150: Sure this shouldn't rather be an actual error case?

codepitbull (approved):
All good code wise.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/listeners/OpcUaSubscriptionLifecycleHandler.java
mschoenert-hivemq: Yes. I want to avoid - with this release - that stuff that worked before suddenly fails.

PR #1579, opened 2026-06-21
Houston follow-up: atomic adapter swap + browse serialisation/readiness (EDG-602, 493, 576, 577)

marregui:
Bundled Houston follow-up (one PR for economy). Four cohesive lifecycle/browse fixes, one commit per ticket.

## Tickets
- **EDG-602** — https://linear.app/hivemq/issue/EDG-602 — adapter update ran delete-then-recreate on `protocolAdapterMap`, leaving a window where `getAdapterDomainTags` returned **404** (flaky `DeviceTagBrowseStressTest`; "adapter unknown state"). Now an **atomic swap**: build replacement → single `put` → stop displaced instance → start new. Adapter id never absent. Both update paths routed through it.
- **EDG-576** — https://linear.app/hivemq/issue/EDG-576 — the browse `Semaphore` was created per `browse()` call, serialising only within one browse. Moved to **one permit per `OpcUaProtocolAdapter`**, so concurrent REST browses sharing one Milo client are serialised (fixes non-deterministic S7-1500 counts).
- **EDG-577** — https://linear.app/hivemq/issue/EDG-577 — `CONNECTED` fires before the namespace table / data-type tree are hydrated, so an early browse is non-deterministic. Adds an **OPC-UA-local readiness signal** (kept off the shared SDK `ConnectionStatus` enum): post-connect async warm-up → `BROWSE_READY`; browse REST answers **503 + Retry-After** until ready.
- **EDG-493** — https://linear.app/hivemq/issue/EDG-493 — classify **additive-only imports** (+ debug log) before `updateAdapter`; groundwork for the deferred skip-restart path.
- **EDG-579** — duplicate of EDG-576 (closed).

## Verification
- Unit-verified: `ProtocolAdapterManagerTest` (28), `DeviceTagBrowsingResourceImplTest` (42), `OpcUaNodeBrowserTest` (38), `OpcUaProtocolAdapterTest` (29), `DeviceTagImporterTest` (47) — all green; both modules compile. New regression tests for the swap invariant, shared-semaphore serialisation, and the 503 gate.
- **Not** hardware-verified: EDG-576's S7-1500 consistency and EDG-577's warm-up against a real Milo server need the physical PLC.

## Design notes for review
- EDG-577 readiness is **OPC-UA-scoped**, not on the shared SDK `ConnectionStatus` enum (deliberate, lower blast radius). Consequence: it's surfaced via the browse 503→200 transition, not a status-API field.

## Follow-up (separate, testsuite repo)
Replace the two `Thread.sleep` workarounds in `DeviceTagBrowseImportHardwareTest` with a wait on the browse 503→200 transition.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/edge/adapters/browse/rest/DeviceTagBrowsingResourceImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/edge/adapters/browse/rest/DeviceTagBrowsingResourceImpl.java
@@ -126,6 +126,19 @@ public class DeviceTagBrowsingResourceImpl extends AbstractApi implements Device
                     Response.Status.CONFLICT, "Adapter '" + adapterId + "' does not support bulk tag browsing");
         }
 
+        // EDG-577: the adapter may be CONNECTED but still loading its address-space metadata, during which a
+        // browse would be non-deterministic. Answer 503 with Retry-After so the caller retries once it is ready.
+        if (!browser.isBrowseReady()) {
+            return Response.status(Response.Status.SERVICE_UNAVAILABLE)
+                    .header("Retry-After", "2")
+                    .entity(errorBody(
+                            Response.Status.SERVICE_UNAVAILABLE.getReasonPhrase(),
+                            "Adapter '" + adapterId
+                                    + "' is connected but still loading device metadata; retry shortly"))
+                    .type(APPLICATION_JSON)
+                    .build();
+        }
+
         // Browse
         if (maxDepth != null && maxDepth < 0) {
             return errorResponse(Response.Status.BAD_REQUEST, "maxDepth must be >= 0 (0 = unlimited)");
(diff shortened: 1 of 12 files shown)


hivemq-edge/src/main/java/com/hivemq/edge/adapters/browse/rest/DeviceTagBrowsingResourceImpl.java
```diff
@@ -126,6 +126,19 @@ public DeviceTagBrowsingResourceImpl(
                     Response.Status.CONFLICT, "Adapter '" + adapterId + "' does not support bulk tag browsing");
         }
 
+        // EDG-577: the adapter may be CONNECTED but still loading its address-space metadata, during which a
+        // browse would be non-deterministic. Answer 503 with Retry-After so the caller retries once it is ready.
+        if (!browser.isBrowseReady()) {
+            return Response.status(Response.Status.SERVICE_UNAVAILABLE)
```
caoccao: Is this error documented in openapi spec? We should define a dedicated error object in openapi spec and rely on the openapi generated entity for the error.

caoccao (approved):
The error message issues are not blocking.

hivemq-edge/src/main/java/com/hivemq/edge/adapters/browse/rest/DeviceTagBrowsingResourceImpl.java
marregui: Good catch, no, the 503 isn't in the spec yet. The browse operation already standardises its other errors (404/409/504) on ProblemDetails, so I'll bring the 503 in line with that:
  
 1. Add a '503' response to browseDeviceTags in ext/hivemq-edge-openapi.yaml referencing #/components/schemas/ProblemDetails (description: "Adapter connected but address-space metadata still loading; retry after the indicated delay"), and document the Retry-After response header there too.
 2. Regenerate, and return the generated ProblemDetails entity from the handler instead of the hand-rolled errorBody(...) map for this path.

PR #1587, opened 2026-06-27
EDG-747: Replace user-facing Pulse names in Edge frontend

mschoenert-hivemq:
Rename the user-facing "Pulse" / "Pulse Agent" strings to "HiveMQ Platform" per Product (Magi's worksheet), reframing activation language to connection language (Activate -> Connect, activation token -> connection string). "Assets" keeps its name (drop the Pulse qualifier).

- translation.json: all user-facing Pulse strings replaced.
- NodePulse.tsx: canvas toolbar title now uses the existing i18n key instead of the hard-coded node label.
- nodes-utils.ts: the node data.label literal updated too, since it still surfaces in filter panels and the property drawer.

Out of scope (deferred): the docs URL pulse-asset-mapping.html (docs team owns it; restructuring in progress), config, module rename, logs.

**Motivation**

Resolves #<issue>

**Changes**


hivemq-edge-frontend/cypress/e2e/pulse/activation.spec.cy.ts
```diff
@@ -39,16 +39,16 @@ describe('Pulse Agent Activation', () => {
       homePage.task(ONBOARDING.TASK_ADAPTER).should('contain.text', 'From Devices To HiveMQ Edge')
       homePage.task(ONBOARDING.TASK_BRIDGE).should('contain.text', 'From HiveMQ Edge to the Enterprise')
       homePage.task(ONBOARDING.TASK_CLOUD).should('contain.text', 'Connect To HiveMQ Cloud')
-      homePage.task(ONBOARDING.TASK_PULSE).should('contain.text', 'Connect to HiveMQ Pulse')
+      homePage.task(ONBOARDING.TASK_PULSE).should('contain.text', 'Connect to HiveMQ Platform')
 
       homePage.taskSections(ONBOARDING.TASK_PULSE).should('have.length', 1)
       homePage.taskSection(ONBOARDING.TASK_PULSE, 0).within(() => {
-        pulseActivationPanel.trigger.should('have.text', 'Activate Pulse')
+        pulseActivationPanel.trigger.should('have.text', 'Connect with HiveMQ Platform')
```
caoccao, line 46: API, variable names would be in the scope of this change?

caoccao (approved):
LGTM

hivemq-edge-frontend/cypress/e2e/pulse/activation.spec.cy.ts
mschoenert-hivemq: They are NOT in the scope. Magi and I agreed that we do the absolute minimum. No changes to the API, no changes to the element in the config, no internal variable names, we don't rename the external module, etc.

PR #1588, opened 2026-06-28
EDG-762: Redesign InternalTopicFilterSubscriber with factory + builder

mschoenert-hivemq:
Replace the abstract-class-and-subclass design with a concrete, final class obtained from an injected InternalTopicFilterSubscriberFactory via a fluent builder. The class had zero users, so the redesign carries no migration cost.

Why: dynamic bridging needs runtime mutability of topic filters — change what a live subscription carries without tearing down and re-establishing it. The v1 lifecycle was monolithic (filter set fixed at construction).

- Processor lambda (withProcessor) replaces the overridden process().
- Factory holds the three Edge singletons; callers no longer thread them.
- Lifecycle decomposed into verbs: attach / consume / pause / detach / deallocate; start/stop are convenience compositions.
- Runtime filter mutation: withTopicFilter (absolute, diffed against the tree when attached), addTopicFilter / removeTopicFilter (relative); same three verbs on builder and subscriber. Empty filter set is valid.

The underlying mechanism (topic-tree add/remove, publish-available callback, SingleWriter poll loop, queue clear) is unchanged; the v1 explanatory comments are preserved.

Tests: a real-LocalTopicTree test pins the subscribe-id == poll-id contract, plus state-machine, filter-reconciliation, and builder validation. Full hivemq-edge suite green; Spotless clean.

**Motivation**

Resolves #<issue>

**Changes**

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java
+++ b/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java
@@ -109,6 +109,18 @@ public final class InternalTopicFilterSubscriber {
     // The per-message handler. Set once at construction (Builder requires it).
     private final @NotNull Processor processor;
 
+    // The factory that built this subscriber, kept as a back-reference so that attach()/detach() can
+    // (de)register this subscriber in the factory's registry — see isExcludedIngressClientId() below.
+    private final @NotNull InternalTopicFilterSubscriberFactory factory;
+
+    // excludedIngressClientId — ingress-exclusion (generalized No Local). If set, a message whose
+    //            INGRESS client id (the publishing client, i.e. the distributor's `sender`) equals this
+    //            value must NOT be delivered to us. Set once via the builder, immutable after build();
+    //            null means "no exclusion — accept from everyone". The check happens at distribution
+    //            time (PublishDistributorImpl), BEFORE the message is queued, so an excluded message
+    //            never enters our queue. See isExcludedIngressClientId().
+    private final @Nullable String excludedIngressClientId;
+
     // ── mutable lifecycle state, all touched only via the verbs below ────────────────────────────
     // topicFilters — the CURRENT desired filter set. When detached this is just a remembered set
     //                (replayed by the next attach()); when attached the topic tree is kept reconciled
@@ -123,6 +135,12 @@ public final class InternalTopicFilterSubscriber {
     //             drained and handed to the processor).
     private boolean consuming = false;
 
+    // deallocated — true once deallocate() has cleared the queue. TERMINAL: once deallocated the
+    //               subscriber is dead and cannot be resurrected; every verb is a no-op (or, for
+    //               attach/consume, rejected) from then on. This flag makes deallocate()/stop() safely
+    //               idempotent and guards the other verbs against use-after-deallocate.
+    private boolean deallocated = false;
+
     // endregion
 
     // region constructor — package-private, only the factory/builder construct this
@@ -135,18 +153,24 @@ public final class InternalTopicFilterSubscriber {
     //                   configuration ID of the owning instance.
     // processor       — the per-message handler (runs on the SingleWriter thread).
     // initialFilters  — the topic filter set assembled in the builder (may be empty; that is valid).
+    // excludedIngressClientId — ingress-exclusion id, or null for none (see the field above).
+    // factory         — the factory that built us, used as the registry for ingress-exclusion lookups.
     //
     InternalTopicFilterSubscriber(
             final @NotNull String componentPrefix,
             final @NotNull String instanceId,
             final @NotNull Processor processor,
             final @NotNull Set<String> initialFilters,
+            final @Nullable String excludedIngressClientId,
+            final @NotNull InternalTopicFilterSubscriberFactory factory,
             final @NotNull LocalTopicTree topicTree,
             final @NotNull ClientQueuePersistence clientQueuePersistence,
             final @NotNull SingleWriterService singleWriterService) {
         this.clientId = INTERNAL_SUBSCRIBER_PREFIX + componentPrefix + "::" + instanceId;
         this.processor = processor;
         this.topicFilters.addAll(initialFilters);
+        this.excludedIngressClientId = excludedIngressClientId;
+        this.factory = factory;
         this.topicTree = topicTree;
         this.clientQueuePersistence = clientQueuePersistence;
         this.singleWriterService = singleWriterService;
@@ -185,22 +209,27 @@ public final class InternalTopicFilterSubscriber {
 
         private final @NotNull String componentPrefix;
         private final @NotNull String instanceId;
+        private final @NotNull InternalTopicFilterSubscriberFactory factory;
         private final @NotNull LocalTopicTree topicTree;
         private final @NotNull ClientQueuePersistence clientQueuePersistence;
         private final @NotNull SingleWriterService singleWriterService;
 
         private @Nullable Processor processor = null; // required; checked in build()
         private final @NotNull Set<String> topicFilters = new LinkedHashSet<>();
+        private @Nullable String excludedIngressClientId = null; // optional; ingress-exclusion
 
-        // Package-private — only the factory creates builders (it supplies the injected singletons).
+        // Package-private — only the factory creates builders (it supplies the injected singletons and
+        // itself, as the registry the built subscriber registers with on attach()).
         Builder(
                 final @NotNull String componentPrefix,
                 final @NotNull String instanceId,
+                final @NotNull InternalTopicFilterSubscriberFactory factory,
                 final @NotNull LocalTopicTree topicTree,
                 final @NotNull ClientQueuePersistence clientQueuePersistence,
                 final @NotNull SingleWriterService singleWriterService) {
             this.componentPrefix = componentPrefix;
             this.instanceId = instanceId;
+            this.factory = factory;
             this.topicTree = topicTree;
             this.clientQueuePersistence = clientQueuePersistence;
             this.singleWriterService = singleWriterService;
@@ -209,6 +238,15 @@ public final class InternalTopicFilterSubscriber {
         // Required: the per-message work.
         public @NotNull Builder withProcessor(final @NotNull Processor processor) {
             this.processor = processor;
+            return this;
+        }
+
+        // Optional: ingress-exclusion (generalized No Local). A message whose ingress (publishing)
+        // client id equals this value will not be delivered to the built subscriber. Build-time only;
+        // immutable after build(). For a bridge, this is the peer it forwards to — so a message it sent
+        // us is not sent straight back. The plain No Local case is excludedIngressClientId == our own id.
+        public @NotNull Builder withExcludedIngressClientId(final @NotNull String excludedIngressClientId) {
+            this.excludedIngressClientId = excludedIngressClientId;
             return this;
         }
 
@@ -259,6 +297,8 @@ public final class InternalTopicFilterSubscriber {
                     instanceId,
                     processor,
                     topicFilters,
+                    excludedIngressClientId,
+                    factory,
                     topicTree,
                     clientQueuePersistence,
                     singleWriterService);
@@ -280,6 +320,7 @@ public final class InternalTopicFilterSubscriber {
     }
 
     public synchronized @NotNull InternalTopicFilterSubscriber withTopicFilter(final @NotNull List<String> newFilters) {
+        throwIfDeallocated();
         final Set<String> target = new LinkedHashSet<>(newFilters);
         if (attached) {
             // Reconcile the tree with the new target: remove what is no longer wanted, add what is new.
@@ -305,6 +346,7 @@ public final class InternalTopicFilterSubscriber {
 
     public synchronized @NotNull InternalTopicFilterSubscriber addTopicFilter(
             final @NotNull List<String> topicFiltersToAdd) {
+        throwIfDeallocated();
         for (final String topicFilter : topicFiltersToAdd) {
             if (topicFilters.add(topicFilter) && attached) {
                 addToTree(topicFilter);
@@ -319,6 +361,7 @@ public final class InternalTopicFilterSubscriber {
 
     public synchronized @NotNull InternalTopicFilterSubscriber removeTopicFilter(
             final @NotNull List<String> topicFiltersToRemove) {
+        throwIfDeallocated();
         for (final String topicFilter : topicFiltersToRemove) {
             if (topicFilters.remove(topicFilter) && attached) {
                 removeFromTree(topicFilter);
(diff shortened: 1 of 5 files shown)


hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriberFactory.java
```diff
@@ -0,0 +1,97 @@
...
+            final @NotNull SingleWriterService singleWriterService) {
+        this.topicTree = topicTree;
+        this.clientQueuePersistence = clientQueuePersistence;
+        this.singleWriterService = singleWriterService;
+    }
+
+    // ── registry — called by the subscriber's attach()/detach(), read by the distributor ────────────
+
+    void register(final @NotNull InternalTopicFilterSubscriber subscriber) {
+        registry.put(subscriber.clientId(), subscriber);
```
caoccao: There's no dedup. Is this behavior expected?

hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriberFactory.java
```diff
@@ -0,0 +1,97 @@
...
+    // Begin building a subscriber for the given component identity.
+    //
+    // componentPrefix — the Edge component type, unique across Edge (e.g. "tynebridge"). Becomes the
+    //                   middle segment of the reserved internal client ID "$INTERNAL::<prefix>::<id>".
+    // instanceId      — unique within the componentPrefix namespace; typically the config ID of the
+    //                   owning instance.
+    //
+    public @NotNull InternalTopicFilterSubscriber.Builder create(
+            final @NotNull String componentPrefix, final @NotNull String instanceId) {
+        return new InternalTopicFilterSubscriber.Builder(
```
caoccao, line 102: `create` doesn't check if there's an existing subscriber using the identical prefix and id. Is this expected?

hivemq-edge/src/test/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriberTest.java
```diff
@@ -0,0 +1,285 @@
...
+                .isInstanceOf(IllegalStateException.class)
+                .hasMessageContaining("deallocated");
+        assertThatThrownBy(() -> s.withTopicFilter("x/#"))
+                .isInstanceOf(IllegalStateException.class)
+                .hasMessageContaining("deallocated");
+    }
+
+    @Test
+    void attach_registersInFactory_detachDeregisters() {
+        final InternalTopicFilterSubscriber s = build("sensors/#");
```
caoccao, line 252: What will happen, if there are multiple calls to the `build()` with the same filter?

caoccao (changes requested):
I hold some concerns.

hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriberFactory.java
mschoenert-hivemq: Good catch — and you're right that all three of these point at the same gap: a duplicate `componentPrefix::instanceId` was silently accepted. That was an oversight, not intended behaviour. Fixed:

- **`build()` now registers the subscriber eagerly** (not on `attach()`), and **`register()` rejects a duplicate identity** — `putIfAbsent`, throwing `IllegalStateException` if the `componentPrefix::instanceId` is already in use. So a second `build()` with the same identity is refused at construction rather than clobbering the first in the registry.
- **`deallocate()` deregisters**, freeing the identity for reuse once the queue is destroyed. Registration now brackets the subscriber's whole live span (build→deallocate), not just its attached span — which also means the registry is the single uniqueness guard *and* the publish-path lookup.
- The identity is now genuinely 1:1 with the live subscriber object: it can't collide, because the only way to express a collision (a second `build()`) throws.

Re #3 specifically — added `build_duplicateIdentity_throws_andDeallocateFreesItForReuse` to pin exactly this: two builds with the same identity → second throws; after the first deallocates, the same identity builds cleanly again. Also reworked the old `attach_registersInFactory_detachDeregisters` test into `build_registersInFactory_deallocateDeregisters` to match the new boundary.

The design doc / Lore page are updated to describe the build→deallocate registration span. Thanks for pushing on this.

hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java
```diff
@@ -59,123 +85,494 @@ public abstract class InternalTopicFilterSubscriber {
...
+    // the genuine additions/removals are pushed to the tree, so a withTopicFilter(wholeNewSet) becomes
+    // exactly the diff against what is currently registered.
     //
-    public void start() {
-        doStart();
+    public synchronized @NotNull InternalTopicFilterSubscriber withTopicFilter(final @NotNull String topicFilter) {
+        return withTopicFilter(List.of(topicFilter));
+    }
+
+    public synchronized @NotNull InternalTopicFilterSubscriber withTopicFilter(final @NotNull List<String> newFilters) {
```
caoccao, line 329: If all synchronized methods in this class are the only ones that have write access to the topic tree, this algorithm is fine. Otherwise, there will be a race condition. Please confirm.

caoccao: ## LocalTopicTree Write API

- [`LocalTopicTree.addTopic(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/mqtt/topic/tree/LocalTopicTree.java:75)
- [`LocalTopicTree.removeSubscriber(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/mqtt/topic/tree/LocalTopicTree.java:404)

Internal mutation helpers:

- [`TopicTreeNode.addChildNodeIfAbsent(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/mqtt/topic/tree/TopicTreeNode.java:58)
- [`MatchingNodeSubscriptions.addSubscriber(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/mqtt/topic/tree/MatchingNodeSubscriptions.java:72)
- [`MatchingNodeSubscriptions.removeSubscriber(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/mqtt/topic/tree/MatchingNodeSubscriptions.java:97)

## Production Write Paths

- Startup bootstrap: [`TopicTreeStartup.populateTopicTree()`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/mqtt/topic/tree/TopicTreeStartup.java:73) loads persisted subscriptions and calls `topicTree.addTopic(...)` at lines 97 and 108.

- Normal MQTT SUBSCRIBE / UNSUBSCRIBE: [`IncomingSubscribeService`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/mqtt/handler/subscribe/IncomingSubscribeService.java:323) and [`UnsubscribeHandler`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/mqtt/handler/unsubscribe/UnsubscribeHandler.java:88) call subscription persistence. Actual tree writes happen in [`ClientSessionSubscriptionPersistenceImpl`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/persistence/clientsession/ClientSessionSubscriptionPersistenceImpl.java:127), lines 127, 154, 346, 228, 235, 272, and 440.

- Extension subscription service: [`SubscriptionStoreImpl`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/extensions/services/subscription/SubscriptionStoreImpl.java:95) exposes add/remove subscription APIs to extensions, forwarding to `ClientSessionSubscriptionPersistenceImpl`.

- Session cleanup: [`ClientSessionPersistenceImpl`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/persistence/clientsession/ClientSessionPersistenceImpl.java:135), lines 135, 295, and 347 call `subscriptionPersistence.removeAll(...)`, which removes subscriptions from `LocalTopicTree`.

- Bridge forwarding: [`BridgeService.internalStartBridge(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/bridge/BridgeService.java:239) starts forwarders. [`MessageForwarderImpl.addForwarder(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java:156) adds shared subscriptions. Stop/remove goes through [`BridgeService`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/bridge/BridgeService.java:350) to [`MessageForwarderImpl.removeForwarder(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java:279).

- Sampling: [`SamplingResourceImpl.startSamplingForTopic(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/api/resources/impl/SamplingResourceImpl.java:105) calls [`SamplingService.startSampling(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/sampling/SamplingService.java:57), which calls `addTopic(...)`. [`SamplingService.stopSampling(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/sampling/SamplingService.java:67) calls `removeSubscriber(...)`.

- Data combining topic-filter sources: [`DataCombinerManager`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/combining/runtime/DataCombinerManager.java:226) starts runtimes. [`DataCombiningRuntime`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/combining/runtime/DataCombiningRuntime.java:129) creates `InternalTopicFilterConsumer`; its `start()` writes at line 268 and `close()` removes at line 262.

- Generic internal subscribers: [`InternalTopicFilterSubscriber`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java:122) provides `start()` and `stop()` wrappers that call `addTopic(...)` at line 132 and `removeSubscriber(...)` at line 156.

- Commercial bidirectional adapter writing: [`ProtocolAdapterWrapper.startWriting()`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterWrapper.java:787) starts writing. [`ProtocolAdapterWritingServiceImpl`](hivemq-edge-commercial-modules/hivemq-edge-bidirectional-adapter/src/main/java/com/hivemq/edge/modules/bidirectional/write/ProtocolAdapterWritingServiceImpl.java:130) creates subscriptions via `localTopicTree.addTopic(...)` at line 197. Stop goes through [`ProtocolAdapterWrapper.stopWriting()`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterWrapper.java:818) to [`ProtocolAdapterWritingServiceImpl`](hivemq-edge-commercial-modules/hivemq-edge-bidirectional-adapter/src/main/java/com/hivemq/edge/modules/bidirectional/write/ProtocolAdapterWritingServiceImpl.java:177), then removes at line 218.

## Separate Retained-Message Topic Trees

These use `PublishTopicTree`, not `LocalTopicTree`.

- [`PublishTopicTree.add(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/persistence/local/xodus/PublishTopicTree.java:31)
- [`PublishTopicTree.remove(...)`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/persistence/local/xodus/PublishTopicTree.java:44)

Retained-message writers:

- Memory retained persistence: [`RetainedMessageMemoryLocalPersistence`](hivemq-edge/hivemq-edge/src/main/java/com/hivemq/persistence/local/memory/RetainedMessageMemoryLocalPersistence.java:104), lines 104, 119, 159, and 188.
- Commercial RocksDB retained persistence: [`RetainedMessageRocksDBLocalPersistence`](hivemq-edge-commercial-modules/hivemq-edge-mqtt-persistence/src/main/java/com/hivemq/edge/modules/persistence/retained/rocksdb/RetainedMessageRocksDBLocalPersistence.java:190), lines 190, 240, 302, and 347.
- Commercial Xodus retained persistence: [`RetainedMessageXodusLocalPersistence`](hivemq-edge-commercial-modules/hivemq-edge-mqtt-persistence/src/main/java/com/hivemq/edge/modules/persistence/retained/xodus/RetainedMessageXodusLocalPersistence.java:193), lines 193, 236, 306, and 348.

caoccao (changes requested):
I hold only one concern.

hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java
mschoenert-hivemq: Good point. This is actually safe.

**1. All changes to the topic tree for this subscription go through the ITFS code.**
- To be honest - this isn't strictly enforced, or strictly enforceable. I.e. nothing stops someone from synthesizing `$INTERNAL::component::instance` and calling `LocalTopicTree.addTopic(...)` / `removeSubscriber(...)` directly, bypassing ITFS entirely.
- But we can add a stern warning against this in the docs, and could even check it mechanically — a grep for the literal `"$INTERNAL::"` over the source shows it appears in exactly one place today (`InternalTopicFilterSubscriber`'s own constant); everywhere else already goes through `INTERNAL_SUBSCRIBER_PREFIX`. That invariant is a one-line source-scan test if we want it enforced rather than just documented.

**2. Even if someone did that, it's not catastrophic., i.e. nothing would crash.**
- `LocalTopicTree` is concurrency-safe on its own — it has its own internal striped locking, independent of anything in ITFS.
- But ITFS's own diff logic (`withTopicFilter`) assumes the tree and its remembered `topicFilters` set stay aligned. A direct, out-of-band tree write breaks that assumption silently — nothing crashes, but the tree and ITFS's own state diverge, since ITFS only ever diffs against what *it* remembers.

**3. All modifications of the subscription's own internal state are synchronized on one monitor.**
- So if two threads concurrently call, say, `subscription.withTopicFilter(...)` and `subscription.addTopicFilter(...)`, they execute one after the other, never interleaved.
- It's only impossible to predict which one goes first — not whether they'll corrupt anything. Same monitor, same guarantee, regardless of which pair of verbs is racing.

So: the tree's own locking carries tree-safety: our `synchronized` carries subscriber-state-safety; and the only real gap is #1 — bypassing ITFS to touch the tree directly — which is a discipline convention we can document, and optionally pin with a test, rather than a bug in this code.

caoccao: Sounds good.

caoccao (approved):

PR #1590, opened 2026-06-30
EDG-764: Add southbound write to the EtherNet/IP ODVA adapter

mschoenert-hivemq:
## Description
Adds southbound (write) support to the new EtherNet/IP ODVA adapter (`hivemq-edge-module-etherip-cip-odva`). Stacked on #1433 (EDG-96, the adapter merge); this PR covers the entire write path.

The model is two per-tag flags:
- **RW** — `READ_ONLY` / `WRITE_ONLY` / `READ_WRITE`: tag direction. Gates whether a southbound mapping is allowed, and is part of the tag grouping key (read-only and write-only tags at the same address don't mix).
- **WM** — `OVERWRITE_ZERO` / `READ_MODIFY_WRITE`: how a write is applied to the enclosing CIP attribute (the atomic write unit). RMW preserves sibling bytes; OVERWRITE_ZERO is for full composites and write-only/command attributes.

A composite write is a single atomic `Set_Attribute_Single`. Cross-attribute atomicity is out of scope (CIP models atomicity at the attribute).

Full design and code map: see the EDG-764 Linear document.

## Status
Draft — landing incrementally on this branch:
- [x] RW/WM flags on the tag definition (+ validation)
- [x] Grouping on (C/I/A, RW); read loop skips non-readable tags
- [x] OVERWRITE_ZERO write
- [x] READ_MODIFY_WRITE
- [x] BOOL encoders + failure reporting
- [ ] Tests


modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/EthernetIPCipOdvaPollingProtocolAdapter.java
```diff
@@ -250,6 +264,11 @@ public void poll(final @NotNull BatchPollingInput pollingInput, final @NotNull B
                     return;
```
caoccao: The whole try block smells bad. This early return looks weird.

caoccao (changes requested):
Something doesn't seem right.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/EthernetIPCipOdvaPollingProtocolAdapter.java
mschoenert-hivemq: Significantly rewritten.
Thanks for pointing me here.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/EthernetIPCipOdvaPollingProtocolAdapter.java
```diff
@@ -232,43 +248,41 @@ public void poll(final @NotNull BatchPollingInput pollingInput, final @NotNull B
...
-            if (errors.isEmpty()) {
-                pollingOutput.finish();
-            } else {
-                String errorString = String.join(",", errors);
+        }
 
-                LOG.warn("Adapter '{}'. {}", adapterId, errorString);
-                pollingOutput.fail(errorString);
-            }
+        reportOutcome(pollingOutput, errors);
```
caoccao, line 276: This code structure is clear now.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/EthernetIPCipOdvaPollingProtocolAdapter.java
```diff
@@ -232,43 +248,41 @@ public void poll(final @NotNull BatchPollingInput pollingInput, final @NotNull B
...
-            // FIXME: Can easily add support for "read tag" as well, with same Decoders/Encoders
-
-            // Create Requests, set handlers
-            for (TagGroup tagGroup : tagGroups.getTagGroups()) {
-                if (isAdapterNotStarted()) {
-                    errors.add(getNotStartedMessage());
-                    return;
+            for (final TagGroup tagGroup : tagGroups.getTagGroups()) {
+                // Write-only tags have no readable attribute; never poll them.
+                if (!tagGroup.isReadable()) {
```
caoccao: It would be better to use `stream().filter()` to simplify this for loop.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/EthernetIPCipOdvaPollingProtocolAdapter.java
```diff
@@ -410,48 +424,277 @@ public void createTagSchema(
...
+            if (node == null) {
+                throw new IllegalArgumentException("No value supplied for tag '" + t.getName() + "'.");
+            }
+            return jsonToCipValue(t, node);
+        };
+
+        // Serialize the read->write window of a PARTIAL_WRITE per attribute so it is not interleaved with another
+        // write to the same attribute. COMPLETE_WRITE does not read, but sharing the lock is harmless and keeps
+        // concurrent writes to one attribute ordered.
+        final Object lock = writeLocks.computeIfAbsent(tagGroup.getTagAddress(), a -> new Object());
```
caoccao: This could be a memory leak point. The lock objects are never removed until the adapter is closed. Also I doubt the device can handle multiple writes simultaneously.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/EthernetIPCipOdvaPollingProtocolAdapter.java
```diff
@@ -410,48 +424,277 @@ public void createTagSchema(
...
+        synchronized (lock) {
```
caoccao: The scope of the lock is too large. It should only guard `readAttributeBytes()`.

caoccao (changes requested):
I have a few new concerns.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/EthernetIPCipOdvaPollingProtocolAdapter.java
caoccao: Codex gave us a more insightful review. My previous comment is not accurate enough. It should guard both read and write.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/EthernetIPCipOdvaPollingProtocolAdapter.java
mschoenert-hivemq: DONE

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/EthernetIPCipOdvaPollingProtocolAdapter.java
mschoenert-hivemq: Fixed. Replaced the per-address lock map with a single adapter-level writeLock, so there are no accumulating lock objects. And you're right that the device can't handle simultaneous writes — the bundled client already serializes all connection I/O (etherip.protocol.Connection.execute is synchronized), so per-attribute parallelism was never actually available; one lock is the honest model. (Note: it still wraps the whole read-modify-write window, per your later correction on the sibling comment.)

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/EthernetIPCipOdvaPollingProtocolAdapter.java
mschoenert-hivemq: Agreed — the lock guards both the read and the write, so a read-modify-write's read→write window can't interleave with another write. Narrowing it to just readAttributeBytes() would let two concurrent partial writes read the same old bytes and lose one update. Left the scope as-is.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/config/EipSpecificAdapterConfig.java
```diff
@@ -141,9 +112,7 @@ public boolean equals(final Object o) {
         if (!(o instanceof final EipSpecificAdapterConfig that)) {
             return false;
         }
-        return port == that.port
-                && backplane == that.backplane
-                && slot == that.slot
+        return slot == that.slot
                 && com.google.common.base.Objects.equal(id, that.id)
```
caoccao: `com.google.common.base.Objects` shouldn't be referenced. We usually use `java.util.Objects`. There are 7 occurrences.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
```diff
@@ -37,28 +56,82 @@ private TagGroup registerTag(CipTag cipTag) throws OdvaException {
...
         }
 
-        for (CipTag cipTag : cipTags) {
-            registerTag(cipTag);
+        // Build and validate first, and only mark registration complete once it has fully succeeded. If any
+        // step throws, discard the partial group state and leave the flag unset, so a later start() retries
+        // registration from scratch instead of continuing on stale or half-built groups.
+        try {
+            for (CipTag cipTag : cipTags) {
+                registerTag(cipTag);
```
caoccao, line 68: `registerTag()` is not good enough. Line 2 should be `final @NotNull TagGroup tagGroup`.

```java
    private TagGroup registerTag(CipTag cipTag) throws OdvaException {
        TagGroup tagGroup = getOrCreateTagGroup(cipTag.getDefinition());
        tagGroup.add(cipTag);
        return tagGroup;
    }
```

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
```diff
@@ -37,28 +56,82 @@ private TagGroup registerTag(CipTag cipTag) throws OdvaException {
...
+        }
     }
 
     @NotNull
     private TagGroup getOrCreateTagGroup(CipTagDefinition definition) throws OdvaException {
         AtomicReference<OdvaException> maybeException = new AtomicReference<>();
 
-        TagGroup tagGroup = tagAddressToTagGroup.computeIfAbsent(definition.getAddress(), address -> {
+        final GroupKey key = new GroupKey(definition.getAddress(), definition.getReadWrite());
+        TagGroup tagGroup = tagAddressToTagGroup.computeIfAbsent(key, k -> {
```
caoccao: It should be `final @Nullable TagGroup tagGroup`. The quality of the whole method is not good.

caoccao (changes requested):
A few new concerns.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/config/EipSpecificAdapterConfig.java
mschoenert-hivemq: Done. Switched to `java.util.Objects (equals/hash)` here, and also in `CipTagDefinition` and `TagGroup`, which had the same Guava usage — so the module is consistent.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
mschoenert-hivemq: Done — `final @NotNull` on the local, and on the parameter and return type too.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
mschoenert-hivemq: Agreed. Rewrote it as a plain get-then-create: `TagGroup`'s constructor is checked-throwing, so the `computeIfAbsent` + `AtomicReference` was only there to smuggle the exception out of the lambda. It now does a get, returns if present, otherwise constructs, puts, and returns — letting `OdvaException` propagate directly. The `@Nullable` local reads honestly now.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
```diff
@@ -15,61 +15,126 @@
...
-            throw maybeException.get();
         }
+    }
 
-        return tagGroup;
+    private @NotNull TagGroup getOrCreateTagGroup(final @NotNull CipTagDefinition definition) throws OdvaException {
+        final GroupKey key = new GroupKey(definition.getAddress(), definition.getReadWrite());
+        // Plain get-then-create: TagGroup's constructor is checked-throwing, so don't smuggle the exception out
+        // of a computeIfAbsent lambda — just create it directly and let OdvaException propagate.
+        final @Nullable TagGroup existing = tagAddressToTagGroup.get(key);
```
caoccao, line 131: It was thread-safe. Now it's not thread-safe. Have you confirmed the whole operation is thread-safe?

caoccao (approved):
Just one question, but approved.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
mschoenert-hivemq: Good question. It's still fine, and it wasn't relying on `computeIfAbsent` for thread-safety in the first place: the backing map is a plain `TreeMap`, not a concurrent map, so `computeIfAbsent` on it is no more thread-safe than a get-then-put — concurrent mutation of a `TreeMap` is unsafe either way. The reason it's safe is that registration is single-threaded: `registerTagsIfEmpty` runs once from `start()`, which the framework invokes serially (the `tagsRegistered` gate makes it idempotent across restarts, not concurrent). So the whole operation is thread-confined to start, and the rewrite doesn't change that.

PR #1591, opened 2026-06-30
EDG-759: Send hmq-bridge user property on bridge CONNECT

mschoenert-hivemq:
## Description (the why)
Edge bridge clients now send the `hmq-bridge=edge` MQTT 5 user property on CONNECT. A HiveMQ broker keys off the property name to assign a larger receive buffer to these connections and to skip shrinking it under overload protection — the differentiated receive buffer from [BCD-515](https://linear.app/hivemq/issue/BCD-515). This is the Edge-side counterpart, [EDG-759](https://linear.app/hivemq/issue/EDG-759). Always-on; no config field, schema, or frontend.

- [ ] Edge sends `hmq-bridge` = `edge` as a CONNECT user property on every bridge client.
- [ ] The marker is present on the initial connect and on every reconnect.
- [ ] No new configuration surface (field/XML schema/frontend).

## Non-Goals
- Edge does not read or act on the property; only the broker consumes it.
- No change to `hmq-bridge-token` (the broker does not consume it).
- The broker-side pin-vs-autotune buffer design is out of scope.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/bridge/mqtt/BridgeMqttClient.java
+++ b/hivemq-edge/src/main/java/com/hivemq/bridge/mqtt/BridgeMqttClient.java
@@ -15,6 +15,8 @@
  */
 package com.hivemq.bridge.mqtt;
 
+import static com.hivemq.edge.HiveMQEdgeConstants.BRIDGE_MARKER_PROPERTY;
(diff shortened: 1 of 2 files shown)


caoccao (approved):
LGTM

Remit (approved):

PR #1592, opened 2026-06-30
EDG-763: Nevsky v2 review round 1 — tags-only reload, per-tag poll, lifecycle shutdown & config-schema validation

caoccao:
## Summary

Round-1 code-review remediation for the Nevsky v2 protocol-adapter framework. All five review findings were
assessed against the authoritative design/plan docs, **all were valid, and all are fixed**. The legacy framework is
untouched (no behavioral change outside the v2 subsystem).

### BUG-001 (High) — tags-only reload rebuilt all tag runtimes
`TagAspectRuntimeCoordinator.updateTagSet` tore down and rebuilt every tag runtime on a `TAGS_ONLY` transition, and
the rebuilt aspects started disconnected — so after a reload of a *connected* adapter every surviving tag stopped
polling (parked in `WAITING_FOR_ADAPTER_READY`) until the next reconnect, which a tags-only change never triggers.

**Fix:** a real in-place diff — a tag unchanged in node-string, access mode, and poll interval keeps its running
`TagRuntime` and keeps polling untouched; a removed tag is deactivated; an added tag is verified against the live
connection by replaying the coordinator's tracked adapter-readiness phase. Reuses the aspect machine's existing
runtime-add path; no FSM change.

### BUG-002 (High) — per-tag poll intervals collapsed to the shortest
The coordinator polled every tag at one adapter-wide shortest interval, overpolling slow tags. The per-tag
`poll-interval-millis` is now carried end to end (`ProtocolAdapterConfigSupport.pollIntervalMillisByTagName` →
the `UpdateTagSet` command → the coordinator → each `TagRuntime`).

### BUG-003 (High) — lifecycle stop leaked live wrappers
`ProtocolAdapterLifecycle.stop()` closed only the manager tick/dispatcher/clock and left every wrapper container
(dispatcher binding, tick, metrics) leaked and the adapters unstopped. It now drives a graceful, actor-model
shutdown — a latch-bearing `ShutdownRequested` (CONTROL) message processed on the manager dispatch thread — that
stops every adapter and closes every container before the runtime is torn down.

### BUG-004 (Medium) — adapter config not validated against the schema
A new `AdapterConfigurationSchemaValidator` validates the `<adapter-configuration>` against the owning type's
`adapterConfigSchema()` in the wrapper-factory create path (the only place the factory is available — the `<v2>`
extractor runs before the module-loader-discovered factory registry exists), surfaced as the existing clear
per-adapter `ERROR` handle. The chaos adapter's config schema was corrected from a mismatched `STRING` scalar to
`AnySchema`.

### GUIDELINE-001 (Medium) — internal labels in source comments
Removed internal design/scenario labels (`D2`, `D6`, `D7`, `D8`, `S24`, `S32`) and the `PAM` abbreviation from
shipped source comments, replacing them with behavior descriptions (rules N2/N4). A repo-wide re-scan of the
v2/SDK-v2/chaos main source is clean.

## Tests
- chaos scenario matrix: strengthened S14 (survivor / added-tag) + a per-tag poll-interval test
- core: `ProtocolAdapterManagerTest` shutdown tests, `AdapterConfigurationSchemaValidatorTest`,
  `DefaultProtocolAdapterWrapperFactoryTest` schema-rejection cases, `TagSetReloadAndPollIntervalTest`

## Verification
`spotlessApply` clean; `:hivemq-edge-build:hivemq-edge:check` (ErrorProne + NullAway + Spotless + full core unit
suite) green; the chaos module check green; the full `com.hivemq.edge.api.v2.*` wired end-to-end suite (in
hivemq-edge-test) green.

## Companion PRs
- `hivemq-edge-adapter-sdk` — Javadoc (GUIDELINE-001)
- `hivemq-edge-test` — wired end-to-end tests for BUG-001 / BUG-002


marregui:
**Recovery gap: a schema-rejected adapter can't be fixed without a restart**

The new schema preflight in `createAdapter` (L293–296) correctly parks a config-invalid adapter as a non-real ERROR container via `registerErrorAdapter`. The problem is the way back out — as far as I can tell there isn't one short of a process restart.

On reload, `applyDifference` takes the `!existing.isReal()` path (L199) and only rebuilds when the *protocol-id* changes:

```java
if (running.getProtocolId().equals(updated.getProtocolId())) {
    existing.appliedEntity(updated);   // records the new config, but never rebuilds
} else {
    stopAndDiscard(adapterId, updated); // rebuild
}
```

A corrected config keeps the same protocol-id, so we land on `appliedEntity(updated)`: the fixed config is recorded but no build is re-attempted, and the container is still the non-real one (`NO_OP_WRAPPER_SENDER`), so activate/deactivate are no-ops. The adapter stays ERROR until restart.

Repro:
1. External config for an adapter with a schema-invalid `adapter-configuration` (e.g. missing required `host`). Entity validation passes — schema isn't checked there, it needs the factory — so it reaches the preflight and is parked as ERROR.
2. Fix `host`, reload.
3. Expected: it rebuilds and comes up. Actual: still ERROR; only a restart recovers it.

So it's exactly the configs the new validation is meant to catch that can't self-heal — feels like the wrong side of the tradeoff.

Suggested fix: in the `!isReal` branch, re-attempt the build when the applied config actually changed, not only when the protocol-id changed — i.e. if `updated` differs from `running`, `stopAndDiscard(adapterId, updated)` (or a direct recreate) rather than just recording it.

Flagging this as the one I'd want resolved before merge-down; the rest of my notes are polish. Let me know if I'm misreading the `!isReal` path.

marregui (approved):
LGTM

left a comment https://github.com/hivemq/hivemq-edge/pull/1592#issuecomment-4856939201

marregui:
Addressed in `88e07d1c3`.

The `!isReal()` branch in `applyDifference` now rebuilds on **any** configuration change (`!running.equals(updated)` → `stopAndDiscard`) instead of only on a protocol-id change. So a schema-rejected (or otherwise un-instantiable) adapter is rebuilt from the corrected configuration when the operator fixes it and reloads, rather than staying in ERROR until a process restart.

Covered by a new `ProtocolAdapterManagerTest` regression — `schemaRejectedAdapter_recoversWhenTheConfigurationIsFixedAndReloaded` — which reproduces the stranded-ERROR path (reject on `host=bad`, then reload with `host=good`) and fails under the old protocol-id-only logic. Compile + the manager test class + spotless all green locally.

PR #1603, opened 2026-07-07
EDG-791: Prevent errorprone log listener from wedging the build

mschoenert-hivemq:
The addStandardErrorListener callback appended to the errorprone report file assuming its parent directory existed. When the listener fires while doFirst is skipped (UP-TO-DATE / FROM-CACHE), the directory is absent and appendText throws FileNotFoundException on a Gradle worker thread, which wedges the whole build. Create the parent dir at append time and swallow I/O errors so logging can never fail the build.

**Motivation**

Resolves #<issue>

**Changes**


estefania-hivemq (approved):
LGTM

PR #1604, opened 2026-07-07
EDG-803: Validate tag byte-offset layout in the EtherNet/IP – Logical Addressing adapter

mschoenert-hivemq:
## Description (the why)
Tag byte offsets in the EtherNet/IP – Logical Addressing adapter were accepted without any validation, so nonsensical layouts passed silently. The dangerous case is two tags at the **same attribute and direction** claiming **overlapping bytes**: on write they corrupt each other (last-writer-wins, no warning); on read they decode the same bytes into different fields. Found in QA by Estefania Amundaray (finding 2 on EDG-777).

Linear: EDG-803.

- [x] Per-tag validation at registration: `batchByteIndex >= 0`, `numberOfElements >= 1`, `batchBitIndex` in 0–7.
- [x] Per `(address, direction)` group: reject tags whose byte ranges `[batchByteIndex, batchByteIndex + width * numberOfElements)` overlap.
- [x] Byte width from a single source of truth (`CipDataType.staticByteWidth()`, mirroring the encoders).
- [x] Strings (no static width) and bit-addressed BOOLs (share a byte at different bits) are correctly **not** treated as overlaps.
- [x] Tags at different attributes or different directions are never compared (they are independent operations).
- [x] Unit tests for each reject case and the legitimate non-overlap cases; all EIP integration tests stay green.

## Non-Goals
- Validating a byte offset **past the end of the attribute**: a generic CIP device does not expose an attribute's size, so this is only detectable at read/write time (handled there), not at config time.
- Findings 1 and 3 from the QA pass (separate issues: EDG-791 packaging, and runtime-vs-startup validation consistency).

### DIFF
--- a/modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/config/CipDataType.java
+++ b/modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/config/CipDataType.java
@@ -15,6 +15,8 @@
  */
 package com.hivemq.edge.adapters.etherip_cip_odva.config;
 
+import org.jetbrains.annotations.NotNull;
+
 public enum CipDataType {
     BOOL,
     SINT, // 8 bit signed
@@ -34,5 +36,21 @@ public enum CipDataType {
     // DWORD // 32 bit string
     // LWORD // 64 bit string
 
-    COMPOSITE // artificial type to indicate creation of a composite of all tags at given tag address
+    COMPOSITE; // artificial type to indicate creation of a composite of all tags at given tag address
+
+    /**
+     * The fixed number of bytes one element of this type occupies on the wire, or empty if the width is not
+     * statically known. Strings ({@link #SSTRING}/{@link #STRING}) are variable-length (a length prefix plus the
+     * character bytes), and {@link #COMPOSITE} is not a wire type at all, so those return empty. The widths mirror
+     * the encoders (see {@code CipTagEncoders}); this is the single source of truth for byte-layout validation.
+     */
+    public @NotNull java.util.OptionalInt staticByteWidth() {
+        return switch (this) {
+            case BOOL, SINT, USINT -> java.util.OptionalInt.of(1);
+            case INT, UINT -> java.util.OptionalInt.of(2);
+            case DINT, UDINT, REAL -> java.util.OptionalInt.of(4);
+            case LINT, LREAL -> java.util.OptionalInt.of(8);
+            case SSTRING, STRING, COMPOSITE -> java.util.OptionalInt.empty();
+        };
+    }
 }
--- a/modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/config/CipDataType.java
+++ b/modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/config/CipDataType.java
@@ -15,6 +15,7 @@
  */
 package com.hivemq.edge.adapters.etherip_cip_odva.config;
 
+import java.util.OptionalInt;
 import org.jetbrains.annotations.NotNull;
 
 public enum CipDataType {
@@ -44,13 +45,13 @@ public enum CipDataType {
      * character bytes), and {@link #COMPOSITE} is not a wire type at all, so those return empty. The widths mirror
      * the encoders (see {@code CipTagEncoders}); this is the single source of truth for byte-layout validation.
      */
-    public @NotNull java.util.OptionalInt staticByteWidth() {
+    public @NotNull OptionalInt staticByteWidth() {
         return switch (this) {
-            case BOOL, SINT, USINT -> java.util.OptionalInt.of(1);
-            case INT, UINT -> java.util.OptionalInt.of(2);
-            case DINT, UDINT, REAL -> java.util.OptionalInt.of(4);
-            case LINT, LREAL -> java.util.OptionalInt.of(8);
-            case SSTRING, STRING, COMPOSITE -> java.util.OptionalInt.empty();
+            case BOOL, SINT, USINT -> OptionalInt.of(1);
+            case INT, UINT -> OptionalInt.of(2);
+            case DINT, UDINT, REAL -> OptionalInt.of(4);
+            case LINT, LREAL -> OptionalInt.of(8);
+            case SSTRING, STRING, COMPOSITE -> OptionalInt.empty();
         };
     }
 }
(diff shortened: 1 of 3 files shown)


modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/config/CipDataType.java
```diff
@@ -34,5 +36,21 @@ public enum CipDataType {
...
-    COMPOSITE // artificial type to indicate creation of a composite of all tags at given tag address
+    COMPOSITE; // artificial type to indicate creation of a composite of all tags at given tag address
+
+    /**
+     * The fixed number of bytes one element of this type occupies on the wire, or empty if the width is not
+     * statically known. Strings ({@link #SSTRING}/{@link #STRING}) are variable-length (a length prefix plus the
+     * character bytes), and {@link #COMPOSITE} is not a wire type at all, so those return empty. The widths mirror
+     * the encoders (see {@code CipTagEncoders}); this is the single source of truth for byte-layout validation.
+     */
+    public @NotNull java.util.OptionalInt staticByteWidth() {
```
caoccao: We don't use QFN here. Import `java.util.OptionalInt` on the top.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
```diff
@@ -65,10 +67,12 @@ public boolean registerTagsIfEmpty(Collection<CipTag> cipTags) throws OdvaExcept
         // registration from scratch instead of continuing on stale or half-built groups.
         try {
             for (CipTag cipTag : cipTags) {
+                validateTagDefinition(cipTag);
```
caoccao, line 70: If one tag definition validation failure can cause the whole for-loop fail, calling `registerTag()` would be a waste. The code is OK, just not optimized enough.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
```diff
@@ -124,6 +128,102 @@ private void validateCompositesHaveSiblings() throws OdvaException {
...
+     * <ul>
+     *   <li><b>Strings</b> ({@code SSTRING}/{@code STRING}) have no static width — their length is only known at
+     *       write/read time — so their span cannot be checked up front and they are skipped here.</li>
+     *   <li><b>Bit-addressed {@code BOOL}s</b> (a {@code BOOL} with a {@code batchBitIndex}) occupy a single bit,
+     *       so several may legitimately share one byte at different bit positions; they are not byte-overlaps.</li>
+     * </ul>
+     * The composite tag itself carries no bytes (it aggregates its siblings), so it is not considered here.
+     */
+    private void validateNoByteRangeOverlap() throws OdvaException {
+        for (final TagGroup tagGroup : tagAddressToTagGroup.values()) {
```
caoccao, line 181: The performance of this function is pool. It is an O(N^2) algorithm.
I think it can be for sure optimized. E.g. binary search tree.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
```diff
@@ -124,6 +128,102 @@ private void validateCompositesHaveSiblings() throws OdvaException {
...
+    /** A tag occupies a checkable byte range if its type has a static width and it is not a bit-addressed BOOL. */
+    private static boolean occupiesAStaticByteRange(final @NotNull CipTag cipTag) {
+        final CipTagDefinition definition = cipTag.getDefinition();
+        final boolean bitAddressedBool =
+                definition.getDataType() == CipDataType.BOOL && definition.getBatchBitIndex() != null;
+        return definition.getDataType().staticByteWidth().isPresent() && !bitAddressedBool;
+    }
+
+    private static boolean byteRangesOverlap(final @NotNull CipTag a, final @NotNull CipTag b) {
+        final int aStart = a.getDefinition().getBatchByteIndex();
```
caoccao, line 216: It assumes `a` is always ahead of `b`. Does that mean `List<CipTag> byteRangedTags` is an ordered list?

caoccao (changes requested):
I hold some concerns.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
mschoenert-hivemq: It does not assume that. Actually it is easier to see if you negate it. The two ranges DONT overlap if either b is completely behind a: `aEnd <= bStart` or a is completely behind a: `bEnd <= aStart`. So they do overlap iff `! ((aEnd <= bStart) || (bEnd <= aStart))`. The rest is boolean re-arrangement.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
mschoenert-hivemq: Sure. But N will be so small that we don't care.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
mschoenert-hivemq: I want to do the minimal amount - to avoid that another problem sneaks in. And I'm not worried about performance at all.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/config/CipDataType.java
mschoenert-hivemq: Will DO

mschoenert-hivemq:
Thank you. Addressed.

modules/hivemq-edge-module-etherip-cip-odva/src/main/java/com/hivemq/edge/adapters/etherip_cip_odva/tag/TagGroups.java
caoccao: OK

caoccao (approved):

PR #1596, opened 2026-07-03
fix(EDG-784,EDG-785): browse RESOLVE completeness + abort resilient to a throwing browseCancel

marregui:
Fixes EDG-784 and EDG-785 from the round-2 review (EDG-779). `hivemq-edge` only.

## EDG-784 — RESOLVE completeness
The browse engine accepted any `readNodeAttributes` result and dropped discovered selectable nodes that had no resolved attributes in `finish()`, returning an incomplete browse as `200 OK` (short, silent). It now tracks the requested node ids per resolve batch and **fails the browse** on a missing, duplicate, or unrequested attribute instead of dropping.

## EDG-785 — abort resilient to a throwing `browseCancel`
`ProtocolAdapterBrowseEngine.abort()` called `browseCancel()` unguarded before `reset()`. A direct adapter throwing from `browseCancel` skipped the reset — wedging the in-flight slot into a permanent `409 ALREADY_IN_FLIGHT` — and skipped the caller's rejection. `browseCancel` is now best-effort with `reset()` in a `finally`; the wrapper completes the REST future in a `finally` on the deadline and connection-loss paths (defence in depth).

## Tests
- Engine: missing / unrequested resolve → browse fails; a complete resolve still succeeds; a throwing `browseCancel` no longer propagates from `abort()` and the slot is released.
- Wrapper: deadline and connection-loss with a throwing `browseCancel` still resolve `TIMED_OUT` / `NOT_CONNECTED`, leave the engine reusable, and keep the transition intact.

Verified locally: compile clean, `ProtocolAdapterWrapperBrowseTest` 7/7, `ProtocolAdapterBrowseEngineTest` 10/10, no regressions in `v2.wrapper.*` (56) or opcua `conformance.*` (18).

PR #1605, opened 2026-07-07
fix(EDG-784,EDG-785): browse RESOLVE completeness + abort resilient to a throwing browseCancel

marregui:
Two related High-severity browse bugs from [EDG-779](https://linear.app/hivemq/issue/EDG-779) round-2 review (Sam Cao), on the Nevsky v2 browse engine.

## EDG-784 — RESOLVE silently drops discovered nodes when the result is incomplete

The RESOLVE phase never checked that the adapter returned one `ResolvedAttributes` per requested node. `onAttributesResolved` stored whatever came back and `finish()` emitted only the discovered nodes that happened to have attributes — silently omitting a **selectable** node whose attributes were missing, and completing the browse as `200 OK` short a tag. This bites a well-behaved adapter (e.g. an OPC-UA node that browses selectable but denies the attribute read), not only a buggy one.

**Fix:** track the requested node ids per RESOLVE batch; fail the whole browse on a **missing / duplicate / unrequested** attribute instead of dropping. The failure routes through a shared `failBrowse` helper (also used by `onBrowseError`). SDK javadoc (`readNodeAttributes` / `readAttributesResult`) documents the now-enforced "exactly one attribute per requested node" contract — those changes land on the matching `ma/follow-up-v2-review` branch of `hivemq-edge-adapter-sdk` (separate repo, picked up by the composite build).

## EDG-785 — `abort()` ran `browseCancel()` unguarded; a throwing adapter stranded the REST future and wedged the engine

`ProtocolAdapterBrowseEngine.abort()` ran `browseCancel()` before `reset()`, unguarded. A raw adapter throwing from `browseCancel` skipped `reset()` (slot never released → permanent `409 ALREADY_IN_FLIGHT`), skipped the caller's rejection, and — from `failPendingBrowse` — escaped into the wrapper FSM/tick loop.

**Fix:** `browseCancel()` is best-effort (catch + log) and `reset()` runs in a `finally`; the deadline and connection-loss paths complete the pending REST future in a `finally` around `abort()`, so the caller is always released with the intended `TIMED_OUT` / `NOT_CONNECTED` rejection.

## Tests

- **Engine unit** (`ProtocolAdapterBrowseEngineTest`): complete / missing / duplicate / unrequested RESOLVE batches; throwing `browseCancel` in `abort()` does not propagate and the slot is reusable.
- **Wrapper** (`ProtocolAdapterWrapperBrowseTest`): incomplete resolve fails (`FAILED`) instead of a short `200` and releases the slot; deadline and connection-loss with a throwing `browseCancel` still resolve `TIMED_OUT` / `NOT_CONNECTED` without wedging the FSM.

All pass; spotless applied. Supersedes the earlier bundled PR #1596.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/protocols/v2/browse/ProtocolAdapterBrowseEngine.java
+++ b/hivemq-edge/src/main/java/com/hivemq/protocols/v2/browse/ProtocolAdapterBrowseEngine.java
@@ -106,6 +106,7 @@ public final class ProtocolAdapterBrowseEngine {
     private final @NotNull Map<String, String> pathByNode = new HashMap<>();
     private final @NotNull Map<String, String> tagNameByNode = new HashMap<>();
     private final @NotNull Map<String, ResolvedAttributes> resolvedByNode = new HashMap<>();
+    private final @NotNull List<String> pendingResolveBatch = new ArrayList<>();
     private int resolveOffset;
 
     /**
@@ -169,6 +170,7 @@ public final class ProtocolAdapterBrowseEngine {
         pathByNode.clear();
         tagNameByNode.clear();
         resolvedByNode.clear();
+        pendingResolveBatch.clear();
         resolveOffset = 0;
         activePath = "";
         activeContinuation = null;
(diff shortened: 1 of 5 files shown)


caoccao (approved):
LGTM

PR #1606, opened 2026-07-07
EDG-788: Nevsky v2 — file-v2 and http-v2 production adapter modules

caoccao:
## Summary

Ports the **File** and **HTTP** protocol adapters to **SDK v2** (Project Nevsky) as new, parallel, **production-listable** modules, running side by side with the untouched v1 modules. Distinct protocol-ids (`file-v2` / `http-v2`) and Java packages (`…adapters.file.v2` / `…adapters.http.v2`); disjoint config sections, factory registries, and REST surfaces — the two never interact.

### `hivemq-edge-module-file-v2`
- Stateless northbound poll-only reader on `AbstractProtocolAdapter` (File has no connection, so `doConnect` acknowledges immediately).
- The five content types (`BINARY`, `TEXT_PLAIN`, `TEXT_JSON`, `TEXT_XML`, `TEXT_CSV`) and the **64 KB** size cap are ported verbatim.
- Empty `capabilities()` — no write, browse, or subscription.

### `hivemq-edge-module-http-v2`
- Owns the JDK `java.net.http.HttpClient` across `doConnect`/`doDisconnect` (optional trust-all TLS gated behind `allowUntrustedCertificates`, default off) — no third-party HTTP dependency.
- `doPoll` fires `sendAsync` and reports through the thread-safe `output` façade from the client's completion thread, **preserving v1's per-cycle parallelism**; the per-request timeout bounds every in-flight poll.
- GET/POST/PUT, custom headers, per-tag body/timeout, `HiveMQ-Edge; <version>` User-Agent, JSON/Base64 response decoding, and success-status filtering carried over.

### Build / distribution wiring
- `settings.gradle.kts` `includeBuild` for both modules.
- `build.gradle.kts` `edgeModule(...)` + `edgeProjectsToUpdate` entries so both ship in the distribution and are module-loader discovered as listable v2 types.
- Each module mirrors the v1 module build (shadow jar + third-party-license consumable) so it slots into `hivemqEdgeZip`/`hivemqEdgeFullZip`.

## Parity deltas (documented per-module README)
1. **No auto-removal** after repeated poll errors — the framework counts failures with escalating log severity instead.
2. The **MQTT payload envelope** belongs to the framework's northbound mappings.
3. The **poll interval is per-tag** (`TagEntity.pollIntervalMillis`).
4. HTTP: the dormant v1 southbound write path is **excluded** under strict parity (documented extension point).

## Tests
- Module `check` **green** for both (`file-v2`: 7 test files; `http-v2`: 8, including an in-process JDK `HttpServer` + Awaitility for the async/concurrency/timeout paths). ErrorProne/NullAway clean, Spotless clean.
- Wired end-to-end coverage lives in the companion `hivemq-edge-test` PR.

## Related PRs
- `hivemq-edge-test`: wired e2e suites + module-loader binary wiring.
- `hivemq-edge-composite`: composite distribution wiring.


modules/hivemq-edge-module-file-v2/src/main/java/com/hivemq/edge/adapters/file/v2/FileProtocolAdapter.java
```diff
@@ -0,0 +1,160 @@
...
+        output.disconnected();
+    }
+
+    @Override
+    protected void doPoll(final @NotNull Node node) {
+        if (!(node instanceof final FileNode fileNode)) {
+            output.nodeError(node, "the file adapter received a node of an unexpected type", false);
+            return;
+        }
+        final Path path = Path.of(fileNode.filePath());
```
marregui: if by any hickup `fileNode.filePath()` were null, then the whole adapter would be in ERROR state, intended or should it simply fail the tag, log the error and carry on?

modules/hivemq-edge-module-file-v2/src/main/java/com/hivemq/edge/adapters/file/v2/FileProtocolAdapter.java
```diff
@@ -0,0 +1,160 @@
...
+                                + path.toAbsolutePath()
+                                + "' of size '"
+                                + length
+                                + "' exceeds the limit '"
+                                + MAX_FILE_SIZE_BYTES
+                                + "'.",
+                        false);
+                return;
+            }
+            final byte[] fileContent = Files.readAllBytes(path);
```
marregui: a file that grew in the meantime (since last call to Files.size(path)) may silently be longer than the max limit.

modules/hivemq-edge-module-file-v2/src/main/java/com/hivemq/edge/adapters/file/v2/FileNode.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @param filePath    the absolute path to the file to read.
+     * @param contentType the type of the content within the file.
+     */
+    @JsonCreator
+    public FileNode(
+            @JsonProperty(value = "filePath", required = true) final @NotNull String filePath,
+            @JsonProperty(value = "contentType", required = true) final @NotNull FileContentType contentType) {
+        this.filePath = filePath;
```
marregui: null checks for required fields. Also, those that map to enum, can be validated here.

modules/hivemq-edge-module-file-v2/src/main/java/com/hivemq/edge/adapters/file/v2/FileProtocolAdapterFactory.java
```diff
@@ -0,0 +1,104 @@
...
+                        false));
+        // The reused v1 Schema has no enum constraint, so the content type projects as a plain string; the accepted
+        // values are the names of the FileContentType constants.
+        properties.put(
+                "contentType",
+                new ScalarSchema(
+                        ScalarType.STRING,
+                        null,
+                        null,
+                        "Content Type",
```
marregui, line 88: I have seen an enum somewhere

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpAdapterConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+        return new HttpAdapterConfiguration(
+                Math.min(connectTimeout, MAX_TIMEOUT_SECONDS),
+                boolField(node, "allowUntrustedCertificates", false),
+                boolField(node, "assertResponseIsJson", false),
+                boolField(node, "httpPublishSuccessStatusCodeOnly", true));
+    }
+
+    private static int intField(final @NotNull JsonNode node, final @NotNull String field, final int defaultValue) {
+        final JsonNode value = node.get(field);
+        return (value != null && value.isNumber()) ? value.asInt() : defaultValue;
```
marregui: the value might be a TextNode("14") ... will `isNumber()` return true or will it return false because it is text?. The adapter config has no schema so what you get are text nodes, strings.

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpAdapterConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+     * connect timeout to a sane ceiling. Tolerant of an absent, empty, or additionally-populated configuration.
+     *
+     * @param adapterConfig the reused v1 configuration value handed to the adapter.
+     * @param objectMapper  the mapper used to read the configuration map.
+     * @return the parsed configuration.
+     */
+    public static @NotNull HttpAdapterConfiguration parse(
+            final @NotNull DataPoint adapterConfig, final @NotNull ObjectMapper objectMapper) {
+        final JsonNode node = objectMapper.valueToTree(adapterConfig.getTagValue());
+        final int connectTimeout = intField(node, "httpConnectTimeoutSeconds", DEFAULT_TIMEOUT_SECONDS);
```
marregui, line 55: worth checking > 0

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpProtocolAdapter.java
```diff
@@ -0,0 +1,312 @@
...
+            httpClient = builder.build();
+            output.connected();
+        } catch (final Exception e) {
+            output.error(ErrorScope.CONNECTION, "Unable to build the HTTP client: " + e.getMessage());
+        }
+    }
+
+    @Override
+    protected void doDisconnect() {
+        httpClient = null;
```
marregui: would it need to be closed?

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpProtocolAdapter.java
```diff
@@ -0,0 +1,312 @@
...
+        final HttpRequest.Builder builder = HttpRequest.newBuilder()
+                .uri(URI.create(node.url()))
+                .timeout(Duration.ofSeconds(node.httpRequestTimeoutSeconds()))
+                .setHeader(USER_AGENT_HEADER, userAgent);
+        node.httpHeaders().forEach(header -> builder.setHeader(header.name(), header.value()));
+        switch (node.httpRequestMethod()) {
+            case GET -> builder.GET();
+            case POST -> {
+                builder.POST(HttpRequest.BodyPublishers.ofString(bodyOrEmpty(node)));
+                builder.header(
```
marregui: content type will be set twice potentially, as per the setHeader in a loop above

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpProtocolAdapter.java
```diff
@@ -0,0 +1,312 @@
...
+                    .version(HttpClient.Version.HTTP_1_1)
+                    .followRedirects(HttpClient.Redirect.NORMAL)
+                    .connectTimeout(Duration.ofSeconds(configuration.httpConnectTimeoutSeconds()));
+            if (configuration.allowUntrustedCertificates()) {
+                builder.sslContext(createTrustAllContext());
+            }
+            httpClient = builder.build();
+            output.connected();
+        } catch (final Exception e) {
+            output.error(ErrorScope.CONNECTION, "Unable to build the HTTP client: " + e.getMessage());
```
marregui, line 113: on failure, should the client be nulled away?

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpProtocolAdapter.java
```diff
@@ -0,0 +1,312 @@
...
+        // what preserves v1's per-cycle parallelism, and the dispatch thread does not block on it.
+        final var unused = client.sendAsync(request, HttpResponse.BodyHandlers.ofString())
+                .whenComplete((response, throwable) -> {
+                    if (throwable != null) {
+                        output.nodeError(
+                                node,
+                                "the HTTP request to '" + httpNode.url() + "' failed: " + describe(throwable),
+                                false);
+                    } else {
+                        handleResponse(httpNode, node, response);
```
marregui: might leak an unchecked exception

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpProtocolAdapter.java
```diff
@@ -0,0 +1,312 @@
...
+    private static @NotNull String bodyOrEmpty(final @NotNull HttpNode node) {
+        final String body = node.httpRequestBody();
+        return body != null ? body : "";
+    }
+
+    private void handleResponse(
+            final @NotNull HttpNode node, final @NotNull Node key, final @NotNull HttpResponse<String> response) {
+        final int status = response.statusCode();
+        final boolean success = status >= 200 && status <= 299;
+        if (!success && configuration.httpPublishSuccessStatusCodeOnly()) {
```
marregui, line 201: the second check should be a negative check

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpNode.java
```diff
@@ -0,0 +1,173 @@
...
+     */
+    @JsonCreator
+    public HttpNode(
+            @JsonProperty(value = "url", required = true) final @NotNull String url,
+            @JsonProperty("httpRequestMethod") final @Nullable HttpMethod httpRequestMethod,
+            @JsonProperty("httpRequestTimeoutSeconds") final @Nullable Integer httpRequestTimeoutSeconds,
+            @JsonProperty("httpRequestBodyContentType") final @Nullable HttpContentType httpRequestBodyContentType,
+            @JsonProperty("httpRequestBody") final @Nullable String httpRequestBody,
+            @JsonProperty("httpHeaders") final @Nullable List<HttpHeader> httpHeaders) {
+        this.url = url;
```
marregui: check null

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpNode.java
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * @return the custom request headers.
+     */
+    public @NotNull List<HttpHeader> httpHeaders() {
+        return httpHeaders;
+    }
+
+    @Override
+    public @NotNull String nodeId() {
+        return url;
```
marregui: two tags looking at the same url with different method/body will make it not UNIQUE

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpNode.java
```diff
@@ -0,0 +1,173 @@
...
+            @JsonProperty("httpRequestBodyContentType") final @Nullable HttpContentType httpRequestBodyContentType,
+            @JsonProperty("httpRequestBody") final @Nullable String httpRequestBody,
+            @JsonProperty("httpHeaders") final @Nullable List<HttpHeader> httpHeaders) {
+        this.url = url;
+        this.httpRequestMethod = Objects.requireNonNullElse(httpRequestMethod, HttpMethod.GET);
+        this.httpRequestBodyContentType = Objects.requireNonNullElse(httpRequestBodyContentType, HttpContentType.JSON);
+        this.httpRequestBody = httpRequestBody;
+        // Ensure a reasonable timeout so a hung endpoint cannot wedge a poll, exactly as the v1 adapter does.
+        this.httpRequestTimeoutSeconds = httpRequestTimeoutSeconds != null
+                ? Math.min(httpRequestTimeoutSeconds, MAX_TIMEOUT_SECONDS)
```
marregui: check the min is still > 0

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpProtocolAdapterInformation.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Override
+    public @NotNull String description() {
+        return "Connects HiveMQ Edge to arbitrary web endpoint URLs via HTTP(s), consuming structured JSON or plain"
+                + " data.";
+    }
+
+    @Override
+    public @NotNull String version() {
+        return "1.0.0";
```
marregui, line 64: do these follow the Edge's version or do they follow their own version?

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpAdapterConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+        return new HttpAdapterConfiguration(
+                Math.min(connectTimeout, MAX_TIMEOUT_SECONDS),
```
marregui: but is there an adapterConfigSchema we can coerce these settings to?

marregui (changes requested):
I found a few nits worth addressing

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpAdapterConfiguration.java
caoccao: I don't think so. Adapters don't see each other. If there is a shared base config, it has to be in the SDK or Edge. I don't think having such entity in SDK is the right approach. And I don't think an adapter should reference Edge.

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpProtocolAdapter.java
caoccao: Addressed in e41ba25e3. The async completion handler now catches RuntimeException from response processing, logs it, and reports a per-node error instead of leaving the future exceptionally completed.

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpAdapterConfiguration.java
caoccao: Addressed in e41ba25e3. Runtime configuration values may be TextNodes, so HttpAdapterConfiguration now parses textual integral and boolean values as well as native JSON scalar values. Added coverage for the string form.

modules/hivemq-edge-module-file-v2/src/main/java/com/hivemq/edge/adapters/file/v2/FileProtocolAdapter.java
caoccao: Addressed in e41ba25e3. FileNode now rejects null required fields, and FileProtocolAdapter converts an InvalidPathException into a per-node error rather than allowing it to escape the poll path.

modules/hivemq-edge-module-file-v2/src/main/java/com/hivemq/edge/adapters/file/v2/FileProtocolAdapter.java
caoccao: Addressed in e41ba25e3. Polling now reads at most 64,001 bytes from an InputStream, so the 64 KB limit is enforced during the actual read even if the file grows after metadata inspection.

modules/hivemq-edge-module-file-v2/src/main/java/com/hivemq/edge/adapters/file/v2/FileProtocolAdapterFactory.java
caoccao: No code change here: the v2 ScalarSchema API available to this module has no enum-values constraint. The schema documents the accepted string names, and Jackson validates them when deserializing FileContentType; this is covered by a regression test.

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpProtocolAdapter.java
caoccao: Addressed in e41ba25e3. POST and PUT now use setHeader for the configured Content-Type, replacing any custom value instead of appending a second header. The HTTP test asserts exactly one received Content-Type value.

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpNode.java
caoccao: Addressed in e41ba25e3. nodeId() now uses the stable full node string, so URL, method, timeout, content type, body, and headers distinguish otherwise same-URL requests. Added regression coverage for method/body/header differences.

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpProtocolAdapterInformation.java
caoccao: This is the adapter semantic version, not the Edge distribution version. The v2 Chaos adapter follows the same 1.0.0 convention; configuration compatibility is tracked separately via currentConfigVersion(). No code change needed.

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpNode.java
caoccao: Addressed in e41ba25e3. Per-request timeout is now clamped to the inclusive range 1..60 seconds, with a regression for zero.

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpAdapterConfiguration.java
caoccao: Addressed in e41ba25e3. The connection timeout is now clamped to the inclusive range 1..60 seconds, with a regression for zero.

modules/hivemq-edge-module-file-v2/src/main/java/com/hivemq/edge/adapters/file/v2/FileNode.java
caoccao: Addressed in e41ba25e3. FileNode now requires non-null filePath and contentType. Jackson already rejects unknown FileContentType values during deserialization; a regression test now documents that behavior.

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpProtocolAdapter.java
caoccao: Addressed in e41ba25e3. The client is now retired on stop, disconnect, and reconnect with shutdownNow(), which cancels in-flight polls without blocking the adapter dispatcher. Callbacks from retired clients are ignored.

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpNode.java
caoccao: Addressed in e41ba25e3. HttpNode now requires a non-null URL, with a focused regression. A malformed non-null URL remains a per-node request-build error.

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpProtocolAdapter.java
caoccao: Addressed in e41ba25e3. doConnect now retires any existing client before building a replacement, so a construction failure cannot leave a previous client available for polling.

modules/hivemq-edge-module-http-v2/src/main/java/com/hivemq/edge/adapters/http/v2/HttpProtocolAdapter.java
caoccao: No code change was needed: the condition is already the negative check, if (!success && configuration.httpPublishSuccessStatusCodeOnly()). Existing tests cover both the error and publish-on-404 branches.

caoccao:
All inline review feedback has been addressed or clarified in e41ba25e3. The File v2 and HTTP v2 module tests plus Spotless checks pass. Each inline thread now has a dedicated response and is resolved.

marregui (approved):
lgtm

PR #1607, opened 2026-07-07
feat(EDG-786): surface resolved browse data (path, tag name, data type, access, description)

marregui:
## EDG-786 — stop discarding resolved browse data at the REST boundary

**Source:** EDG-779 round-2 review (IMP-001, Sam Cao).

The two-phase browse engine assembles rich `BrowsedNode` results — `{entry, path, tagName, attributes}`, where `attributes` carries the resolved `dataType`, `access`, and `description` — but the wrapper converted them back to the bare DISCOVER `BrowseNode` before returning: `completeBrowse` copied only `node.entry()`, and the `BrowseResultEntry` DTO exposed only `nodeId`/`nodeString`/`type`/`selectable`. The entire RESOLVE output (plus the assembled path and suggested default tag name) was dropped at the edge, so the browse UI couldn't pre-fill tag creation.

### Change

- **Carry the resolved result through the boundary:** the browse future and its message chain switch `BrowseNode → BrowsedNode` (`completeBrowse`, `ProtocolAdapterWrapperBrowseRequest`, `ProtocolAdapterManagerMessage.BrowseRequested`, the resource future). `completeBrowse` no longer strips the result.
- **Extend the REST DTO:** `BrowseResultEntry` (OpenAPI) gains `path`, `tagName`, `dataType`, `access`, and `description`. `access` reuses the existing `AccessFlags` schema (readable/writable/pollable/subscribable) rather than a lossy `READ|READ_WRITE|WRITE` string — faithful and consistent with how tags already expose access. The v2 spec bundle (`ext/hivemq-edge-openapi-v2.yaml`) is regenerated.
- `ProtocolAdaptersResourceImpl.toBrowseResult` maps every field; a `toAccessFlags(SDK AccessFlags)` overload mirrors the existing entity mapper.

### Tests

- `ProtocolAdaptersResourceImplTest` (27) — asserts the mapped path / tag name / data type / access / description.
- `ProtocolAdapterManagerBrowseTest` (3), `ProtocolAdapterWrapperBrowseTest` (8) — follow the boundary type change.
- Real-HTTP e2e coverage lands in the companion **hivemq-edge-test** PR (see below).

### Cross-repo

- **hivemq-edge-test** companion PR: asserts the resolved fields end-to-end over real HTTP (`BrowseAndActivationEndToEndTest`, 17 tests green).
- CI-matching branches `ma/follow-up-v2-review-imp` pushed for `hivemq-edge-adapter-sdk` and `hivemq-edge-composite` (both == their initiative tips) so the composite build resolves the `api.v2` SDK instead of falling back to `master`.

Base: `initiative/nevsky-protocol-adapter-v2`.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/api/v2/resources/impl/ProtocolAdaptersResourceImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/api/v2/resources/impl/ProtocolAdaptersResourceImpl.java
@@ -15,6 +15,12 @@
  */
 package com.hivemq.api.v2.resources.impl;
 
+import static com.hivemq.api.v2.errors.ProtocolAdapterErrorFactory.adapterActivationInvalidError;
+import static com.hivemq.api.v2.errors.ProtocolAdapterErrorFactory.adapterNotFoundError;
+import static com.hivemq.api.v2.errors.ProtocolAdapterErrorFactory.adapterTypeNotFoundError;
+import static com.hivemq.api.v2.errors.ProtocolAdapterErrorFactory.tagNotFoundError;
+import static com.hivemq.util.ErrorResponseUtil.errorResponse;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.hivemq.adapter.sdk.api.schema.SchemaJsonRepresentation;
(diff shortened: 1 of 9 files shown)


hivemq-edge/src/main/java/com/hivemq/api/v2/resources/impl/ProtocolAdaptersResourceImpl.java
```diff
@@ -453,6 +453,18 @@ public ProtocolAdaptersResourceImpl(
                         access.getSubscribable().name()));
     }
 
+    private @NotNull AccessFlags toAccessFlags(final @NotNull com.hivemq.adapter.sdk.api.v2.node.AccessFlags access) {
+        return new AccessFlags()
+                .readable(com.hivemq.edge.api.v2.model.AccessTriState.fromValue(
```
caoccao, line 478: Can we remove the FQN?

caoccao (approved):

PR #1608, opened 2026-07-07
EDG-805: Build/packaging hygiene for the EtherNet/IP – Logical Addressing adapter

mschoenert-hivemq:
## Description (the why)
The adapter was contributed externally and moved into hivemq-edge; a systematic comparison against the 8 sibling OSS adapters surfaced several build/packaging deviations from convention. None change adapter behavior; they affect the build, the shipped artifact, or repo hygiene. Grouped so they are fixed together rather than discovered one at a time during a release.

Linear: EDG-805.

- [x] Rename the adapter icon `ethetnet-ip-odva.png` → `ethernet-ip-odva.png` (file + `getLogoUrl` reference); verified it ships in the jar under the correct name.
- [x] Use the version catalog (`libs.guava`, `libs.slf4j.jultoslf4j`) instead of hardcoded versions, matching the siblings.
- [x] Write third-party license output to `build/reports/` instead of into the source tree (`src/distribution/`); verified a full build no longer dirties `src/`.
- [x] Remove the stray `extra/` directory (demo scripts + a 164 KB config.xml) left over from the external contribution.

## Non-Goals
- No adapter behavior change.
- The split-package library-embedding problem is tracked separately in EDG-804.

## Note
Switching guava to the catalog exposed a leaked transitive import (`org.checkerframework...NonNull`) in `CompositeValuesFactory` — the old hardcoded guava was silently supplying `checker-qual`. Replaced with the project-standard `org.jetbrains.annotations.NotNull`.


caoccao (approved):
LGTM

PR #1609, opened 2026-07-08
EDG-786 browse discards resolved BrowsedNode data (path, default tag name, data type, access)

marregui:
card: [EDG-786](https://linear.app/hivemq/issue/EDG-786/rest-browse-discards-resolved-browsednode-data-path-default-tag-name)


marregui:
open in error

PR #1611, opened 2026-07-08
EDG-776: Drop undecodable OPC UA custom-struct values instead of publishing base64

marregui:
## Summary

Fixes [EDG-776](https://linear.app/hivemq/issue/EDG-776/opc-ua-custom-struct-value-published-as-base64-instead-of-json-under): under load, an OPC UA custom-struct value could be published with its `value` as a base64 string (raw binary encoding) instead of the expected JSON object — silent payload corruption around session (re)connect. Surfaced as flaky `SubscriptionResumeOpcUaIT` / `OpcUAAdapterNestedWriteTransformationIT` failures in the EDG-763 composite CI.

## Root cause (verified in Milo 1.1.4 source)

Milo resets the per-client dynamic codec registry (`DataTypeTree` / `DataTypeManager` / `EncodingContext`) on **every** session (re)activation and rebuilds it lazily on next use. The rebuild is best-effort: `ClientBrowseUtils` swallows browse/read failures at DEBUG and `DataTypeTreeBuilder` silently leaves `DataTypeDefinition = null` on bad status codes, so under load a "successful" rebuild can lack the codec for a custom struct. `ExtensionObject.decode` then throws `no codec registered for encodingId=…`, and `OpcUaToJsonConverter` caught this in a `catch (Throwable)` and published the raw binary body base64-encoded, logging only at DEBUG. The miss was never retried until the next session bounce.

(Refinement over the ticket text: Milo's `Lazy` is synchronized, so value delivery already blocks on the rebuild — the defect is the silently *partial* rebuild with no retry, not a value racing ahead of it.)

## Changes

- **`OpcUaToJsonConverter`**: removed the three base64 fallbacks; `ExtensionObject` decode failures now propagate (`UaSerializationException`).
- **`OpcUaSubscriptionLifecycleHandler.onDataReceived`**: on decode failure — WARN with tag context, error metric, drop the notification batch (the server resamples the monitored items; a half-built datapoint cannot be published), and reset the dynamic type registry so the next notification triggers a fresh rebuild. The reset is throttled (10 s, monotonic clock) so a permanently undecodable type cannot trigger a full `DataTypeTree` browse per notification.
- **`subscribe()` / `onTransferFailed`**: best-effort warm-up of the dynamic registry, keeping the expensive first build off the value-notification path. Deliberately non-fatal — failing startup on a slow tree browse would regress adapters that never use custom structs.

## Behaviour change

Servers whose custom types are **permanently** undecodable (non-compliant: no `DataTypeDefinition`/encoding nodes exposed) previously got base64 published as a "valid" value; they now get dropped samples with WARN logs and `subscription.data.error.count` increments. Visible errors instead of silent binary-in-JSON.

## Testing

- New unit tests: decode failure propagates (no base64), batch dropped + registry reset on codec-miss, throttle prevents repeated resets, happy path still publishes.
- Full `hivemq-edge-module-opcua` suite: 209/209 passed (1 pre-existing `@Disabled` skip).
- Test-side robustness (ticket item 3) is covered separately by hivemq/hivemq-edge-test#377.

### DIFF
--- a/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/listeners/OpcUaSubscriptionLifecycleHandler.java
+++ b/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/listeners/OpcUaSubscriptionLifecycleHandler.java
@@ -33,6 +33,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.eclipse.milo.opcua.sdk.client.OpcUaClient;
@@ -41,6 +43,7 @@ import org.eclipse.milo.opcua.sdk.client.subscriptions.MonitoredItemSynchronizat
 import org.eclipse.milo.opcua.sdk.client.subscriptions.OpcUaMonitoredItem;
 import org.eclipse.milo.opcua.sdk.client.subscriptions.OpcUaSubscription;
 import org.eclipse.milo.opcua.stack.core.UaException;
+import org.eclipse.milo.opcua.stack.core.UaSerializationException;
 import org.eclipse.milo.opcua.stack.core.types.builtin.DataValue;
 import org.eclipse.milo.opcua.stack.core.types.builtin.NodeId;
 import org.eclipse.milo.opcua.stack.core.types.builtin.StatusCode;
@@ -52,7 +55,10 @@ import org.slf4j.LoggerFactory;
 public class OpcUaSubscriptionLifecycleHandler implements OpcUaSubscription.SubscriptionListener {
 
     public static final long KEEP_ALIVE_SAFETY_MARGIN_MS = 5_000L;
+    public static final long TYPE_REGISTRY_RESET_THROTTLE_MS = 10_000L;
 
+    private static final long TYPE_REGISTRY_RESET_THROTTLE_NANOS =
+            TimeUnit.MILLISECONDS.toNanos(TYPE_REGISTRY_RESET_THROTTLE_MS);
     private static final @NotNull Logger log = LoggerFactory.getLogger(OpcUaSubscriptionLifecycleHandler.class);
     private static final int MAX_MONITORED_ITEM_COUNT = 5;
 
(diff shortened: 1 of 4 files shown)


marregui:
**CI run #1 failed — root-caused and fixed in bc8159ec3.**

The 69 failures in `:hivemq-edge-test:test` were all OPC UA connection/subscription lifecycle (reconnect-test timeouts, a leaked reconnect loop from an adapter blocked in stop, and the resulting ERROR-log pollution failing ~30 unrelated tests via the shared log check). Master baseline for the same job is green (the two recent UNSTABLE master builds each had a single unrelated flaky test).

Notably, the log contains **zero** `UaSerializationException` / "no codec registered" — the converter change and the drop/reset recovery never executed. The only branch code exercised by those ITs was the synchronous type-registry warm-up, which:

- ran a full DataTypeTree browse inside `OpcUaClientConnection.start()`'s synchronized block, so `stop()` blocked behind in-flight (re)connects (log shows a stop blocked 6.4 s behind a connection attempt → "failed to transition from Idle to Stopping" → leaked 20-minute reconnect loop);
- issued the same browse on Milo's callback thread in `onTransferFailed`, adding request traffic exactly during session instability (reconnect phase shows a session-churn loop: 8 sessions built and discarded in ~6 s).

It was also redundant: EDG-577 (`OpcUaProtocolAdapter.warmUpBrowseReadiness`, on master) already hydrates the namespace table and DataTypeTree asynchronously after every successful (re)connect.

`bc8159ec3` removes the warm-up from both call sites. The EDG-776 correctness fix — no base64 fallback + drop/throttled-reset recovery on codec-miss — is unchanged; every code path exercised by the failing ITs is now back to master semantics.

marregui:
**CI build #3 triage: down from 69 failures to 2, neither caused by this branch.**

- `EtherIpCipOdvaBatchReadIT#test_twoTagsAtSameAddress_bothPublishIndependently` — the only test that failed the task. `ContainerLaunchException`: the ENIP test fixture binds fixed host port 44818 and the CI Docker host already had it in use ("failed to bind host port 0.0.0.0:44818/tcp: address already in use"); the static-fixture failure is sticky per JVM, so all retries fail by construction. Different protocol module, untouched here.
- `DeviceTagEndToEndIT#test_browseAndImport_yaml_roundTrip` — browse REST returned 503 (adapter not yet connected under parallel load), passed on retry. Pre-existing startup-timing flake, unrelated to the northbound decode changes.

Incidentally, the new decode-failure path executed exactly once in this run — a late value notification arrived during another test's adapter teardown after codecs were unregistered (`no codec registered for encodingId=ns=2;s=CustomStructType.BinaryEncoding`). It logged the WARN, dropped the batch, reset the registry, and no base64 was published and no test failed — the exact EDG-776 scenario, handled as designed.

Retriggered for a clean run (build #4).

PR #1612, opened 2026-07-08
fix(EDG-808): publish v2 adapter values northbound

caoccao:
Linear: https://linear.app/hivemq/issue/EDG-808/nevsky-bug-fix-missing-northboundtagconsumer

## Summary
- add the missing v2 northbound bridge that registers mapped `NorthboundTagConsumer`s with `TagManager`
- stamp accepted v2 adapter datapoints with the mapped tag name and adapter id before feeding them northbound
- keep the tag aspect state machine as the acceptance gate so ignored/deactivated datapoints are not published
- adapt the v2 adapter identity into the legacy publish service path used by `NorthboundTagConsumer`

## Root cause
The v2 wrapper accepted datapoints from protocol adapters, but only routed them into the tag aspect runtime. Unlike the v1 wrapper, it never registered northbound consumers with `TagManager` and never fed accepted mapped datapoints into the northbound publish path. As a result, mapped tags such as `t1` could update the v2 runtime without producing MQTT messages.

## Fix
The wrapper factory now creates a v2 northbound consumer registry when `TagManager` is available. The registry builds consumers from v2 northbound mappings, registers them with `TagManager`, and updates them when tag/mapping configuration changes. `routeDataPointToTags` now receives the stamped accepted datapoint from the tag runtime and feeds it to `TagManager` for MQTT publishing.

## Validation
- `./gradlew :hivemq-edge-build:hivemq-edge:test --tests com.hivemq.protocols.v2.manager.DefaultProtocolAdapterWrapperFactoryTest --tests com.hivemq.protocols.v2.runtime.DataPointStampingTest --no-daemon`
- `./gradlew :hivemq-edge-test:test --tests com.hivemq.edge.api.v2.NorthboundMqttEndToEndTest --no-daemon`
- `./gradlew :hivemq-edge-build:hivemq-edge:spotlessCheck :hivemq-edge-test:spotlessCheck --no-daemon`
- `git diff --check` in `hivemq-edge` and `hivemq-edge-test`

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/protocols/northbound/NorthboundConsumerFactory.java
+++ b/hivemq-edge/src/main/java/com/hivemq/protocols/northbound/NorthboundConsumerFactory.java
@@ -16,6 +16,7 @@
 package com.hivemq.protocols.northbound;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.hivemq.adapter.sdk.api.ProtocolAdapter;
 import com.hivemq.adapter.sdk.api.events.EventService;
 import com.hivemq.adapter.sdk.api.services.ProtocolAdapterMetricsService;
 import com.hivemq.edge.modules.adapters.impl.ProtocolAdapterPublishServiceImpl;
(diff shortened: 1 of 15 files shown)


marregui (approved):
lgtm!

PR #1619, opened 2026-07-17
EDG-812: Nevsky v2 — databases-v2 adapter module and multi-value poll boundary

caoccao:
**Motivation**

Part of [EDG-812](https://linear.app/hivemq/issue/EDG-812): port the Databases adapter to SDK v2 as a new, parallel, production-listable module (protocol-id `databases-v2`, v1 untouched). Databases is the first port that is not adapter-only: to preserve `spiltLinesInIndividualMessages` (one MQTT message per result row) exactly, the shared framework gains an explicit poll-completion boundary so a poll may publish 0..N values before its cadence resumes.

Companion PRs on the same branch: hivemq-edge-adapter-sdk (`pollComplete` output contract), hivemq-edge-test (end-to-end suite), hivemq-edge-composite (distribution).

**Changes**

*New module `hivemq-edge-module-databases-v2`*
- A genuine fat shadow jar bundling the PostgreSQL, MariaDB, and MSSQL drivers plus HikariCP, with the MariaDB LGPL license exclusion mirrored from v1; wired into the distribution and included builds. No changes to the `hivemq-edge` version catalog.
- `DatabasesProtocolAdapter` (actor on `AbstractProtocolAdapter`): connect-time classloader dance + driver registration, HikariCP pool open/validate on connect and close on disconnect, per-tag SQL polls with the verbatim `java.sql.Types` row-to-JSON mapping, both message-shaping modes (array / split-lines).
- Three v1 defects deliberately corrected (documented in the README for review): connection-timeout doubling (`*2000` → `*1000`), the dead `trustCertificate` setting (now drives MSSQL `trustServerCertificate`), and the statement/result-set leak (full try-with-resources).
- Tests: unit coverage over a stubbed JDBC layer, plus wrapper-driven integration tests running a real `ProtocolAdapterWrapper` with the real polled read-aspect machines against real PostgreSQL, MySQL (via the MariaDB driver), and MSSQL containers — plain `GenericContainer`s, readiness gated through the same drivers the adapter bundles (the engine-specific Testcontainers modules stay in hivemq-edge-test only).

*Framework enhancement — multi-value polls*
- `PollCompleted` wrapper event in the **DATA** band (within-band FIFO delivers it strictly after the values it terminates) with a named ERROR absorb row; routed through the tag-aspect coordinator to the polled read aspect.
- The polled read aspect now publishes-and-stays on a value; only the completion (or a per-node error) schedules the next poll — so a zero-value poll completes instead of hanging.
- Migration: HTTP-v2 opts out of the template auto-completion and completes from its `whenComplete` handler on every path; File-v2 unchanged (synchronous template path).
- New aspect tests: multi-value poll, zero-value poll, stale-value absorption after completion, DATA-band FIFO ordering, completion trailing a failure, ERROR-state absorption.

*Chaos — split poll and poll completion in the script DSL*
- `PollBehavior.values(...)` (multi-value poll), `valueThenCompletionAfterTicks(...)` (value now, completion when its tick comes due — the poll stays observably open in between), and `completionWithoutValue()` (zero-value poll).
- Simulator unit tests plus wrapper-driven scenario-matrix coverage of the new boundary.


marregui (approved):
bazinga

PR #1620, opened 2026-07-17
EDG-813: durable southbound writes — message-driven delivery on the actor mailbox

marregui:
**EDG-813 — durable southbound writes.** The complete MQTT → device write path for v2 protocol adapters: durable, back-pressured, fire-and-forget.

> ⚠️ **This PR has been substantially rewritten since it was opened.** The original description (per-command verdicts, crash-replay dedup) describes a design that **no longer exists on this branch**. See *What changed* below.

## What the feature does

A publish to a southbound mapping's command topic is executed on the device **at least once**, surviving device silence, device rejection, connection loss, southbound deactivation, tags-only reloads, adapter recreates and Edge restarts — while the adapter keeps a strict single-in-flight, never-buffering write window. Bursts wait in the broker's durable client queue, never in the adapter. **Fire-and-forget: no reply, no error topic, no result message.**

## What changed since this PR was opened

1. **Per-command verdicts were removed** (Martin's decision, 2026-07-21). The path is fire-and-forget; client feedback is a settled product decision, not a gap. `SouthboundWriteCompletion` and the verdict reporter are gone.
2. **The delivery layer was rebuilt on the actor mailbox.** It was the only part of the v2 runtime that wasn't single-threaded, and it paid for that with a monitor, a documented lock order, disposal outside that monitor, and a ~150-line recovery ladder compensating for an edge-triggered wakeup. All delivery state now lives on the adapter's single dispatch thread and everything it reacts to arrives as a message. **~350 lines smaller, and three wedge bug classes stopped being expressible.**
3. **Every finding from Sam's Code Review 01 is now closed** — B1–B7 and I1–I3. Highlights:
   - **B6** — an adapter recreate could erase every pending command. Internal southbound queues are now exempt from the broker's orphan cleanup (subscriber absence is not a liveness signal for a queue whose consumer unsubscribes for the whole span of a recreate, and that job runs every 4 s). Reclamation is explicit instead.
   - **B2** — a re-pointed tag delivered the old node's commands to the new one. **Martin's ruling: dead-letter them.** Detected on the `node-string` at config level, so the SDK is untouched. Covered on both doors: the tags-only reload *and* the full recreate a bundled connection-critical change turns it into.
   - **B3** — QoS 0. Now delivered **best-effort with a loud warning** rather than refused (see below).
   - **B4** — a late duplicate `writeResult` misattributed. Fixed by an attempt id on the SDK contract (`hivemq-edge-adapter-sdk#131`).
   - **B5** — an undeletable queue entry caused perpetual re-execution. The channel now remembers disposed ids and refuses a second execution, counted on `writes.redeliveries-refused`.
   - **I2** — a partially-constructed intake leaked subscriptions.
4. **QoS 0 policy changed late.** It was briefly *refused*; it is now **delivered best-effort** and warned about. What the publisher gave up is real and unrecoverable — the broker hands a QoS 0 publish out and removes it in one read, so there is nothing to redeliver — but a command Edge can execute should be executed. **Note this reverses what the earlier B3 fix did.**
5. **Full command-lifecycle logging.** Every command now traces *picked up → handed to the device → executed/committed* at INFO, with the non-success outcomes logged too. Previously a successful command was entirely silent.
6. **Durability is now actually tested.** `SouthboundWriteXodusRestartEndToEndTest` is the first test in either repo to exercise `ClientQueueXodusLocalPersistence` — and so the first to run the bootstrap marker sweep the whole at-least-once argument rests on.

## New observability

| metric | meaning |
|---|---|
| `…writes.timeout` | the adapter accepted a write and never acknowledged it |
| `…writes.dead-lettered` | a command destroyed — device refusal, undecodable payload, or a re-pointed tag |
| `…writes.redeliveries-refused` | a disposed command the store handed back — **must read zero** |

## Review notes

- **`SystemDispatcher.close()` now joins the dispatch thread.** This is the one change that reaches **every v2 actor** and is exposed through an SDK-facing handle. It wants a non-author sign-off. Two latent consequences, neither reachable with the adapters in tree today: teardown can cost up to 5 s per binding if an adapter blocks its dispatch thread, and `RecordingDispatcher` / `IdempotentHandle` hold a monitor across the join.
- **This branch also carries EDG-824**, which is not yet merged into `initiative/nevsky-protocol-adapter-v2`. Rebasing onto the current initiative surfaced four places where EDG-824's enforcements meet newer initiative code: late-poll escalation, spontaneous-loss deferral, the access model, and schema conformance. Each is reconciled here, but **the clean resolution is to land EDG-824 into initiative first**.
- Every guard added was **mutation-verified** — deleted, the test covering it fails.

## Companion PRs

`hivemq-edge-test#388` · `hivemq-edge-adapter-sdk#131` · `hivemq-edge-commercial-modules#347`


hivemq-edge/src/main/java/com/hivemq/protocols/v2/southbound/SouthboundBrokerRuntime.java
```diff
@@ -0,0 +1,40 @@
...
+ * backlogs lease from, the publish service the verdicts go out through, and the retained-message store the
+ * crash-replay dedup recovers the last verdict from. Absent as a whole in unit rigs — the plane then falls back
+ * to the interim in-memory backlogs.
+ *
+ * @param topicTree                  the broker topic tree.
+ * @param clientQueuePersistence     the durable client queue store.
+ * @param internalPublishService     the broker-internal publish path (verdicts).
+ * @param retainedMessagePersistence the retained-message store (verdict recovery).
+ */
+public record SouthboundBrokerRuntime(
```
caoccao, line 31: This record name doesn't make sense. Is it a runtime? Or could we have a better name?

hivemq-edge/src/main/java/com/hivemq/protocols/v2/southbound/SouthboundMqttIntake.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /** The minimal payload seam: UTF-8 payload carried as a JSON-flagged value; no payload → untranslatable. */
+    private @NotNull SouthboundPublishTranslator translator(final @NotNull String tagName) {
+        return publish -> {
+            final byte[] payload = publish.getPayload();
+            if (payload == null) {
+                return null;
+            }
+            return dataPointFactory.createJsonDataPoint(tagName, new String(payload, UTF_8));
```
caoccao, line 218: Do we have an integration test case for this? I found there are some issues with `createJsonDataPoint`.

caoccao (changes requested):
Here is my review.

https://linear.app/hivemq/document/edg-813-durable-southbound-code-review-01-2ff9632ecb66

marregui:
Superseded by the mono-repo port: hivemq/hivemq-edge2#23. New work lands in hivemq-edge2 since 2026-09-09; this branch stays for reference.

PR #1633, opened 2026-07-21
PLT-1510: ci: load the composite trigger from the shared library

ymengesha:
Part of [PLT-1510](https://linear.app/hivemq/issue/PLT-1510). **Do not merge before** the `hivemq-edge-composite` Jenkins global library exists and the multibranch job is repointed at [`hivemq-edge-composite-job`](https://github.com/hivemq/hivemq-edge-composite-job) — see the merge order on the ticket. Merging early leaves this repo unable to trigger a composite build.

## What

Replaces this repo's hand-rolled `CompositeJenkinsfile` trigger pipeline with a call to the shared `edgeCompositeTrigger()` var:

```groovy
@Library('hivemq-edge-composite@master') _
edgeCompositeTrigger()
```

## Why

Every Edge component repo carries its own near-identical copy of the same ~35-line trigger, and each one clones `hivemq-edge-composite` to seed a missing branch. The composite pipeline now lives in `hivemq-edge-composite-job`, which backs the multibranch job — so the seed target is wrong in every copy, and fixing it in place means the same edit in six repos, again, the next time it changes.

The shared version also fixes two things the copies get wrong (both hit in the Platform composite as PLT-1199 / PLT-1235): it decides branch (re)creation from the git ref itself rather than from whether `build job:` threw, and it rescans the multibranch project after a recreate so a missed webhook cannot leave the child job absent.

This is the same trigger the Platform component repos already load from `hivemq-composite`.


marregui (approved):
LGTM

PR #1634, opened 2026-07-22
EDG-142: Add OIDC (OpenID Connect) authentication

mschoenert-hivemq:
**Motivation**

EDG-142 — add OpenID Connect (SSO) login for the Edge REST API via an external Identity Provider. Customer-committed (Miele) for Q3. Design doc: https://linear.app/hivemq/issue/EDG-142/edge-auth-openid-integration

The backend stays fully stateless. The SPA receives a short-lived Edge-issued JWT whose roles are mapped from the IdP claim. It reuses the existing JWT/auth stack (`JwtAuthenticationProvider`, `ITokenGenerator`, `BearerTokenAuthenticationHandler`, `@RolesAllowed`), so the new surface is narrow.

**Changes**

- `<oidc-authentication>` config entity (issuer, client, role-mappings, optional truststore, connection-timeout) wired into `AdminApiEntity`; runtime `OidcConfiguration`; `ApiConfigurationService` getter/setter; `ApiConfigurator` mapping. OIDC is disabled by default.
- `OidcStateStore` — in-memory state → nonce/PKCE store (one-time-use, TTL, fixed-capacity ring).
- `OidcService` / `OidcServiceImpl` — Nimbus-backed: cached IdP discovery with issuer + HTTPS-endpoint validation, redirect-disabled outbound calls, PKCE authorization-code flow, callback (code exchange → JWKS-verified ID-token validation, asymmetric-only signing set → role mapping → Edge JWT). The callback result page is a FreeMarker template.
- Endpoints `/api/v1/auth/mode` and `/api/v1/auth/oidc/{login,callback}` (OpenAPI + `AuthMode` schema), implemented in `AuthenticationResourceImpl`; Dagger-bound in `ApiModule`. Login-start is a `POST` (it allocates one-time server-side state) with a `no-store` response.
- Frontend: auth-mode detection, a "Login with SSO" flow (popup + `postMessage`, single in-flight login, in-memory token), no local fallback in OIDC-only mode.
- New dependency: Nimbus OAuth2 OIDC SDK.
- Tests: `OidcStateStoreTest`, `OidcConfigurationTest`, discovery-validation and redirect tests, and a Testcontainers-Keycloak integration test proving the full flow end to end (alice→ADMIN, bob→USER, unmapped→denied). Cypress `LoginPage` / `Login` component specs cover the SPA flow.

**Review**

Three rounds of code review by @caoccao, all findings addressed or declined with reasoning (anchored on the review documents on EDG-142): review 1, review 2 (12 findings), and review 3 (5 findings — discovery-metadata validation, redirect-disabling, POST login-start + `no-store`, log sanitization, config-equality cleanup). A self-authored complexity/gap review (7 findings) was also worked through.

**Acceptance criteria**

- [x] OIDC login via an external IdP issues an Edge JWT with mapped roles; verified against real Keycloak.
- [x] IdP roles map to Edge roles (`admin`/`super`/`user`); a user with no mapped role is denied.
- [x] OIDC disabled by default; local username/password login unchanged when unconfigured.
- [x] Backend stateless (except the bounded in-memory login-state store); PKCE + nonce + state validated.
- [x] Frontend "Login with SSO" flow.
- [x] Operator documentation (docs PR https://github.com/hivemq/hivemq4-documentation/pull/2887 — merges separately, on its own release cadence).

**Non-goals (this PR)**

- Making policies/authorization customer-editable; anything beyond authentication.

**Notes for merge**

- Known non-blocking red checks: SonarCloud (tracked as EDG-832) and a Snyk `dompurify` transitive CVE, both unrelated to OIDC.
- QA (Estefania) is scheduled against `master` after merge.


gradle/libs.versions.toml
```diff
@@ -146,6 +147,7 @@ mockito-junit-jupiter = { module = "org.mockito:mockito-junit-jupiter", version.
 mssql = { module = "com.microsoft.sqlserver:mssql-jdbc", version.ref = "mssql" }
 mtconnect-protocol = { module = "com.hivemq:hivemq-mtconnect-protocol", version.ref = "mtconnect-protocol"}
 nullaway = { module = "com.uber.nullaway:nullaway", version.ref = "nullaway" }
+nimbus-oauth2-oidc-sdk = { module = "com.nimbusds:oauth2-oidc-sdk", version.ref = "nimbus-oauth2-oidc-sdk" }
```
caoccao: Please sort the properties by name. `nullaway` is also in the wrong order.

hivemq-edge-frontend/src/locales/en/translation.json
```diff
@@ -113,6 +113,9 @@
     "submit": {
       "label": "Sign in"
     },
+    "sso": {
+      "label": "Login with SSO"
```
caoccao: This key is too general. I recommend we design a better, specific key.

hivemq-edge-frontend/src/modules/Auth/hooks/useOidcLogin.ts
```diff
@@ -0,0 +1,81 @@
...
+        window.clearInterval(pollRef.current)
+        pollRef.current = undefined
+      }
+      pendingRef.current?.resolve(data.token)
+      pendingRef.current = null
+      popupRef.current?.close()
+      popupRef.current = null
+    }
+
+    window.addEventListener('message', onMessage)
```
caoccao, line 63: We use React, but still use raw event listener. Something is not right.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
```diff
@@ -0,0 +1,346 @@
...
+                .type(MediaType.TEXT_PLAIN)
+                .entity(message)
+                .build();
+    }
+
+    /**
+     * Minimal HTML page delivered on success: posts the token to the opener window (the popup pattern)
+     * and closes the popup, keeping the JWT out of URLs. The origin is derived from the redirect URI.
+     */
+    private static @NotNull String tokenDeliveryHtml(final @NotNull String jwt, final @NotNull URI redirectUri) {
```
caoccao: Do we really have to put raw html string in the Java code? If yes, as we have freemarker template engine in Edge, I would recommend we convert it to a template in resource.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
```diff
@@ -0,0 +1,346 @@
...
+                + "</script></body></html>";
+    }
+
+    private static @NotNull String originOf(final @NotNull URI uri) {
+        final String scheme = uri.getScheme();
+        final String host = uri.getHost();
+        final int port = uri.getPort();
+        final StringBuilder sb = new StringBuilder();
+        sb.append(scheme).append("://").append(host);
+        if (port != -1 && !(("https".equals(scheme) && port == 443) || ("http".equals(scheme) && port == 80))) {
```
caoccao: I don't think hardcoding ports is correct because ports can be customized in the Edge config. Also, there could be reverse proxy in the front. Maybe I was wrong, but please double check this.

caoccao (changes requested):
I left some comments and also https://linear.app/hivemq/document/sam-code-review-2026-07-23-3a20a2f387da.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
mschoenert-hivemq: Double-checked, and I think this one's fine — these aren't Edge listener ports.

originOf builds the targetOrigin string for postMessage, derived from the configured redirect-uri. The 443/80 test omits the port only when it is the scheme default, because `https://edge.example.com:443` and `https://edge.example.com` are the same origin, and postMessage compares origins as canonical strings — including the explicit :443 would make the target never match the opener and silently break every login.

A custom port is preserved: `http://localhost:8080/callback` yields `http://localhost:8080`. So an operator running Edge on any port works, because the origin comes from whatever URL is configured, not from a listener.

Reverse proxies are covered for the same reason. The redirect-uri is the URL the browser reaches — the proxy's public address — so the origin is the proxy's, which is exactly what the SPA's window has. Edge's own bind address and port never enter into it.

The one real constraint is that redirect-uri must be the browser-facing URL rather than an internal one, which is now enforced at startup (absolute http/https with a host) and documented in the operator guide.

hivemq-edge-frontend/src/locales/en/translation.json
mschoenert-hivemq: changing to  `login.sso.buttonLabel`

gradle/libs.versions.toml
mschoenert-hivemq: will do

hivemq-edge-frontend/src/modules/Auth/hooks/useOidcLogin.ts
mschoenert-hivemq: This one is intentional, and I don't think there's a React-idiomatic alternative.

React's synthetic event system only covers DOM events on elements React renders. message is a window event fired by postMessage from another browsing context — there's no React element to attach it to, and no synthetic equivalent. Subscribing with addEventListener inside useEffect and unsubscribing in the cleanup is the standard React pattern for window- and document-level events; it's what useSyncExternalStore and every library-level event hook does underneath.

Worth noting your comment predates the rewrite for findings #7 and #13. The hook now also correlates on event.source (only the popup it opened), requires an explicit message type, permits one login in flight, and routes every exit — success, failure, popup closed, unmount — through a single settle() so the promise resolves exactly once. The cleanup is stricter than what you reviewed.

If the concern is that the subscription logic is bare rather than the listener itself, I could extract a small useWindowMessage hook — but that's wrapping, not removing, the listener. Happy to do it if you'd prefer the shape.

caoccao: It would be nice to have a dialog component for it. But, this is acceptable to me as long as in the near future we would rewrite the whole thing. Up to you.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
caoccao: Fine. Thank you.

caoccao:
I've had a second round review. The doc is https://linear.app/hivemq/document/sam-code-review-2026-07-24-7f891f543df1. Some of them are worth fixing.

mschoenert-hivemq:
**On the red SonarCloud and frontend-lint checks — expected, not addressed here by design.**

Both are triggered only by the auto-generated OpenAPI client files this PR adds (src/api/__generated__/…), which are already excluded from our own ESLint and Prettier config. SonarCloud flags the generator's blanket /* eslint-disable */ header, and the lint check trips on the same generated files — neither is a defect in hand-written code. Aligning SonarCloud with the existing ESLint/Prettier exclusions is tracked separately in EDG-832; we're overriding these two gates for this PR rather than hand-editing generated output. All other checks (backend, builds, OpenAPI, Cypress, unit tests, internal SonarQube) are green.

caoccao:
I think https://linear.app/hivemq/document/sam-code-review-2026-07-27-39deaf362c2c might be the final review.

caoccao (approved):

PR #1635, opened 2026-07-22
EDG-824: fixes for Nevsky QA findings (config / topic / access / capability validation)

marregui:
**Motivation**

Fixes for the findings of the Nevsky QA campaign: [EDG-824 — Nevsky QA: protocol-adapter v2 findings (config / topic / access validation gaps)](https://linear.app/hivemq/issue/EDG-824/nevsky-qa-protocol-adapter-v2-findings-config-topic-access-validation).

The ticket description is the current index of every finding and its disposition. Three review rounds by Sam (2026-07-23, 07-30, 07-31) and three QA rounds by Estefania are folded into it.

**Reviewed / current revisions**

| Repository | Revision |
| -- | -- |
| `hivemq-edge` (this PR) | `053eba09d` |
| `hivemq-edge-test` PR #396 | `546703b7c` |
| `hivemq-edge-adapter-sdk` companion branch | `726460167` |
| `hivemq-edge-composite` companion branch | `7ae71f643` |

The three companion revisions are unchanged since Sam's round-3 review; only `hivemq-edge` moved.

**Changes — QA campaign findings** (ticket numbering)

- **#2 (SEVERE) reconfigure wedge** — `TagAspectRuntimeCoordinator` mirrors the adapter's live phase and replays it into rebuilt tag aspects, so a runtime reconfigure re-verifies tags in place instead of parking them in `WAITING_FOR_ADAPTER_READY`; `UpdateTagSet` carries `pollIntervalMillis` so interval-only changes apply.
- **#3 (SEVERE) southbound write unwired** — new `southbound/` package (`SouthboundWriterRegistry`, `SouthboundWriteBridge`, `WritePayload`) wires MQTT→adapter writes through the v1 `InternalProtocolAdapterWritingService` seam into the wrapper mailbox / write aspect.
- **#4 (SEVERE) malformed config node-fatal** — `ProtocolAdapterExtractor` scopes every validation failure to its adapter and always returns SUCCESS: an invalid replacement keeps the previously-applied config running, an invalid new adapter surfaces as a REST-visible ERROR handle, and `sync()` re-emits the verbatim parsed section so the derived view never clobbers `config.xml`.
- **#14 (HIGH) access model decorative** — enforced at config translation (`ProtocolAdapterConfigSupport`): `readable` is the master read gate and only `YES` permits; a forbidden tag is never polled, subscribed or delivered; `writable≠YES` accepts no write. Omitted `<access>` is unrestricted, explicit is enforced, contradictions raise config WARNINGs.
- **#17 (HIGH) capabilities** — closed 2026-07-28 as a stale premise; polling is the ungated baseline and the genuine enforcement (browse against a non-BROWSE type → 4xx) is pinned by a contract test.
- **#13 / #12 topic validation** — northbound mappings reject `$`-prefixed topics, wildcards/NUL and >65 535 UTF-8 bytes; southbound filters get subscribe-semantics validation.
- **#11 (LOW) colliding topics** — a WARNING per topic mapped more than once, never a rejection.
- **#6 (MEDIUM) typed schema unenforced** — `runtime/SchemaConformance` enforced at the northbound routing point against the same schema REST projects; violations are refused northbound and surfaced per tag without wedging the aspect.
- **#15 (LOW) stalled poll** — a poll-result deadline on the aspect's single timer slot (fed by `command-timeout-millis`), escalation through re-verification after three consecutive failures, **and** a five-minute no-published-value deadline that folds a producing-looking tag to ERROR until a reading actually arrives.
- **#16 (LOW) skipped power cycle** — a spontaneous subscription loss now cancels first; `BatchCollector` dispatches a same-tick remove-then-add as a `REMOVE_THEN_ADD` power cycle.
- **#7 (MINOR) green-while-dead** — `ProtocolAdapterWrapper.receive` catches `Throwable` from adapter-facing calls → defensive-reset metric + ERROR, with a guarded error path and an ERROR-snapshot fallback.
- **#19 (HIGH) a stop that never lands had no terminal state** — the goal-driven `stop()` out of `ERROR` is now issued at most once per adapter life, so an adapter that will not stop settles in a terminal `RED_ERROR` instead of cycling through `WAITING_FOR_STOPPED` forever behind a permanent "stopping" status. Covers all three shapes: no acknowledgment (watchdog-paced), an `error(ADAPTER)` answer and a throw — the latter two re-commanded the stop with no delay at all. A new `wrapperStopFailed` signal lets the manager complete a discard or recreate that was waiting on a `stopped()` that is never coming.

**Changes — Sam Code Review 02**

CR2 2 construction rollback · CR2 3 Jackson value carriers · CR2 5 stalled-poll status honesty · CR2 6 fatal-JVM propagation at every catch-all · CR2 7 `$share` rejected outright · CR2 8 the out-of-range workload E2E · CR2 test-gaps 2, 3, 6, 7.

**Changes — Sam Code Review 03**

- **CR3 1 / QA #19** — above.
- **CR3 2 fatal VM error left a dead dispatcher with a green REST snapshot** — the dying thread now corrects its own status before the fatal is rethrown: the machine is never transitioned (a `VirtualMachineError` is not an adapter error) but the published snapshot stops advertising an adapter nobody is running, and a new `wrapperDied` signal lets the manager release the dead actor — above all the periodic tick, which runs on the clock's scheduler and would otherwise keep filling a mailbox nobody drains. The adapter stays visible in ERROR and a configuration change recreates it. **Product decision (Miguel, 2026-07-31): one adapter's implementation failing does not take the process down** — Edge carries on and reports the error, so no process-level terminator was added.
- **CR3 3 the five-minute deadline ignored successful-but-unpublished polls** — it is now evaluated at the top of every poll request, not only on the failure path, so a cooperative adapter whose values the schema refuses (or that completes with no values at all) can no longer read healthy forever. Placement is deliberate: a published reading is recorded *after* the completing value has driven the machine, so evaluating at completion would judge a cycle before its own publish was counted. The deadline is five minutes **or one whole poll cycle, whichever is longer**, so a device polled every ten minutes is not declared unreadable during every in-flight poll.
- **CR3 4 construction rollback could not undo constructor-internal acquisition** — both registries now acquire nothing in their constructors; the factory registers each with the `ConstructionScope` and only then wires it, so a failure on the second consumer or a throwing writing-service start is rolled back by the scope that already owns it.
- **Review thread** — gratuitous fully-qualified type names removed across the v2 packages. The remaining one in `ProtocolAdapterPublishIdentity` is unavoidable: that class is the v1↔v2 bridge and imports the **v1** `ProtocolAdapterInformation`, so the v2 type cannot share the simple name.

**Accepted residuals — deliberate, tracked**

- **R1 / CR2 1 (Critical) southbound acknowledge-then-drop** and **R2 / CR2 4 (High) writing-availability callback** are owned by [EDG-813](https://linear.app/hivemq/issue/EDG-813), which replaces this seam wholesale. Sequencing is **EDG-824 merges first, EDG-813 follows**; per Sam's round-3 gate 7 this must not ship before EDG-813 provides truthful acknowledgement/durability semantics.
- **R5 (part) tags-only reload transactionality** — adapter *construction* is now all-or-nothing, but the tags-only multi-component commit is deliberately left to EDG-813, which already rewrote those exact lines. Fixing it here would be discarded and would re-conflict on every rebase.
- **EDG-829 / EDG-830 / EDG-831** — poll-to-result correlation (SDK contract), last-known-good persistence, per-tag poll cadence.

**Verification**

- `com.hivemq.protocols.v2.*`: **358 tests / 51 classes**, 0 failures.
- `hivemq-edge` module suite: **5 003 tests**, 0 failures. Chaos module: 56, 0 failures. All 13 adapter modules green. `spotlessCheck` clean. JDK 21.
- Every fix in the round-3 batch was verified fail-first: the test was written against the unfixed code and observed failing before the fix landed.
- E2E coverage in the companion `hivemq-edge-test` PR #396.

**Open gates before merge**

1. The Nevsky acceptance / tripwire suite run against these exact heads.
2. An exact-head composite Jenkins build (the last recorded build predates the 77/777 correction).


hivemq-edge/src/main/java/com/hivemq/protocols/v2/southbound/WritePayload.java
```diff
@@ -0,0 +1,42 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.hivemq.adapter.sdk.api.writing.WritingPayload;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * The southbound MQTT payload for a v2 adapter write (EDG-824 #3): the conventional
+ * {@code {"value": <json>}} shape every v1 module uses. The value is kept as raw JSON — the v2 write path carries it
+ * to the adapter as a JSON {@code DataPoint}; interpreting it against the node is the protocol adapter's job.
+ */
+public class V2WritePayload implements WritingPayload {
```
caoccao: As the package name already has `v2`. I don't recommend any class to be named as `V2`.

caoccao (changes requested):
I hold some concerns. https://linear.app/hivemq/document/sam-code-review-2026-07-23-bf027245f184

hivemq-edge/src/main/java/com/hivemq/protocols/v2/config/ProtocolAdapterEntity.java
```diff
@@ -257,6 +264,32 @@ public void validate(final @NotNull List<ValidationEvent> validationEvents) {
 
         retryPolicy.validate(validationEvents);
         tags.forEach(tag -> tag.validate(validationEvents));
+        // The access model is enforced at runtime (EDG-824 #14): an activated aspect with no permitted capability
+        // silently never operates. That must be a deliberate choice, so surface the contradiction as a warning. An
+        // omitted <access> element is unrestricted (all flags YES, via TagEntity's initializer) and so never trips
+        // this warning; only an explicit <access> element — whose omitted child attributes default to NO — can.
+        tags.forEach(tag -> {
+            final AccessFlagsEntity access = tag.getAccess();
```
caoccao: Could this whole block be moved into tag.validate()?

caoccao (changes requested):
I left a comment and attached a new review doc https://linear.app/hivemq/document/sam-code-review-2026-07-30-b13bdd29d548

marregui:
**Code Review 02 — all implementable findings addressed.**

Fixed since your review:

| CR2 | Fix | Commit |
|--|--|--|
| 3 (High) — Jackson carriers rejected | `SchemaConformance` unwraps JSON scalar carriers to their logical Java value before judging type or range | `09055eccb` |
| 6 (Medium) — fatal helper bypassed | `rethrowIfFatal` is now the first statement in all four catch-alls you named | `09055eccb` |
| 8 (Medium) — `777` / red check | test split: in-range delivers, out-of-range is refused and surfaced | edge-test `d0873967` |
| 2 (High) — construction rollback | one construction scope, reverse-order unwind, `catch (Throwable)` | `05f9c2378` |
| 5 (Medium) — stalled poll reads healthy | 5-minute no-value deadline, cleared only by a published reading | `a59d72a7b` |
| 7 (Medium) — `$share` | refused outright | `40eb895a0` |
| gap 2 | subscribable tag-set reload regression tests | `e012de377` |
| gap 3 | real file-watcher E2E | edge-test `546703b7` |
| gap 6, 7 | construction fault injection; fatal-throwable policy | `05f9c2378`, `09055eccb` |
| hygiene | `EdgeCaseValidationEndToEndTest` javadoc | edge-test `790f71d9` |

Your finding 3 is wider than scoped: `DataPointListBuilderImpl` → `DataPointWithMetadata` is the whole `DataPointBuilder` path, so **OPC-UA and EtherIP CIP/ODVA** were both affected on any ranged scalar schema. The wrapper-route tests now use the production builder rather than a flag on a mock.

Three dispositions:

**Gate 1 — the Critical stays deferred, and I think that is safe.** The V2 southbound path is reachable only through explicit southbound mappings, `initiative/nevsky-protocol-adapter-v2` is not a release branch, and EDG-813 replaces the bridge before anything is cut from it. The exposure is initiative-internal for as long as this is unmerged.

**Gate 8 — EDG-813 follows, it does not block.** Rebasing 813 onto the current initiative produces four independent collisions, each one an EDG-824 enforcement newer initiative code does not know about, so landing 824 first is what stops that repeating. EDG-829 / EDG-830 / EDG-831 stay linked as accepted follow-up scope.

**Finding 2 — half of it is a deliberate accepted residual.** Adapter *construction* is fixed. The tags-only reload multi-component commit is not, and I do not intend to fix it here: EDG-813 already rewrote those exact lines and turns the problem into plain statement order on one thread. Fixing it in 824 would be discarded work and would re-conflict on every 813 rebase.

Your finding 7 turned out not to be interim work — EDG-813 does not touch `SouthboundMappingEntity`, and its intake mints its own share name from the adapter id, which doubles as the marker keeping a durable command queue from being reaped. So the share name is Edge's on both branches and the rejection is permanent. Note this is a behaviour change: a config using `$share` previously loaded clean and silently dropped every write.

On finding 6, one consequence worth your view: with the wrapper now rethrowing, a *transient* `OutOfMemoryError` kills the dispatch thread while REST keeps serving the last snapshot — the green-while-dead shape #7 was about. I implemented your prescription literally. If you prefer the fallback ERROR snapshot published before the rethrow in `publishSnapshot`, that is a one-line reorder.

Still open: findings 1 and 4 and test-gaps 1 and 4 (EDG-813), and the QA tripwire acceptance run, which Estefania has parked pending this merge. The EDG-824 description has been restructured into fixed / partially fixed / deferred / tracked as you asked.

GH Actions is green on the last three heads (5,750 tests, 0 failures on `40eb895a`); the composite check on `80076c228` still needs confirming.

hivemq-edge/src/main/java/com/hivemq/protocols/v2/southbound/WritePayload.java
marregui: Agreed and done in the round-1 batch — `e83d4d08e` renamed `V2WritePayload` to `WritePayload`.

One left that your comment implies: `V2ConfigEntity`, in the same package. That one I would keep, since its name mirrors the `<v2>` XML element it binds rather than the package it sits in — but say the word and I will rename it too.

hivemq-edge/src/main/java/com/hivemq/protocols/v2/config/ProtocolAdapterEntity.java
marregui: Moved in `80076c228` — but as `TagEntity.validateAccessContradictions(validationEvents, adapterId)` rather than inside `validate`.

You are right that the logic is tag-local; every predicate reads that one tag. It could not go into `validate` itself because the warnings name the adapter, and `EntityValidatable` fixes that signature at `validate(List<ValidationEvent>)`. Dropping the adapter id would leave `tag [temperature] is read-activated but…` with no way to tell which adapter it came from once two of them declare the same tag name, and every other message in this validator names the adapter. So it is a sibling method taking the id as a parameter, called from the loop that already runs `tag.validate`.

Also worth flagging: the block had no test coverage at all, so the move was unverified by definition. Added four cases — read-contradiction, write-contradiction, the permitted case, and omitted `<access>` being unrestricted rather than deny-all.

hivemq-edge/src/main/java/com/hivemq/protocols/v2/wrapper/ProtocolAdapterWrapper.java
```diff
@@ -55,6 +56,9 @@ public final class ProtocolAdapterWrapper implements MessageHandler<ProtocolAdap
      * @param snapshot the reference the wrapper publishes its status into — created by the manager's handle in
      *                 production, by the test fixture in unit tests.
      */
+    private static final @NotNull org.slf4j.Logger log =
```
caoccao: We don't use FQN here.

caoccao (changes requested):
I left one comment and round 3 review doc https://linear.app/hivemq/document/sam-code-review-2026-07-31-9feda74c1025

hivemq-edge/src/main/java/com/hivemq/protocols/v2/tag/TagAspectRead.java
```diff
@@ -483,7 +704,10 @@ private boolean holdsSubscription() {
 
     private void scheduleTimer(final long delayMillis, final @NotNull Runnable onFire) {
         cancelActiveTimer();
-        activeTimer = timers.schedule(clock.nowMillis() + delayMillis, onFire);
+        // Saturate: a near-Long.MAX_VALUE configured delay must mean "practically never", not an overflowed
+        // negative deadline that fires immediately.
+        final long fireAtMillis = clock.nowMillis() + delayMillis;
+        activeTimer = timers.schedule(fireAtMillis < 0 ? Long.MAX_VALUE : fireAtMillis, onFire);
```
caoccao, line 715: Why do we set a timer with `Long.MAX_VALUE`? Why don't we just ignore the negative milliseconds?

```java
import java.util.Timer;
import java.util.TimerTask;

public class Main {
    public static void main(String[] args) {
        Timer timer = new Timer();
        TimerTask task = new TimerTask() {
            @Override
            public void run() {
                System.out.println("Executing task!");
            }
        };

        // Throws java.lang.IllegalArgumentException: Illegal execution time.
        timer.schedule(task, Long.MAX_VALUE); 
    }
}
```

caoccao (changes requested):
Only one question.

hivemq-edge/src/main/java/com/hivemq/protocols/v2/tag/TagAspectRead.java
marregui: `timers` here isn't `java.util.Timer` — it's the actor's `PriorityTimerQueue`, and its `schedule(long fireAtMillis, Runnable)` takes an **absolute** fire time in `Clock.nowMillis()` units, not a delay. It does no argument validation; `fireDue(nowMillis)` pops entries while `fireAtMillis <= nowMillis`, so `Long.MAX_VALUE` is a legal deadline that is simply never due — "practically never", no exception.

On ignoring the negative: the negative isn't a configured negative delay, it's the wrap-around of `now + delayMillis`. `TagEntity.validate` only enforced `poll-interval-millis > 0`, so `<poll-interval-millis>9223372036854775807</poll-interval-millis>` was accepted; `now + that` goes negative, the deadline lands in the past, and the aspect polls on every tick — the slowest configurable cadence behaving as the fastest. Skipping the schedule entirely would also never fire, but it leaves the aspect in `WAITING_FOR_POLL_INTERVAL` with an empty timer slot, so `timers.size()` and any "no pending timer" diagnostic would read that state as un-armed; one armed timer per `scheduleTimer` call keeps that invariant uniform.

That said, you're right that the actor is the wrong place for the guarantee, so I've pushed the bound: `TagEntity.MAXIMUM_POLL_INTERVAL_MILLIS` (7 days) is now enforced in `validate`, which is orders of magnitude beyond any industrial cadence and keeps the sum far from overflow. The clamp stays as defence in depth for a delay that did not come through that validation. Tests added in 8163717:

- `ProtocolAdapterEntityTest.pollIntervalBeyondTheMaximum_isRejected` / `nonPositivePollInterval_isRejected` — the bound, its boundary value, and 0/-1.
- `TagAspectReadPolledTest.anOverflowingPollInterval_neverProducesADeadlineInThePast` — `Long.MAX_VALUE` interval with the clock advanced before activation (at `now == 0` the sum does not overflow and the guard is never exercised); the aspect stays in `WAITING_FOR_POLL_INTERVAL`, no `pollBatch`, no data points. Mutation-checked: it fails with the clamp removed.

caoccao (approved):
LGTM

PR #1703, opened 2026-07-28
EDG-844: Paginate Data Hub listings so all items are shown

mschoenert-hivemq:
**Motivation**

Resolves EDG-844.

On the Data Hub UI, once a customer has more than the default page size (~50) of scripts, policies, or schemas, newly created items stop appearing in the list. Reported at Knorr-Bremse: newly created scripts stopped showing past ~47–50; deleting older ones brought them back. The items were created successfully and were retrievable via the REST API — only the UI failed to list them.

**Root cause**

The Data Hub list endpoints are cursor-paginated (`limit` 10–500, default **50**; response is `{ items, _links.next }`). The frontend list hooks did a single `useQuery` fetch of the first page and never followed `_links.next`, so anything beyond the first page was never fetched → invisible in the UI. Deleting items back under 50 made them reappear (the "delete-to-restore" symptom).

**Changes**

- Add a shared `usePaginatedList` helper (`src/extensions/datahub/api/hooks/usePaginatedList.ts`): `useInfiniteQuery` that follows `_links.next` (extracting the `cursor` param from the next-page URL) until exhausted, with a `select` that flattens all pages back into the single `{ items }` shape the callers already consume.
- Route the four listing hooks through it: `useGetAllScripts`, `useGetAllSchemas`, `useGetAllDataPolicies`, `useGetAllBehaviorPolicies`.
- No component or backend changes — the tables keep calling the hooks the same way and receive the full list.

**Non-goals**

- Backend changes (the API is correct and already paginated).
- User-visible pagination controls — the lists should just show everything.

**Tests**

- New `usePaginatedList.spec.ts`: proves the cursor is followed across pages (page 1 with cursor + page 2 without → all items returned) and unit-tests `extractCursor`.
- Existing DataHub hook specs and the four table Cypress component specs (ScriptTable, PolicyTable, SchemaTable, DataHubListings) remain green.

Short-term workaround for affected customers: use the REST API with `?limit=500`.

### DIFF
--- a/hivemq-edge-frontend/src/extensions/datahub/api/hooks/DataHubBehaviorPoliciesService/useGetAllBehaviorPolicies.ts
+++ b/hivemq-edge-frontend/src/extensions/datahub/api/hooks/DataHubBehaviorPoliciesService/useGetAllBehaviorPolicies.ts
@@ -1,27 +1,21 @@
-import { useQuery } from '@tanstack/react-query'
 import { DATAHUB_QUERY_KEYS } from '../../utils.ts'
 import { useHttpClient } from '@/api/hooks/useHttpClient/useHttpClient.ts'
-import type { ApiError, BehaviorPolicyList } from '@/api/__generated__'
+import type { BehaviorPolicy } from '@/api/__generated__'
+import { usePaginatedList } from '@datahub/api/hooks/usePaginatedList.ts'
 
 interface GetAllBehaviorPoliciesProps {
   fields?: string
   policyIds?: string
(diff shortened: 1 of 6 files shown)


caoccao:
@codex Have a code review on this PR.

mschoenert-hivemq:
Reverting the limit=500. It breaks many, many tests. That is a shortcoming of those tests. But it is just not worth fixing the tests here for something that is a not very important optimization.

caoccao (approved):

PR #1704, opened 2026-07-29
EDG-847: refactor: fix all ErrorProne and javac warnings

caoccao:
Fixes every warning reported by the core, adapter-module and SDK compile tasks. Part of [EDG-847](https://linear.app/hivemq/issue/EDG-847/errorprone-fix-2026-07-29); the commercial-module half is in hivemq/hivemq-edge-commercial-modules.

## Verification

- All listed `compileJava` / `compileTestJava` tasks re-run with `--rerun-tasks --no-build-cache`: **0 warnings, 0 errors**.
- `:hivemq-edge:test`, `:hivemq-edge-module-opcua:test`, `:hivemq-edge-module-mtconnect:test`, `:hivemq-edge-module-plc4x:test` all green.
- `spotlessApply` is a no-op on this branch.

## Cross-cutting policies applied

| Check | Count | Fix |
| --- | --- | --- |
| `DefaultCharset` | 86 (whole task set) | explicit `StandardCharsets.UTF_8` |
| `StringSplitter` | 10 (whole task set) | `split(x, -1)` |
| `EqualsGetClass` | 20 (whole task set) | class marked `final` + `instanceof` pattern |
| `MutablePublicArray` | 27 (whole task set) | visibility narrowed, or `clone()` accessor |

For `EqualsGetClass` every affected class was checked for subclasses first (none had any), so `final` + `instanceof` is provably equivalent to the previous `getClass()` comparison.

## Two latent bugs the checks surfaced

**`DeviceTagValidatorTest`** — 11 assertions compared a `ValidationError.Code` enum against a `String` literal:

```java
assertThat(errors).noneMatch(e -> e.code().equals("DUPLICATE_NODE"));
```

`equals` was always `false`, so every one of these `noneMatch` assertions passed vacuously. They now compare enum constants (`e.code() == ValidationError.Code.DUPLICATE_NODE`), matching the idiom already used elsewhere in the same file, and all 11 pass for real.

**MTConnect adapter** — `MtConnectProtocolAdapter` threw `javax.management.modelmbean.XMLParseException`, a JMX ModelMBean class that has nothing to do with MTConnect XML and is deprecated for removal in the JDK. Replaced with a module-local `MtConnectXmlParseException` (the only new file). One test assertion drops the `"XML Parse Exception: "` prefix that the JMX class injected into its message.

## Other changes worth a look

- **`CustomConfigSchemaGenerator`** — victools `JacksonModule` → `JacksonSchemaModule`. The former is `@Deprecated(forRemoval = true, since = "5.0.0")` and is a trivial subclass of the latter with identical constructors, so this is a drop-in swap.
- **`DataPointWithMetadata`** — field `adapterID` renamed to `adapterId` (`InconsistentCapitalization`; the builder parameter already used the lowercase spelling).
- **`OpcUaNodeBrowser.estimateSize()`** — `int` arithmetic was assigned to a `long` return; now widened before the subtraction (`IntLongMath`).
- **`FutureReturnValueIgnored`** in four concurrency tests — the ignored futures use ErrorProne's documented `var unused = ...` idiom. Each lambda already catches its own exceptions and reports through a latch/counter, so the future genuinely carries nothing.
- **5 module `build.gradle.kts`** gain `testCompileOnly(libs.jaxb4.bind)`. `hivemq-edge` declares JAXB as `implementation`, so adapter modules could not resolve the annotations on its JAXB-annotated config entities and javac emitted `unknown enum constant XmlAccessType.NONE` on every test compile.

## Behaviour note

`split(x, -1)` keeps trailing empty segments where the single-argument form dropped them. The only site where that is observable is the test-scope `BasicAuthenticationHandler.parseValue`: a header decoding to `"user:"` now reaches the credentials provider with an empty password instead of being short-circuited. The provider still rejects it.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
@@ -293,7 +293,8 @@ public class OidcServiceImpl implements OidcService {
      * Redirects are disabled: the discovery and JWKS URLs are derived from the validated issuer and answer
      * directly, so following a redirect would only widen the set of hosts Edge contacts (an SSRF surface)
      * and let a chain multiply the per-call timeout. {@link DefaultResourceRetriever} has no redirect
-     * setting, so {@link #openConnection} is overridden to turn instance-following off on the connection.
+     * setting, so {@link DefaultResourceRetriever#openConnection(URL)} is overridden to turn instance-following
+     * off on the connection.
      */
     static @NotNull DefaultResourceRetriever resourceRetriever(final @NotNull OidcConfiguration config) {
         final int timeout = config.getConnectionTimeoutMillis();
--- a/hivemq-edge/src/main/java/com/hivemq/api/json/CustomConfigSchemaGenerator.java
+++ b/hivemq-edge/src/main/java/com/hivemq/api/json/CustomConfigSchemaGenerator.java
@@ -31,8 +31,8 @@ import com.github.victools.jsonschema.generator.SchemaGeneratorConfigBuilder;
 import com.github.victools.jsonschema.generator.SchemaGeneratorConfigPart;
 import com.github.victools.jsonschema.generator.SchemaVersion;
 import com.github.victools.jsonschema.generator.TypeScope;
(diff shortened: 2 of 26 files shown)


marregui (approved):
LGTM

PR #1705, opened 2026-07-30
EDG-845: Separate northbound (read) and southbound (write) tag schemas

mschoenert-hivemq:
**Motivation**

[EDG-845](https://linear.app/hivemq/issue/EDG-845/separate-northbound-read-and-southbound-write-tag-schemas)

A tag has one schema today, used for both directions. It wraps the value in an envelope (`tagName`, `timestamp`, `metadata`, `context`) that can never be written, and marks per-field writability with flags. The southbound mapping editor is handed that schema and has to neutralise most of it in the UI — the screenshot before this change shows one settable property and seven marked *Read-only* or *Not supported*. Meanwhile the same schema is shown in the tag inspection panel behind a button labelled "View the WRITE schema" while its own helper text calls it a valid *source* for transformation.

This also unblocks OPC-UA Alarms & Conditions ([EDG-835](https://linear.app/hivemq/issue/EDG-835/implement-opc-ua-alarms-and-conditions-support-in-edge)): a condition tag reads an alarm event but writes `{eventId, method, comment}`, so one schema plus flags cannot represent it.

Supersedes the never-done frontend cleanup in EDG-59 by solving it on the backend side — the read-only fields are no longer sent to the write form, so there is nothing for the UI to hide.

**Changes**

Backend — one seam, no adapter touched:

- `TagSchemaCreationOutputImpl` now assembles the schema per direction, because it owns the `DataPointSchema`. READ keeps the envelope and the full value; WRITE drops the non-writable envelope and uses the adapter's explicit `writeSchema` when present, else the plain value.
- The future carries the raw `DataPointSchema` rather than a pre-rendered JSON node, so the direction is chosen at the consuming edge.
- The REST endpoint gains an optional `direction` query parameter. It has no default in the spec, so a read caller sends no parameter and produces exactly the URL it did before — existing callers, intercepts and caches are unaffected.

Frontend — each consumer asks for the direction it needs:

- Southbound mapping editor and its destination data model → WRITE.
- Data combiner and other source-side consumers → unchanged (read).
- Tag inspection panel → both, compared: a single "Current schema" when identical, a labelled "Read schema" / "Write schema" pair when they differ.

The write schema does not hide fields that cannot be written; it drops the non-writable envelope, and per-field write permission stays visible as the existing read-only marker. A statically derived write schema cannot express write validity correctly — an array of read-only items admits only `[]`, a required read-only member makes the object unsatisfiable — so enforcement belongs at validation time and is left to Nevsky.

Separately, the second commit restores `TagSchemaCreationOutputImpl.getFuture()` as a deprecated method returning the read schema. Removing it in the first commit broke callers outside this repository — the commercial bidirectional adapter and three integration tests in `hivemq-edge-test` — which surfaced as a `NoSuchMethodError` during `startWriting`. Because adapters start in sequence, that aborted the loop and every adapter after the first writing one silently failed to start. The shim returns the read schema, which is what the method always produced; migrating those callers, and deciding whether the commercial module wants WRITE there, belongs in their own repositories.

**Verification**

- Java: 35 tests (schema + simulation), including new coverage that WRITE drops the envelope, READ keeps it, and an explicit `writeSchema` is used instead of the value. Mutation-checked — reverting either behaviour fails the tests.
- Frontend: 30 Cypress component tests across all specs touching the schema endpoint; `pnpm lint:all` clean.
- Manually against a live OPC-UA device: `Test` (writable node) shows an unlocked `VALUE` in both panels, while `sim-sqr` (read-only signal node) shows it locked — the per-field AccessLevel is real, not decoration.

Requires hivemq/hivemq-edge-adapter-sdk#132 (same branch name).

### DIFF
--- a/hivemq-edge-frontend/src/components/rjsf/MqttTransformation/components/MappingEditor.tsx
+++ b/hivemq-edge-frontend/src/components/rjsf/MqttTransformation/components/MappingEditor.tsx
@@ -36,8 +36,9 @@ const MappingEditor: FC<MappingEditorProps> = ({
 }) => {
   const { t } = useTranslation('components')
   // The properties listed here are the destination of a southbound mapping, i.e. what the user may write to.
-  // The WRITE schema contains only those, so the list no longer has to mark most entries as read-only.
-  const { data, isLoading, isError, error, isSuccess } = useGetSchema(adapterId, topic, 'WRITE')
+  // The SOUTHBOUND schema drops the non-writable envelope; any remaining read-only fields inside the value
+  // are hidden by MappingInstructionList (EDG-59).
+  const { data, isLoading, isError, error, isSuccess } = useGetSchema(adapterId, topic, 'SOUTHBOUND')
 
   const properties = useMemo(() => {
     const allProperties = data ? getPropertyListFrom(data) : []
--- a/hivemq-edge-frontend/src/components/rjsf/MqttTransformation/components/MappingEditor.tsx
+++ b/hivemq-edge-frontend/src/components/rjsf/MqttTransformation/components/MappingEditor.tsx
@@ -35,7 +35,9 @@ const MappingEditor: FC<MappingEditorProps> = ({
   ...props
 }) => {
   const { t } = useTranslation('components')
-  const { data, isLoading, isError, error, isSuccess } = useGetSchema(adapterId, topic)
+  // The properties listed here are the destination of a southbound mapping, i.e. what the user may write to.
+  // The WRITE schema contains only those, so the list no longer has to mark most entries as read-only.
+  const { data, isLoading, isError, error, isSuccess } = useGetSchema(adapterId, topic, 'WRITE')
 
   const properties = useMemo(() => {
     const allProperties = data ? getPropertyListFrom(data) : []
--- a/hivemq-edge-frontend/src/modules/Device/components/TagSchemaPanel.spec.cy.tsx
+++ b/hivemq-edge-frontend/src/modules/Device/components/TagSchemaPanel.spec.cy.tsx
@@ -10,12 +10,12 @@ import { TagSchemaPanel } from './TagSchemaPanel'
 
 const mocTag = MOCK_DEVICE_TAGS('opcua-1', MockAdapterType.OPC_UA)[0]
 
-// The mocks mirror the real backend documents: the read (northbound) document wraps the value in the
-// non-writable envelope, the write (southbound) document is value-only. The backend guarantees the value
-// sub-schemas are identical for a plain value tag (pinned in TagSchemaCreationOutputImplSchemaBuilderTest).
+// The mocks mirror the real backend documents: the northbound document wraps the value in the non-writable
+// envelope, the southbound document is value-only. The backend guarantees the value sub-schemas are
+// identical for a plain value tag (pinned in TagSchemaCreationOutputImplSchemaBuilderTest).
 const MOCK_VALUE_SCHEMA = GENERATE_DATA_MODELS(true, 'test')
 
-const MOCK_READ_SCHEMA: RJSFSchema = {
+const MOCK_NORTHBOUND_SCHEMA: RJSFSchema = {
   $schema: 'https://json-schema.org/draft/2019-09/schema',
   title: 'test',
   type: 'object',
@@ -27,7 +27,7 @@ const MOCK_READ_SCHEMA: RJSFSchema = {
   },
 }
 
-const MOCK_WRITE_SCHEMA: RJSFSchema = {
+const MOCK_SOUTHBOUND_SCHEMA: RJSFSchema = {
   $schema: 'https://json-schema.org/draft/2019-09/schema',
   title: 'test',
   type: 'object',
@@ -52,13 +52,13 @@ const reverseKeyOrder = (value: unknown): unknown => {
   )
 }
 
-const MOCK_WRITE_SCHEMA_REORDERED: RJSFSchema = {
-  ...MOCK_WRITE_SCHEMA,
+const MOCK_SOUTHBOUND_SCHEMA_REORDERED: RJSFSchema = {
+  ...MOCK_SOUTHBOUND_SCHEMA,
   properties: { value: reverseKeyOrder(MOCK_VALUE_SCHEMA) as RJSFSchema },
 }
 
-// A tag whose write shape is not a projection of its read shape, e.g. an OPC-UA condition tag.
-const MOCK_WRITE_SCHEMA_INDEPENDENT: RJSFSchema = {
+// A tag whose southbound shape is not a projection of its northbound shape, e.g. an OPC-UA condition tag.
+const MOCK_SOUTHBOUND_SCHEMA_INDEPENDENT: RJSFSchema = {
   $schema: 'https://json-schema.org/draft/2019-09/schema',
   title: 'test',
   type: 'object',
@@ -80,18 +80,19 @@ const MOCK_WRITE_SCHEMA_INDEPENDENT: RJSFSchema = {
 describe('TagSchemaPanel', () => {
   beforeEach(() => {
     cy.viewport(800, 800)
-    // The read request carries no `direction` (it is the default); the write request is intercepted per test.
-    cy.intercept('/api/v1/management/protocol-adapters/schema/**', MOCK_READ_SCHEMA)
+    // The northbound request carries no `direction` (it is the default); the southbound request is
+    // intercepted per test.
+    cy.intercept('/api/v1/management/protocol-adapters/schema/**', MOCK_NORTHBOUND_SCHEMA)
   })
 
-  const interceptWriteSchema = (response: RJSFSchema | { statusCode: number }) =>
+  const interceptSouthboundSchema = (response: RJSFSchema | { statusCode: number }) =>
     cy.intercept(
       { method: 'GET', pathname: '**/protocol-adapters/schema/**', query: { direction: 'SOUTHBOUND' } },
       'statusCode' in response ? response : { body: response }
     )
 
   it('should render properly', () => {
-    interceptWriteSchema(MOCK_WRITE_SCHEMA)
+    interceptSouthboundSchema(MOCK_SOUTHBOUND_SCHEMA)
     cy.mountWithProviders(<TagSchemaPanel adapterId="test" tag={mocTag} />)
 
     cy.getByTestId('tag-schema-header').should('have.text', 'Tag')
@@ -100,83 +101,84 @@ describe('TagSchemaPanel', () => {
       cy.get('label').should('have.text', 'Current schema')
       cy.get('h3').should('have.text', 'test')
       cy.get('[role="list"] li').should('length', 11)
-      cy.get('#tag-schema-panel-helptext').should('have.text', 'Reading and writing this tag use the same schema.')
+      cy.get('#tag-schema-panel-helptext').should('have.text', 'Both directions use the same schema for this tag.')
     })
     cy.getByTestId('tag-schema-download').should('have.text', 'Download the schema').should('not.be.disabled')
   })
 
-  it('should show a single schema when the read and write value shapes are identical', () => {
-    interceptWriteSchema(MOCK_WRITE_SCHEMA)
+  it('should show a single schema when the northbound and southbound value shapes are identical', () => {
+    interceptSouthboundSchema(MOCK_SOUTHBOUND_SCHEMA)
     cy.mountWithProviders(<TagSchemaPanel adapterId="test" tag={mocTag} />)
 
     // The documents differ (envelope vs value-only) but the value shapes match, so only one panel renders.
-    cy.getByTestId('tag-schema-read').should('be.visible')
+    cy.getByTestId('tag-schema-northbound').should('be.visible')
     cy.getByTestId('tag-schema-panel').find('label').should('have.text', 'Current schema')
-    cy.getByTestId('tag-schema-panel-write').should('not.exist')
+    cy.getByTestId('tag-schema-panel-southbound').should('not.exist')
   })
 
   it('should show a single schema when the value shapes differ only in member order', () => {
-    interceptWriteSchema(MOCK_WRITE_SCHEMA_REORDERED)
+    interceptSouthboundSchema(MOCK_SOUTHBOUND_SCHEMA_REORDERED)
     cy.mountWithProviders(<TagSchemaPanel adapterId="test" tag={mocTag} />)
 
     // Same shape, different insertion order — a stringify-based comparison would split this into two panels
-    // and tell the user the tag reads and writes differently, which is false.
-    cy.getByTestId('tag-schema-read').should('be.visible')
+    // and tell the user the two directions differ, which is false.
+    cy.getByTestId('tag-schema-northbound').should('be.visible')
     cy.getByTestId('tag-schema-panel').find('label').should('have.text', 'Current schema')
-    cy.getByTestId('tag-schema-panel-write').should('not.exist')
+    cy.getByTestId('tag-schema-panel-southbound').should('not.exist')
   })
 
-  it('should show both schemas when the write shape is independent of the read shape', () => {
-    interceptWriteSchema(MOCK_WRITE_SCHEMA_INDEPENDENT)
+  it('should show both schemas when the southbound shape is independent of the northbound shape', () => {
+    interceptSouthboundSchema(MOCK_SOUTHBOUND_SCHEMA_INDEPENDENT)
     cy.mountWithProviders(<TagSchemaPanel adapterId="test" tag={mocTag} />)
 
     cy.getByTestId('tag-schema-panel').within(() => {
-      cy.get('label').should('have.text', 'Read schema')
+      cy.get('label').should('have.text', 'Northbound schema')
       cy.get('#tag-schema-panel-helptext').should(
         'have.text',
         'The data published for this tag. A valid source for data transformation.'
       )
     })
 
-    cy.getByTestId('tag-schema-panel-write').within(() => {
-      cy.get('label').should('have.text', 'Write schema')
-      cy.get('#tag-schema-panel-write-helptext').should(
+    cy.getByTestId('tag-schema-panel-southbound').within(() => {
+      cy.get('label').should('have.text', 'Southbound schema')
+      cy.get('#tag-schema-panel-southbound-helptext').should(
         'have.text',
-        'The shape of a write to this tag. Whether an individual field can be written is shown per field.'
+        'The shape of a southbound message to this tag. Whether an individual field can be written is shown per field.'
       )
     })
   })
 
-  it('should render the read schema without waiting for a slow write request', () => {
-    // The two queries are independent: the read panel must appear while the write request is still in flight.
+  it('should render the northbound schema without waiting for a slow southbound request', () => {
+    // The two queries are independent: the northbound panel must appear while the southbound request is
+    // still in flight.
     cy.intercept(
       { method: 'GET', pathname: '**/protocol-adapters/schema/**', query: { direction: 'SOUTHBOUND' } },
-      { body: MOCK_WRITE_SCHEMA, delay: 8000 }
+      { body: MOCK_SOUTHBOUND_SCHEMA, delay: 8000 }
     )
     // Distinct adapterId: the query client is shared across tests, and this test deliberately leaves an
-    // 8s-pending write request behind — a colliding query key would leak it into the next test.
+    // 8s-pending southbound request behind — a colliding query key would leak it into the next test.
     cy.mountWithProviders(<TagSchemaPanel adapterId="test-slow" tag={mocTag} />)
 
-    // Asserted well inside the 8s write delay (default 4s command timeout): the read panel is up while the
-    // write query still loads.
-    cy.getByTestId('tag-schema-read').should('be.visible')
-    cy.getByTestId('tag-schema-panel').find('label').should('have.text', 'Read schema')
+    // Asserted well inside the 8s southbound delay (default 4s command timeout): the northbound panel is up
+    // while the southbound query still loads.
+    cy.getByTestId('tag-schema-northbound').should('be.visible')
+    cy.getByTestId('tag-schema-panel').find('label').should('have.text', 'Northbound schema')
   })
 
-  it('should still render the read schema and an error when the write schema cannot be loaded', () => {
-    interceptWriteSchema({ statusCode: 500 })
+  it('should still render the northbound schema and an error when the southbound schema cannot be loaded', () => {
+    interceptSouthboundSchema({ statusCode: 500 })
     cy.mountWithProviders(<TagSchemaPanel adapterId="test" tag={mocTag} />)
 
-    // The write failure must be visible, not silently swallowed into a half-answer.
-    cy.getByTestId('tag-schema-read').should('be.visible')
-    cy.getByTestId('tag-schema-panel').find('label').should('have.text', 'Read schema')
-    cy.get('[role="alert"]').should('contain.text', 'Cannot load the write schema')
-    cy.getByTestId('tag-schema-panel-write').should('not.exist')
+    // The southbound failure must be visible, not silently swallowed into a half-answer.
+    cy.getByTestId('tag-schema-northbound').should('be.visible')
+    cy.getByTestId('tag-schema-panel').find('label').should('have.text', 'Northbound schema')
+    cy.get('[role="alert"]').should('contain.text', 'Cannot load the southbound schema')
+    cy.getByTestId('tag-schema-panel-southbound').should('not.exist')
   })
 
   it('should be accessible', () => {
     cy.injectAxe()
-    interceptWriteSchema(MOCK_WRITE_SCHEMA)
+    interceptSouthboundSchema(MOCK_SOUTHBOUND_SCHEMA)
     cy.mountWithProviders(<TagSchemaPanel adapterId="test" tag={mocTag} />)
 
     cy.checkAccessibility()
--- a/hivemq-edge-frontend/src/modules/Device/components/TagSchemaPanel.spec.cy.tsx
+++ b/hivemq-edge-frontend/src/modules/Device/components/TagSchemaPanel.spec.cy.tsx
@@ -22,12 +22,46 @@ describe('TagSchemaPanel', () => {
       cy.get('label').should('have.text', 'Current schema')
       cy.get('h3').should('have.text', 'test')
       cy.get('[role="list"] li').should('length', 8)
-      cy.get('#tag-schema-panel-helptext').should(
-        'have.text',
-        'This schema will be a valid source for data transformation'
-      )
+      cy.get('#tag-schema-panel-helptext').should('have.text', 'Reading and writing this tag use the same schema.')
     })
     cy.getByTestId('tag-schema-download').should('have.text', 'Download the schema').should('not.be.disabled')
+  })
+
+  it('should show a single schema when reading and writing are identical', () => {
+    cy.mountWithProviders(<TagSchemaPanel adapterId="test" tag={mocTag} />)
+
+    // Both directions return the same mock, so only the read panel is rendered.
+    cy.getByTestId('tag-schema-read').should('be.visible')
+    cy.getByTestId('tag-schema-panel-write').should('not.exist')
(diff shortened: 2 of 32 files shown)


hivemq-edge-openapi/openapi/paths/api_v1_management_protocol-adapters_schema_{adapterId}_{tagName}.yaml
```diff
@@ -23,6 +23,18 @@ get:
       schema:
         type: string
         format: urlencoded
+    - description: >-
+        The direction of the schema to retrieve. WRITE returns the southbound schema describing only what can
+        be written to the tag (read-only fields omitted). When omitted, the northbound schema describing the
+        full data shape published for the tag is returned.
```
marregui: **blocker — this says the opposite of what the code does.**

`TagSchemaCreationOutputImpl.writeSchema` does not omit read-only fields. It drops the envelope and keeps the value as-is; its own javadoc says so and gives the right reason (an array of read-only items admits only `[]`, a required read-only member makes the object unsatisfiable).

This is the customer-facing contract, and it's the source `ext/hivemq-edge-openapi.yaml` and `ProtocolAdaptersService.ts` are generated from — fixing it here fixes all three on the next regen.

```suggestion
    - description: >-
        The direction of the schema to retrieve. WRITE returns the southbound schema: the non-writable
        envelope (tagName, timestamp, metadata, context) is dropped, leaving the value shape that a write
        targets. Whether an individual field can be written is carried per field as readOnly and enforced
        when the write is validated. When omitted, the northbound schema describing the full data shape
        published for the tag is returned.
```

hivemq-edge/src/main/java/com/hivemq/protocols/tag/TagSchemaCreationOutputImpl.java
```diff
@@ -17,33 +17,133 @@
...
+     * Nevsky. Only the non-writable envelope is stripped here.
+     */
+    private static @NotNull Schema writeSchema(final @NotNull DataPointSchema tagSchemas) {
+        final Schema writeValue =
+                tagSchemas.writeSchema() != null ? tagSchemas.writeSchema() : tagSchemas.valueSchema();
+
+        final var builder = new SchemaBuilder().startObject();
+        builder.property("value").required().schema(writeValue).readable(false).writable();
+        return builder.endObject().build();
     }
```
marregui: **blocker — read and write can never be equal, so the frontend's identical-path is dead.**

`readSchema` unconditionally emits `tagName` and `timestamp` (L102-103); this emits `{value}` only, and marks it `writeOnly: true`. For **every** tag, in every adapter, the two documents differ — there is no input that makes them equal.

`TagSchemaPanel` compares them with `JSON.stringify` to decide one panel or two, so that comparison is always `false`. Net: the panel renders two schema browsers for every tag, `device.drawer.schema.label` and `.identical` are unreachable, and §3's "show one if identical" isn't achieved. Frontend side in the `TagSchemaPanel.tsx` thread.

nit, same block: `readable(false)` renders as `writeOnly: true`. The frontend drops it — `SCHEMA_SUPPORTED_PROPERTIES` doesn't list it — so no impact today. It does assert the value is not readable, two methods away from a read schema that says it is.

hivemq-edge/src/main/java/com/hivemq/protocols/tag/TagSchemaCreationOutputImpl.java
```diff
@@ -17,33 +17,133 @@
...
+                    .schema(tagSchemas.metadataSchema())
+                    .readable()
+                    .writable(false);
+        }
+        if (tagSchemas.context() != null) {
+            builder.property("context").schema(tagSchemas.context()).readable().writable(false);
+        }
+
+        return builder.endObject().build();
+    }
```
marregui: `readSchema` is a semantically identical copy of the now-deprecated `SchemaJsonRepresentation.toCompositeSchema` — same builder calls, same order. The SDK copy has to survive for adapters built against an older SDK, and the dependency runs the wrong way to delegate, so the two will drift. Silently: third-party adapters see the SDK copy, Edge serves this one.

fix: a test pinning `toCompositeSchema(dps)` equal to `getSchema(READ)`. Cheap, and it fails the day someone edits one side.

nit: `Direction` is REST-layer vocabulary now, so `ProtocolAdaptersResourceImpl` imports an adapter-plumbing impl class to name it. A standalone type — or the generated OpenAPI enum — costs nothing today.

hivemq-edge/src/main/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImpl.java
```diff
@@ -707,9 +707,15 @@ public int getDepth() {
...
     @Override
-    public @NotNull Response getSchema(final @NotNull String adapterId, final @NotNull String tagName) {
+    public @NotNull Response getSchema(
+            final @NotNull String adapterId, final @NotNull String tagName, final @Nullable String direction) {
         final String decodedTagName = URLDecoder.decode(tagName, StandardCharsets.UTF_8);
 
+        // READ (northbound) is the default; WRITE (southbound) returns only the writable fields.
+        final TagSchemaCreationOutputImpl.Direction schemaDirection = "WRITE".equalsIgnoreCase(direction)
+                ? TagSchemaCreationOutputImpl.Direction.WRITE
+                : TagSchemaCreationOutputImpl.Direction.READ;
```
marregui: `"WRITE".equalsIgnoreCase(direction)` makes everything that isn't `WRITE` mean `READ`. `direction=WRTIE`, `direction=southbound`, `direction=` all return 200 with the read schema — while `write` is accepted even though the spec's enum is uppercase-only. A caller who typos the parameter gets a differently-shaped document and no signal; it surfaces later as a mapping editor listing fields that can't be written.

The spec needs a `400` next to the `200` if you take this.

```suggestion
        // READ (northbound) is the default; WRITE (southbound) drops the non-writable envelope.
        final TagSchemaCreationOutputImpl.Direction schemaDirection;
        if (direction == null) {
            schemaDirection = TagSchemaCreationOutputImpl.Direction.READ;
        } else {
            try {
                schemaDirection = TagSchemaCreationOutputImpl.Direction.valueOf(direction);
            } catch (final IllegalArgumentException ignored) {
                return errorResponse(new BadRequestError("Unknown schema direction: " + direction));
            }
        }
```

separate, same file: `getWritingSchema` (L767) still 301s to `…/schema/{adapterId}/{tagName}` with no `direction`. Not a regression — but now that the parameter exists, the one endpoint whose name means "the schema for writing" is the one that returns the read schema. One `.queryParam("direction", "WRITE")` on the `UriBuilder`.

hivemq-edge-frontend/src/modules/Device/components/TagSchemaPanel.tsx
```diff
@@ -29,14 +30,23 @@ interface TagSchemaPanelProps {
...
+  // tag; they differ when a tag's write shape is not a projection of its read shape (e.g. a condition tag).
+  const { data: readSchema, isLoading: isLoadingRead, isError: isErrorRead } = useGetSchema(adapterId, tag.name)
+  const { data: writeSchema, isLoading: isLoadingWrite } = useGetSchema(adapterId, tag.name, 'WRITE')
   const { t } = useTranslation()
 
+  const isLoading = isLoadingRead || isLoadingWrite
+  const areIdentical = useMemo(
+    () => Boolean(readSchema && writeSchema && JSON.stringify(readSchema) === JSON.stringify(writeSchema)),
+    [readSchema, writeSchema]
+  )
```
marregui: this is always `false` — see the `TagSchemaCreationOutputImpl` thread. The read document always carries `tagName`/`timestamp`, the write document is `{value}` only, so `areIdentical` never fires: both `FormControl`s always render, and the `label` / `identical` strings added to `translation.json` are unreachable.

Two ways out. Compare the `value` sub-schemas rather than the whole documents — which is what "the same schema" means to someone looking at the panel, and keeps §3's intent. Or drop the branch and always label the two panels, which is honest about what the backend now returns. Either is fine; the current shape claims a behaviour it can't deliver.

hivemq-edge-frontend/src/modules/Device/components/TagSchemaPanel.tsx
```diff
@@ -50,18 +60,32 @@ export const TagSchemaPanel: FC<TagSchemaPanelProps> = ({ tag, adapterId }) => {
...
         {isLoading && <LoaderSpinner />}
 
-        {isError && <ErrorMessage message={t('device.errors.noSchemaLoaded')} />}
-        {data && (
-          <FormControl isInvalid={isError} data-testid="tag-schema-panel" id="tag-schema-panel">
-            <FormLabel>{t('device.drawer.schema.label')}</FormLabel>
-            <Box borderWidth={1} p={2}>
-              <JsonSchemaBrowser schema={data} hasExamples />
+        {isErrorRead && <ErrorMessage message={t('device.errors.noSchemaLoaded')} />}
+        {!isLoading && readSchema && (
```
marregui: `isError` is only captured for the read query. If the WRITE request 500s: `isLoading` goes false, `writeSchema` is undefined, `areIdentical` is false — so the user gets the read schema labelled "Read schema", no write panel, and no error. A silent half-answer.

Also `isLoading = isLoadingRead || isLoadingWrite`: a slow WRITE request now blanks a read panel that used to render as soon as its own request came back.

hivemq-edge-frontend/src/modules/Device/components/TagSchemaPanel.spec.cy.tsx
```diff
@@ -22,12 +22,46 @@ describe('TagSchemaPanel', () => {
...
+  })
+
+  it('should show a single schema when reading and writing are identical', () => {
+    cy.mountWithProviders(<TagSchemaPanel adapterId="test" tag={mocTag} />)
+
+    // Both directions return the same mock, so only the read panel is rendered.
+    cy.getByTestId('tag-schema-read').should('be.visible')
+    cy.getByTestId('tag-schema-panel-write').should('not.exist')
+    cy.getByTestId('tag-schema-panel').find('label').should('have.text', 'Current schema')
+  })
```
marregui: this test registers no WRITE intercept, so both directions fall through to the `beforeEach` wildcard and return the same body — which is the only reason `areIdentical` is ever true here. The backend can't produce that state (thread on `TagSchemaCreationOutputImpl`), so the test pins the mock, not the contract. `should render properly` (L25) now asserts the same helper text on the same basis.

Whichever way the identical-check goes, these two need to assert something reachable.

hivemq-edge/src/test/java/com/hivemq/schema/TagSchemaCreationOutputImplSchemaBuilderTest.java
```diff
@@ -75,6 +71,104 @@ void test_tagSchemaBuilder_object_completesWithJsonSchema()
...
+        final JsonNode result = output.getSchema(TagSchemaCreationOutputImpl.Direction.WRITE);
+
+        // Only the value survives; tagName / timestamp / metadata can never be written.
+        final JsonNode properties = result.get("properties");
+        assertThat(properties.has("value")).isTrue();
+        assertThat(properties.has("tagName")).isFalse();
+        assertThat(properties.has("timestamp")).isFalse();
+        assertThat(properties.has("metadata")).isFalse();
+        assertThat(properties.get("value").get("type").asText()).isEqualTo("integer");
+    }
```
marregui, line 105: good tests — envelope dropped, envelope kept, explicit `writeSchema` winning in both directions is exactly the set I'd want. Four gaps:

- `context` never appears in this file, so that branch of `readSchema`/`writeSchema` is unexercised. The `metadata` assertions above don't cover it.
- nothing pins `getFuture()` to READ. That's the compatibility guarantee the commercial bidirectional adapter leans on, and it's currently unpinned here.
- no coverage of the REST-layer `direction` parsing, including the invalid and lowercase cases.
- the three tests that moved off `get(1, SECONDS)` now block on `future.get()` and rely on the 30s `orTimeout` for liveness — a hang costs 30s instead of 1s.

The coverage bot has this class at 67.22%, −7.28%, which is roughly the first two.

hivemq-edge-frontend/src/components/rjsf/MqttTransformation/components/MappingEditor.tsx
```diff
@@ -35,7 +35,9 @@ const MappingEditor: FC<MappingEditorProps> = ({
   ...props
 }) => {
   const { t } = useTranslation('components')
-  const { data, isLoading, isError, error, isSuccess } = useGetSchema(adapterId, topic)
+  // The properties listed here are the destination of a southbound mapping, i.e. what the user may write to.
+  // The WRITE schema contains only those, so the list no longer has to mark most entries as read-only.
+  const { data, isLoading, isError, error, isSuccess } = useGetSchema(adapterId, topic, 'WRITE')
```
marregui: the comment says the list no longer has to mark entries read-only, but nothing was removed: `isReadOnly` and `filterReadOnlyInstructions` (`data-type.utils.ts`) and the per-field Read-only / Not-supported alerts (`MappingInstruction.tsx:126,146`) are all still there. Step 4 of the ticket is what closes EDG-59, and it's the deletion, not the `'WRITE'` argument.

Either finish it here or drop the EDG-59 claim — raised on the ticket as a scope call rather than a code one.

marregui (changes requested):
net: right design, two blockers. The SDK side (adapter-sdk#132) is the good part — approved separately.

What's right, so it doesn't get lost under the list:

- composition moved out of `SchemaJsonRepresentation` and into the owner of the `DataPointSchema`. That's the correct seam, and `toJsonSchemaDocument` is a clean one.
- backward compatibility is properly done: the 3-arg convenience constructor keeps the record source- and binary-compatible, `toCompositeSchema` and `getFuture()` survive as documented delegates. `hivemq-edge-test` (3 ITs) and the commercial bidirectional adapter are untouched, and all six in-repo `createTagSchema` implementations need no change — §3's "no adapter changes required" holds.
- **not** pruning read-only fields from the write value is better than what step 1 of the ticket asked for, and the javadoc's reason is sound. Worth recording on the ticket, because the ticket still promises the filter.

Blockers:

1. the `direction` description in the OpenAPI source states the opposite of what `writeSchema` does — and it's the source that `ext/hivemq-edge-openapi.yaml`, the JAX-RS interface and `ProtocolAdaptersService.ts` are all generated from. Suggestion inline.
2. read and write documents can never be equal — read always carries `tagName`/`timestamp`, write is `{value}` only — so `TagSchemaPanel`'s identical-check is dead code. Every tag now renders two schema browsers, the new `label`/`identical` strings are unreachable, and two of the three new Cypress tests assert a state the backend cannot produce. Threads on `TagSchemaCreationOutputImpl` and `TagSchemaPanel.tsx`.

The rest is inline: lenient `direction` parsing, the read envelope now living in two repos, the swallowed WRITE error, test gaps (the coverage bot has `TagSchemaCreationOutputImpl` at −7.28%).

Three non-code items are on EDG-845 rather than here: whether this closes EDG-59, the READ/WRITE vs NORTHBOUND/SOUTHBOUND naming your own ticket note argues for, and the merge order with adapter-sdk#132.

Reviewed at 9226c2db, by reading. No build, no test run.

hivemq-edge-openapi/openapi/paths/api_v1_management_protocol-adapters_schema_{adapterId}_{tagName}.yaml
marregui: fixed in 3d4d6e20a — the description now states what the code does (envelope dropped, per-field writability carried as readOnly and enforced at write validation), the enum is NORTHBOUND/SOUTHBOUND per the ticket's own naming note, and a 400 response is documented for anything else. spec, ext bundle and generated client are consistent.

hivemq-edge/src/main/java/com/hivemq/protocols/tag/TagSchemaCreationOutputImpl.java
marregui: fixed in 3d4d6e20a: the panel compares the value sub-schemas — the envelope is fixed machinery, the value shape is what "the same schema" means to a reader — and the backend now guarantees NB.value renders identically to SB.value for plain tags (pinned in test_southboundValue_equalsNorthboundValue_forPlainValueTags).

on the nit: my claim was mechanically wrong. SchemaBuilder ignores annotations chained after .schema(prebuilt), so readable(false) never rendered anything — the value node's flags come from the adapter's own schema root, in both directions. the misleading calls are gone and the comment now states the real mechanism.

hivemq-edge/src/main/java/com/hivemq/protocols/tag/TagSchemaCreationOutputImpl.java
marregui: both taken in 3d4d6e20a: Direction is now the standalone TagSchemaDirection (the REST layer no longer imports the impl class to name a REST concept), and two pins — northboundSchema == SchemaJsonRepresentation.toCompositeSchema output, and getFuture() == getSchema(NORTHBOUND) — fail the day either copy drifts.

hivemq-edge/src/main/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImpl.java
marregui: as suggested, in 3d4d6e20a: strict valueOf + BadRequestError, the spec carries the 400, and lowercase/typos are rejected (tests cover omitted, SOUTHBOUND, typo, lowercase). getWritingSchema's 301 now carries ?direction=SOUTHBOUND — Location pinned in a test.

hivemq-edge-frontend/src/modules/Device/components/TagSchemaPanel.tsx
marregui: took the first way out: compare the value sub-schemas. the single-panel path is now reachable against the real backend (plain value tags), two labelled panels for independent write shapes — both asserted against realistically shaped mocks.

hivemq-edge-frontend/src/modules/Device/components/TagSchemaPanel.tsx
marregui: both fixed in 3d4d6e20a: the write query's error surfaces (device.errors.noWriteSchemaLoaded), and the two queries are independent — the read panel renders as soon as its own request settles, pinned by a deterministic slow-write test (8s delayed intercept).

hivemq-edge-frontend/src/modules/Device/components/TagSchemaPanel.spec.cy.tsx
marregui: rewritten with real document shapes: read = envelope doc, write = value-only, and a query-scoped direction=SOUTHBOUND intercept. the identical case now asserts the reachable state (matching value shapes), and a write-500 case asserts the error is visible.

hivemq-edge/src/test/java/com/hivemq/schema/TagSchemaCreationOutputImplSchemaBuilderTest.java
marregui: all four in 3d4d6e20a: context branch covered (NB-only), getFuture() pinned to the NORTHBOUND document, REST direction parsing covered incl. invalid and lowercase, and a class-level @Timeout(5) restores liveness. changed-files coverage is 100% per the bot.

hivemq-edge-frontend/src/components/rjsf/MqttTransformation/components/MappingEditor.tsx
marregui: deletion landed in 3d4d6e20a: the read-only card branch in MappingInstruction is gone (with its translation key), and MappingInstructionList hides read-only destinations while pruning stale instructions against the full list — EDG-59 closes with this PR.

two things on the record: (1) MappingContainer:88 persisted [...old, ...new] instructions — pre-existing, but it duplicated every kept instruction per edit and resurrected the pruned ones, defeating the pruning end-to-end; it now persists the replacement list. (2) the second consumer of the list (combiner destination editor) hides envelope-derived read-only fields from inferred schemas too — they were never mappable destinations; deliberate, commented at the filter.

marregui:
Took this over from Martin. All nine review threads are addressed in 3d4d6e20a — per-item closure is inline on each thread.

Three calls that went beyond the findings, on the record:
- **direction vocabulary is NORTHBOUND/SOUTHBOUND** — executing the design doc's own §1.6 naming note ("read/write holds only for variable-shaped tags") while the parameter is still unreleased and free to rename.
- **EDG-59 closes here** — read-only destinations are hidden rather than rendered as neutralised cards.
- **MappingContainer:88** appended the old instruction list to every emitted one (pre-existing on master): duplicated kept instructions per edit and resurrected pruned ones. Fixed here because it defeated the EDG-59 deliverable end-to-end.

QA'd outside-in on a full build against a real Prosys OPC UA server: the direction matrix (400s incl. the old READ/WRITE vocabulary), the 301 redirect carrying direction=SOUTHBOUND, NB.value == SB.value live on both read-only and writable nodes, and a southbound write confirmed at the device through the unchanged getFuture() DataHub path.

hivemq-edge-adapter-sdk#132 is merged; this PR compiles against it.

hivemq-edge/src/main/java/com/hivemq/protocols/tag/TagSchemaCreationOutputImpl.java
```diff
@@ -29,21 +33,114 @@ public class TagSchemaCreationOutputImpl implements TagSchemaCreationOutput {
...
+     * @deprecated retained for source compatibility with callers outside this repository (e.g. the commercial
+     *     bidirectional adapter and the integration tests in hivemq-edge-test). Use
+     *     {@link #getSchema(TagSchemaDirection)} instead, which makes the direction explicit. Note this returns
+     *     the NORTHBOUND schema, which is what this method always produced — the southbound DataHub resources in
+     *     {@code ProtocolAdapterWritingServiceImpl.createDataHubResources} rely on that.
+     */
+    @Deprecated(forRemoval = true)
     public @NotNull CompletableFuture<ObjectNode> getFuture() {
-        return future;
+        return future.thenApply(
```
caoccao, line 64: I think this is a bug because calling it multiple times might not be idempotent.

caoccao (changes requested):
I left one comment and https://linear.app/hivemq/document/sam-code-review-v01-91f16adfceaf

hivemq-edge/src/main/java/com/hivemq/protocols/tag/TagSchemaCreationOutputImpl.java
marregui: Good catch to check, but I don't think it's a bug — it's idempotent in value, just not in identity.

Composition is a pure function of the `DataPointSchema` the adapter finished with: `northboundSchema()` builds a fresh `SchemaBuilder` from immutable `Schema` records, `toJsonSchemaDocument` is stateless, and `finish()` completes the future once. So repeated calls always produce equal documents.

What does change is that each call returns a *separate* `ObjectNode`, and that's deliberate. `ObjectNode` is mutable — the previous `return future;` handed every caller the same instance, so one caller mutating it silently rewrote what every other caller already held. Per-call composition removes that.

Pinned in `test_repeatedCalls_produceEqualButIndependentDocuments`: equal documents, distinct instances, and a mutation to one is invisible to the others.

One real difference your instinct was pointing at: the returned stage is derived, so `cancel()`/`complete()` on it no longer propagates to the source future. No caller does either — the three `hivemq-edge-test` ITs each do a single `getFuture().get(...)`.

I can memoize and hand out `deepCopy()` instead if you'd prefer, but for a deprecated shim called once per writing context at startup that's more machinery for identical observable behaviour.

marregui:
Pushed `4e6965f` addressing all seven findings of [Sam's review](https://linear.app/hivemq/document/sam-code-review-v01-91f16adfceaf). Three companion PRs, because the change spans four repos:

- hivemq/hivemq-edge-adapter-sdk#138 — finding 7, `writeSchema` → `southboundSchema`
- hivemq/hivemq-edge-commercial-modules#355 — finding 2, **release blocker**
- hivemq/hivemq-edge-test#432 — the end-to-end merge gate

## In this commit

| Finding | Change |
|---|---|
| 1 | SDK Javadoc mandates `.writable()`; the A&C test shape now marks it, a new test asserts no `readOnly` on the command fields, another pins the unmarked-schema trap, and a third pins the **exact serialized document** so the Cypress fixture cannot drift from real backend output |
| 3 | OpenAPI path, `ext/` bundle and the TS client now say `readOnly` is descriptive metadata, **not** enforced |
| 4 | `MappingInstructionList` emits the sanitised instruction list as soon as the schema resolves, instead of waiting for the user to edit some other visible mapping |
| 5 | `getProperty` propagates effective `readOnly` to descendants, so a read-only container's children are no longer offered as write destinations |
| 6 | Schema comparison canonicalises object key order (array order preserved) |

## Findings 1 and 2, on the evidence

**Finding 2 is worse than described, and the worse half predates this PR.** `asText()` on a container node returns `""`; DataHub parses that to a `MissingNode` and builds a schema that accepts everything — `validate({"totally":"wrong"})` returns `[]`, verified by execution. It does not throw. Southbound DataHub validation has never validated anything. That may warrant its own ticket for changelog purposes.

**Finding 1 is a contract and test gap rather than a live break.** Every shipped adapter already marks writability, and the EDG-835 branch's `ConditionSchemas.writeSchema()` marks root and every field — it had already hit and solved the trap. What was missing was the SDK stating the requirement and a test pinning it, which is what this adds.

## Verification

Backend 163 tests / 0 failures · vitest 2139 passed · Cypress 131 across 24 affected specs, 0 failing · eslint back to the 81-warning baseline · Prettier clean.

Each behavioural fix was mutation-tested — reverted in isolation to confirm the new tests fail: finding 6 → 1 failing, finding 4 → 2, finding 5 → 1, finding 2 → 4 of 6 commercial tests plus the real-Edge IT.

## Sequencing note

Edge deliberately still calls the deprecated `writeSchema()`. CI resolves the adapter SDK by branch-name match (`git checkout ${{ github.head_ref }} || true`), so a PR whose branch has no SDK counterpart builds against SDK master. Switching to `southboundSchema()` before #138 lands would have broken every PR stacked on this branch — #1707 included. It flips over in a follow-up once #138 is merged.

## Still open

- Finding 4's *"also test submit behaviour end to end"* is not done. The chain works — `MqttTransformationField.handleChange` mutates the item in place and `handleSubmit` pushes it to the form — but nothing tests it.
- Finding 1's optional "safer SDK design" (a southbound builder defaulting to writable) is not done; the builder cannot distinguish an implicit `false` from an explicit `.writable(false)`, as the review itself notes.
- `pnpm build:tsc` is still red on 39 pre-existing errors outside this diff. Root cause found and it is not this PR: `tsconfig.json` is solution-style (`"files": []` + `references`), so the plain `tsc` in `pnpm build` compiles **zero files** and always exits 0 — the frontend has no working type check in CI. One of the 39 is a live bug (`LineChart` reads `serieColor`/`serieId`; nivo 0.99 renamed them to `seriesColor`/`seriesId`, so the metrics tooltip has no colour and no series name). Separate tickets, not this PR.

caoccao (approved):

marregui (approved):
Clearing my CHANGES_REQUESTED from 2026-08-05. Both blockers are addressed and every thread from that review is resolved:

- the OpenAPI `direction` description now states what the code does, and the enum is NORTHBOUND/SOUTHBOUND;
- `TagSchemaPanel` compares the value sub-schemas, so the single-panel path is reachable against a real backend — confirmed live against an OPC-UA adapter, not just against mocks.

Also fixed since: strict `direction` parsing with a 400, the write query's error surfacing independently, the read-only neutralising deletion in `MappingInstructionList`, and the four test gaps.

Reviewed at 4a08915. Checks green (22/22, Sonar gate passed); the composite status on this head is green from edge-composite #21.

Two follow-ups, neither blocking this PR:
- the UI copy `device.drawer.table.actions.schema` ("View the WRITE schema") and `device.drawer.schema.header` ("Manage the schema for the tag") now misdescribe the panel;
- the EDG-845 ticket text still promises `value` filtered to writable sub-fields, and still says the per-field neutralising logic should be deleted. The code deliberately does neither, and EDG-888's F8 shows deleting it would re-expose read-only fields as write destinations. Ticket text needs correcting, not the code.

PR #1706, opened 2026-07-30
EDG-849: Address OIDC QA findings

mschoenert-hivemq:
## Description (the why)

QA of the merged OIDC/SSO feature ([EDG-142](https://linear.app/hivemq/issue/EDG-142), PR #1634) surfaced nine findings ([EDG-849](https://linear.app/hivemq/issue/EDG-849), from Estefania Amundaray's test pass). This PR addresses them: five are fixed here, two are deferred to a follow-up that needs server-side state, and one is a documented won't-fix.

Each finding's rationale is recorded as a comment on EDG-849.

## What changed (per finding)

**Fixed here:**

- **#2 — default admin usable + password disclosed** (`9f0fbc4bf`). The built-in `admin`/`hivemq` account now survives only in the fully-implicit branch; enabling `<username-authentication>` explicitly with no user source is a startup error. XSD requires a present `<users>` to hold ≥1 `<user>`.
- **#5 — verbatim role mode grants admin by group name** (`10bd27f8e`). `<role-mappings>` is now required; verbatim mode removed. Only an explicitly mapped IdP role produces an Edge role.
- **#6 — refresh-token endpoint returns 401 to everyone** (`ccfa226c3`). `/api/v1/auth/refresh-token` was annotated `NO_AUTH_REQUIRED`, so no principal was installed for a handler that requires one. Now requires `admin/super/user`.
- **#7 — plain-http redirect-uri accepted** (`75050b22e`). The redirect-uri carries the auth code and Edge JWT back to the browser; it is now required to be https, with a literal loopback host as the only http exception.
- **#8 — DNS-based loopback check** (`75050b22e`). The loopback check resolved the host through DNS and trusted the result (so `127.0.0.1.nip.io` passed, and DNS rebinding was possible). Now matches loopback literally, removing DNS from the security decision.
- **#9 — four config traps** (`5ae827142`). `<extra-scopes>` is now a proper repeated list; `connection-timeout` renamed to `connection-timeout-millis` with a 100ms floor; a trailing-slash/case/whitespace issuer mismatch now produces a guiding message (exact match still authoritative); a wrong `role-claim-name` now emits a four-way operator diagnostic instead of an opaque `no-roles`.

**Deferred to [EDG-850](https://linear.app/hivemq/issue/EDG-850) (server-side login sessions):**

- **#1 — deprovisioned user keeps session** and **#3 — logout is cosmetic**. Both require token revocation, which a stateless bearer-JWT model cannot do; this is at parity with local username/password today. They are fixed together when a session store is introduced.

**Won't fix:**

- **#4 — login-state flood**. The exposure is out of scope (unauthenticated network reach to the admin port), availability-only, self-limiting, and the proposed fix would degrade a trivially-correct O(1) ring buffer into an O(n) scan that is slower precisely under attack.

## Acceptance Criteria

- [ ] With OIDC on and local auth explicitly enabled but no user source, startup fails (no default admin injected); `/frontend/configuration` discloses no default credentials.
- [ ] An IdP group literally named `admin`/`super`/`user` grants no Edge role without an explicit `<role-mapping>`.
- [ ] `POST /api/v1/auth/refresh-token` with a valid token returns 200 + a fresh token; without a token, 401.
- [ ] A non-loopback plain-http `redirect-uri` is rejected at startup; a literal-loopback http redirect is accepted.
- [ ] A public hostname resolving to loopback (e.g. `127.0.0.1.nip.io`) is rejected for an http issuer/endpoint.
- [ ] `<extra-scopes>` accepts multiple `<extra-scope>` elements; `connection-timeout-millis` below 100 is rejected; a trailing-slash issuer mismatch yields a guiding error; a wrong `role-claim-name` is diagnosable in the logs.

## Non-Goals

- Token revocation / server-side sessions (#1, #3) — tracked in EDG-850.
- Hardening the login-state ring buffer (#4) — declined; see the EDG-849 comment.
- Re-enabling the SPA header reissue / user-facing session renewal — belongs with EDG-850.
- Restructuring OPC-UA `retryIntervalMs` (the other list-as-string field) — out of scope; retry handling is being reworked separately.


hivemq-edge/src/main/resources/config.xsd
```diff
@@ -1038,7 +1038,7 @@
             <xs:element name="users" minOccurs="0">
                 <xs:complexType>
                     <xs:sequence>
-                        <xs:choice minOccurs="0" maxOccurs="unbounded">
+                        <xs:choice minOccurs="1" maxOccurs="unbounded">
```
marregui: This does not do what it says. The choice has a single branch, `<xs:element name="user" minOccurs="0">`, which matches the empty sequence — so a choice with `minOccurs="1"` is still satisfied by zero elements. The content model is unchanged from master.

Verified against this branch's `config.xsd` with JAXP/Xerces (Edge's own validator) and independently with libxml2 — both accept:

```xml
<hivemq><admin-api><users></users></admin-api></hivemq>
```

Consequence beyond the dead guard: `<users></users>` + no `<username-authentication>` + no `<ldap>` + no `<oidc-authentication>` reaches `setUserList(DEFAULT_USERS)`, and `FrontendResourceImpl.getFirstUse()` then serves `admin`/`hivemq` to an unauthenticated `GET /api/v1/frontend/configuration`. An operator who writes an empty `<users/>` to mean "no local accounts" gets the built-in admin instead.

Fix — tested, rejects `<users/>` and accepts one user:

```xml
<xs:choice minOccurs="1" maxOccurs="unbounded">
    <xs:element name="user" minOccurs="1">
```

hivemq-edge/src/main/java/com/hivemq/configuration/reader/ApiConfigurator.java
```diff
@@ -136,15 +136,29 @@ public boolean needsRestartWithConfig(final @NotNull HiveMQConfigEntity config)
             if (entity.getLdap() != null) {
                 apiCfgService.setLdapConnectionProperties(LdapConnectionProperties.fromEntity(entity.getLdap()));
             } else {
+                // The schema requires a present <users> element to contain at least one <user>, so an empty
```
marregui: This premise is false — see `config.xsd:1041`. A present `<users>` may still be empty, so `users.isEmpty()` does not mean "`<users>` is absent", which is exactly what the branch below relies on to decide between injecting the built-in admin and rejecting the config.

Either fix the schema and keep the comment, or distinguish absent from empty here.

hivemq-edge/src/test/java/com/hivemq/configuration/reader/ApiConfiguratorTest.java
```diff
@@ -65,12 +75,43 @@ public void noAuthConfig_defaultsToUsernamePasswordEnabled() throws Exception {
...
+    public void usernameAuthEnabledExplicitlyWithNoSource_isRejected() throws Exception {
+        // Local login turned on explicitly but no <users> and no <ldap>: rather than silently injecting the
+        // built-in admin account (which would expose default credentials, EDG-849), reject the config.
+        writeConfig(usernameAuth(true));
+
+        assertThrows(UnrecoverableException.class, () -> reader.applyConfig());
+    }
+
+    @Test
+    public void emptyUsersElement_isRejectedByTheSchema() throws Exception {
```
marregui: Green for the wrong reason. This config passes schema validation (verified), so `assertThrows(Exception.class)` is catching the `UnrecoverableException` from the new branch in `ApiConfigurator`, not a schema rejection.

The test therefore cannot fail if the schema regresses — and the schema is already regressed. Once `config.xsd:1041` is fixed, worth asserting on the schema-validation exception specifically so the name and the assertion agree.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
```diff
@@ -344,32 +345,88 @@ public OidcServiceImpl(
...
+                log.warn(
+                        "OIDC login denied: none of the {} IdP role(s) under claim '{}' matched a "
+                                + "<role-mapping>: {}",
+                        idpRoles.size(),
+                        claimName,
+                        idpRoles);
             }
+        } else if (claims.toJSONObject().containsKey(claimName)) {
+            log.warn(
+                    "OIDC login denied: the role claim '{}' is present in the ID token but empty; the user has "
```
marregui: A claim that is present but is neither a string nor a string array lands here and is reported as "the user has no roles" — which is the diagnosis, not the problem.

Verified against Nimbus 11.38.2 with `realm_access` in Keycloak's default nested shape:

```
claim=realm_access   getStringListClaim=null  getStringClaim=null  containsKey=true
claim=emptyArr       getStringListClaim=[]    getStringClaim=null  containsKey=true
```

Both collapse to this message, but they are opposite diagnoses: one is a user with genuinely no roles, the other is `role-claim-name` pointed at a claim Edge cannot read. `realm_access` is a plausible operator guess on Keycloak — the IdP in the QA fixture — so this sits squarely inside #9's remit. Suggest a fourth case that names the shape actually found.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
```diff
@@ -344,32 +345,88 @@ public OidcServiceImpl(
      */
     private static @NotNull Set<String> mapRoles(
```
marregui, line 343: The javadoc above (lines 334-343) still documents verbatim mode as the design:

> Two modes, selected by whether `<role-mappings>` is configured … **Verbatim** (no mappings): each claim value that is an Edge role … is used directly … The operator's omission of a mapping is an explicit statement that the IdP already uses Edge role names.

That is finding #5 described as intended behaviour, on the method that implements its removal. Everything else in the diff was updated — the `OidcConfiguration` field, its class javadoc, `resolveEdgeRoles`, the test class — this one was missed.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
```diff
@@ -344,32 +345,88 @@ public OidcServiceImpl(
...
+        if (!idpRoles.isEmpty()) {
+            // An IdP role is "unmatched" when it is not a mapping key (so it produced no Edge role).
+            final List<String> unmatched = idpRoles.stream()
+                    .filter(role -> !mappings.containsKey(role))
+                    .toList();
+            if (!edgeRoles.isEmpty()) {
+                // Login succeeds. Warn only when some roles were dropped -- a possible missing <role-mapping>.
+                if (!unmatched.isEmpty()) {
+                    log.warn(
+                            "OIDC login authenticated with {} Edge role(s); {} IdP role(s) under claim '{}' did "
```
marregui, line 389: This warns on the success path. Keycloak's standard token carries `offline_access`, `uma_authorization` and `default-roles-<realm>` alongside the mapped group, so **every** successful login warns — in the QA fixture and in the field.

That is the noise that desensitises operators to the denial warnings below, which are the ones that matter. Suggest INFO (or DEBUG) here and keeping WARN for the three denial branches.

hivemq-edge/src/main/resources/config.xsd
```diff
@@ -1222,8 +1222,17 @@
...
+                            <xs:complexType>
+                                <xs:sequence>
+                                    <xs:element name="extra-scope"
+                                                type="nonEmptyString"
+                                                minOccurs="1"
+                                                maxOccurs="unbounded"/>
+                                </xs:sequence>
+                            </xs:complexType>
+                        </xs:element>
+                        <xs:element name="role-mappings" minOccurs="1">
```
marregui: `minOccurs="1"` applies regardless of `<enabled>`, so a pre-staged disabled stanza now fails to validate — verified, master accepts it and this branch rejects it.

Every other required OIDC field (`issuer-uri`, `client-id`, `redirect-uri`) is `minOccurs="0"` here and enforced in Java only under `if (oidcEnabled)`. This is the one exception, and it is the one that breaks the case `oidcStanzaDisabled_leavesOidcInactive` documents.

The `Preconditions.checkArgument` you added in `fromEntity` already delivers the security requirement for the enabled case. Suggest `minOccurs="0"` here and keeping the Java check as the single source of the rule.

hivemq-edge/src/main/java/com/hivemq/configuration/entity/api/oidc/OidcAuthenticationEntity.java
```diff
@@ -47,7 +47,10 @@
  *     <client-secret>secret</client-secret>
  *     <redirect-uri>https://edge.example.com/api/v1/auth/oidc/callback</redirect-uri>
  *     <role-claim-name>roles</role-claim-name>
- *     <extra-scopes>email profile</extra-scopes>
+ *     <extra-scopes>
+ *         <extra-scope>email</extra-scope>
+ *         <extra-scope>profile</extra-scope>
+ *     </extra-scopes>
  *     <role-mappings>
  *         <mapping>
```
marregui: `<mapping>` should be `<role-mapping>` — both the XSD and `@XmlElement(name = "role-mapping")` require it (same on line 59).

Now that `<role-mappings>` is mandatory, this is the example operators will copy, and copying it fails schema validation at startup.

marregui (changes requested):
Reviewed against EDG-849's stated requirements and the AC in the description — diff plus the seams it leans on (`ApiConfigurator`, `ApiAuthenticationFeature`/`AbstractApi`, `FrontendResourceImpl`, `OidcServiceImpl.resolveMetadata`, `config.xsd`).

Five fixes, right five. #5, #6, #7 and #8 are correct. The literal-loopback rewrite and the issuer exact-match-with-guiding-message are the right pattern in particular — the security decision is never relaxed, only the message.

One blocker, then small stuff.

## Blocker: the `<users>` schema tightening is a no-op, so #2's fix rests on a guard that isn't there

Detail inline on `config.xsd:1041`. Short version: I validated `<users></users>` against this branch's own `config.xsd` with the JAXP/Xerces stack Edge uses, and it still passes.

```
empty <users/>, no username-authentication      master=ACCEPT   PR=ACCEPT
empty <users/> + username-auth enabled          master=ACCEPT   PR=ACCEPT
```

To be clear about what that does and does not mean: **#2 as scoped is fixed.** With an `<oidc-authentication>` stanza present the presence rule forces `<username-authentication>`, so the configuration QA exploited is closed. What is missing is the backstop the code claims — `users.isEmpty()` still cannot tell "absent" from "present and empty", the comment at `ApiConfigurator.java:139` asserting otherwise is false, and `emptyUsersElement_isRejectedByTheSchema` passes on the Java-side throw rather than on the schema, so it cannot fail if the schema regresses. It already has.

## Not a code change, but it decides whether the ticket can leave QA

`validateDiscoveryMetadata` is reached only from `resolveMetadata`, which is lazy and TTL-cached — it runs on first login, never at startup. R.12 asserts *"login works (normalise) **or** startup rejects it"*; R.13 asserts the failure code differs from a genuinely role-less user. This PR changes neither outcome, only the log line, so both known-defect tests stay red after merge. That may be the right trade, but it is an AC change rather than a restatement. Raised on EDG-849 so Estefania is in the decision.

## Verified clean — worth not regressing

- The refresh-token wiring is real end-to-end, not just a spec edit: `x-roles-allowed` → `apiInterface.mustache` → `@RolesAllowed` on the generated interface → `ApiAuthenticationFeature` registers `AuthenticationFilter`. Both OpenAPI files moved together.
- `isLiteralLoopbackHost` handles the `[::1]` bracket form from `URI.getHost()`, covers all of 127/8, rejects `127.0.0.256` and `127.0.0.1.nip.io`, and takes DNS out of the decision entirely so there is nothing left to rebind.
- `checkIssuerMatch` never returns on the lenient path — the lenient compare selects the message and nothing else.
- The three renames break nobody: `git tag --contains 2245e67f5` is empty and the latest tag is 2026.11, so there is no released OIDC config in the field. Every existing QA config does need updating though — all three old forms are now hard schema rejections.

## One for the ticket, not the code

#6 is fixed only when `enforce-api-auth` is true. `AuthenticationFilter.filter` short-circuits on `!isEnforceApiAuth()` without installing a principal, and `AbstractApi.getAuthenticatedPrincipalFromContext` then throws 401 — the original symptom. Both new tests mock it to `true`. Pre-existing shape, not introduced here.

hivemq-edge/src/main/resources/config.xsd
mschoenert-hivemq: Good catch — confirmed and fixed in 79afc551d. You're right: a single-branch xs:choice minOccurs="1" around <user minOccurs="0"> is satisfied by zero users, so <users/> validated and my "requires ≥1 <user>" claim was wrong. The empty-<users/> → setUserList(DEFAULT_USERS) → unauthenticated admin/hivemq disclosure was genuinely still reachable, so finding #2 was only half-closed — the code rejection worked, the schema guard did not. Fix is exactly yours: inner <user> is now minOccurs="1". Kept the redundant single-branch xs:choice rather than simplify it, to keep the change minimal.

hivemq-edge/src/main/java/com/hivemq/configuration/reader/ApiConfigurator.java
mschoenert-hivemq: You're right the premise was false. It's true now: with the schema fixed (79afc551d), an empty <users> can no longer reach this code, so users.isEmpty() does mean absent. The entity models <users> as an @XmlElementWrapper over a @NotNull list initialised empty, so absent and present-but-empty are indistinguishable here — which is exactly why the rule has to live in the schema rather than this branch. Left the comment and branch as-is; both are correct now that the schema holds.

hivemq-edge/src/test/java/com/hivemq/configuration/reader/ApiConfiguratorTest.java
mschoenert-hivemq: Agreed — it was green for the wrong reason. Rewrote it (79afc551d) to disable local auth, so no code path can reject the config and the schema is the only possible source of failure — which makes it a genuine guard against exactly this regression. Added a positive test that a non-empty <users> still validates and is read as the user source.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
mschoenert-hivemq: Fixed in f4cc782b6 — your "fourth case." A claim that's present but neither a string nor a string-array (a nested object like Keycloak's realm_access) is now detected and reported distinctly: the message names the value's type and points the operator at the role list inside the container, instead of folding it into "the user has no roles." So the wrong-shape case (an unambiguous role-claim-name misconfiguration — a role-less user never produces that shape) is now separated from the genuinely-role-less case. isSupportedRoleClaimShape (string / array-of-strings supported; nested object, mixed array, number, null not) is unit-tested. Thanks for catching it.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
mschoenert-hivemq: Good catch — fixed. You're right that this javadoc was the one spot #5 missed: it still described the removed verbatim mode as current design, on the method that implements its removal. Rewritten to state the actual behaviour — mapping is always strict, <role-mappings> is required, and only an explicitly configured mapping key produces an Edge role (no verbatim mode). Thanks.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
mschoenert-hivemq: Good catch — fixed. You're right that unmatched roles are the norm on the success path, not a problem: a real Keycloak token always carries offline_access/uma_authorization/default-roles-<realm> next to the mapped group, so this warned on every successful login and drowned the denial warnings that actually matter. Dropped it to DEBUG (available when troubleshooting, silent otherwise) and kept WARN for the three denial branches. Thanks.

hivemq-edge/src/main/resources/config.xsd
mschoenert-hivemq: Fixed in ca42b2455, exactly as you suggested. Outer <role-mappings> is now minOccurs="0" so a disabled (pre-staged) stanza validates, and the fromEntity checkArgument is the single source of the enabled-requires-mappings rule. Kept the inner <role-mapping> at minOccurs="1", so a present <role-mappings> must still hold at least one mapping. Covered by tests in ApiConfiguratorTest (disabled-without-mappings validates; enabled-without rejected; present-but-empty rejected by the schema).

hivemq-edge/src/main/java/com/hivemq/configuration/entity/api/oidc/OidcAuthenticationEntity.java
mschoenert-hivemq: Fixed — both <mapping> pairs are now <role-mapping>, matching the XSD and @XmlElement(name = "role-mapping"). You're right this matters more now that <role-mappings> is required: it's the example operators copy, and the old form failed schema validation at startup. Thanks.

hivemq-edge/src/main/resources/config.xsd
mschoenert-hivemq: Correction / follow-up: I've reverted 79afc551d (in e32c9328e) — so my "fixed in the schema" reply above no longer holds. An empty <users> is legitimate (LDAP is the source, OIDC-only, local login disabled, or the first-boot default), so the non-empty requirement is contextual and can't live in a context-free schema; it stays in code, where ApiConfigurator already rejects the one dangerous combination (9f0fbc4bf) — so finding #2 stays closed. The minOccurs="1" had also broken ~289 ITs (the config round-trip emits an empty <users/>). Full reasoning is on the ticket; build green (#9).

hivemq-edge/src/main/java/com/hivemq/configuration/entity/api/oidc/OidcAuthenticationEntity.java
```diff
@@ -84,8 +87,9 @@ public class OidcAuthenticationEntity {
     @XmlElement(name = "role-claim-name", defaultValue = "roles")
     private @NotNull String roleClaimName = "roles";
 
-    @XmlElement(name = "extra-scopes")
-    private @Nullable String extraScopes = null;
+    @XmlElementWrapper(name = "extra-scopes", required = false)
+    @XmlElement(name = "extra-scope")
+    private @Nullable List<String> extraScopes = null;
```
marregui, line 91: **Must fix.** This line is right; the one four lines below it is not, and it breaks config sync.

`roleMappings` (`:96`) is initialised to `new ArrayList<>()`, while both its siblings here — `extraScopes` (this line) and `idTokenSigningAlgorithms` (`:100`) — are `null`. That difference is load-bearing: `@XmlElementWrapper` emits the wrapper for any *non-null* collection, so an absent `<role-mappings>` leaves an empty list that marshals to `<role-mappings/>`, and `config.xsd:1238` requires >=1 `<role-mapping>`. `ConfigFileReaderWriter.createMarshaller()` attaches the schema to the **marshaller**, so the write is validated.

Verified by execution — full unmarshal to marshal cycle against this branch's `config.xsd`, same annotations and initialiser, same schema attachment on both sides:

```
config.xml on disk (pre-staged, DISABLED oidc, no <role-mappings>):
  <hivemq><admin-api>
     <username-authentication><enabled>true</enabled></username-authentication>
     <oidc-authentication><enabled>false</enabled></oidc-authentication>
  </admin-api></hivemq>

1. unmarshal (startup config read)   -> OK
   roleMappings after read = []      (null? false)
2. marshal   (writeConfigWithSync)   -> FAILS
   cvc-complex-type.2.4.b: The content of element 'role-mappings' is not complete.
   One of '{role-mapping}' is expected.
```

This is the same empty-wrapper-vs-schema mechanism you diagnosed for `<users>` in `e32c9328e` — the round-trip emitting an element the schema then refuses.

**Reachability is exactly one configuration**, and it is the one `ca42b2455` added support for:

- OIDC stanza absent -> nothing marshalled -> fine.
- OIDC **enabled** without mappings -> already rejected at startup by `fromEntity` -> unreachable.
- OIDC **disabled** without mappings -> startup accepted (verified: the XSD accepts it) -> **config sync broken from then on.**

**Impact.** `writeConfigWithSync()` is called by `ProtocolAdapterExtractor`, `BridgeExtractor`, `DataCombiningExtractor`, `UnsExtractor`, `AssetMappingExtractor` and `PulseExtractor`, and wraps the write in `catch (final Exception e) { log.error("Configuration file sync failed: ", e); }`. So the API call succeeds, the change applies in memory, and the file is never updated — the operator adds an adapter, gets a 200, and it is gone after restart. `writeConfigToXML(Writer)` rethrows as `UnrecoverableException`, so config download/export fails outright.

**Fix** (verified by execution — marshal then succeeds), which also makes the field consistent with its two siblings:

```java
private @Nullable List<OidcRoleMappingEntity> roleMappings = null;
```

`fromEntity` already handles null (`checkArgument(mappingEntities != null && !mappingEntities.isEmpty(), …)`). Worth extending `disabledOidcStanzaWithoutRoleMappings_isAccepted` to write the config back — it calls `applyConfig()` only, so it certifies the half of the round trip that works.

Attribution, to be fair: `master` carries both the initialiser and the inner `minOccurs="1"`, so this PR did not introduce it. It is in scope because `ca42b2455` promotes the triggering configuration to a supported one.

One caveat worth stating: my harness replicates this entity's annotations rather than loading the class itself, so please reproduce against the real entity before signing this off — the one-line change makes that cheap.

hivemq-edge/src/test/java/com/hivemq/configuration/reader/ApiConfiguratorTest.java
```diff
@@ -106,13 +162,56 @@ public void oidcEnabledWithLocalDisabled_activatesOnlyOidc() throws Exception {
...
+                + "<issuer-uri>https://idp.example.com</issuer-uri>"
+                + "<client-id>edge</client-id>"
+                + "<redirect-uri>https://edge.example.com/api/v1/auth/oidc/callback</redirect-uri>"
+                + "</oidc-authentication>");
+
+        assertThrows(Exception.class, () -> reader.applyConfig());
+    }
+
+    @Test
+    public void oidcWithEmptyRoleMappingsElement_isRejectedByTheSchema() throws Exception {
```
marregui: Green for the wrong reason — the same shape as the `<users>` test in round 1, one test down.

The name asserts the **schema** rejects an empty `<role-mappings>`, but `assertThrows(Exception.class, …)` also passes on the code path: the empty element unmarshals to an empty list, `OidcConfiguration.fromEntity` hits `checkArgument(mappingEntities != null && !mappingEntities.isEmpty(), …)`, and `ApiConfigurator` wraps the `IllegalArgumentException` into `UnrecoverableException`.

So it passes whether or not the schema rejects anything, and **cannot fail if `config.xsd:1238`'s `minOccurs="1"` regresses** — the only thing it names.

The schema rule is real (verified by execution: `ENABLED oidc, empty <role-mappings></role-mappings>` -> `REJECT cvc-complex-type.2.4.b`). It just is not what this test proves. Suggest asserting on the schema-validation exception or its message, as you did for `emptyUsersElement_isAccepted`.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
```diff
@@ -331,46 +332,133 @@ public OidcServiceImpl(
...
+                                    + "not match any <role-mapping> and were ignored: {}",
+                            edgeRoles.size(),
+                            unmatched.size(),
+                            claimName,
+                            unmatched);
+                }
+            } else {
+                log.warn(
+                        "OIDC login denied: none of the {} IdP role(s) under claim '{}' matched a "
```
marregui: This branch logs the raw claim **values** at WARN, and the javadoc justifies it with *"IdP role values are logged (they are the names the operator maps, not personal data)."*

That premise holds only when `role-claim-name` is correct — which is exactly what R.13 exists because you cannot assume. Point it at `email`, `name` or `preferred_username` (all plausible operator guesses, and the class of mistake this diagnostic is meant to catch) and every denied login writes the user's identifier into the log at WARN.

It is also internally inconsistent: the wrong-shape branch you added in `f4cc782b6`, twelve lines down, deliberately does *not* log the value — *"The value's type is named; its contents are not logged."* Same method, opposite posture, and the branch that does log is the one reachable with a misconfigured claim name.

Cheapest fix: counts and claim name at WARN, values at DEBUG.

marregui (changes requested):
Round 2, against the current head `eb74ac445` (12 commits). Same yardstick — EDG-849's requirements and the AC in the description — with scope extended to the config **write** path (`ConfigFileReaderWriter`, JAXB marshalling), which round 1 did not cover and which is where the one new must-fix lives.

Round 1 is well handled: six fixed, `<users>` reverted with a rationale I accept, R.12/R.13 converted into a test amendment now sitting with Estefania. `isSupportedRoleClaimShape`, the rewritten verbatim javadoc, the DEBUG demotion and the `<role-mapping>` example all land cleanly, and all four #9 rules do what they claim — verified by execution against this head's `config.xsd`:

| Config | Result |
|---|---|
| old `<connection-timeout>5000</>` | REJECT |
| `<connection-timeout-millis>1</>` | REJECT (`cvc-minInclusive-valid`) |
| `<connection-timeout-millis>100</>` (boundary) | ACCEPT |
| old `<extra-scopes>email profile</>` | REJECT |
| new `<extra-scopes><extra-scope>…` | ACCEPT |
| DISABLED oidc, no `<role-mappings>` | ACCEPT |
| ENABLED oidc, empty `<role-mappings>` | REJECT |

## One new must-fix

Inline on `OidcAuthenticationEntity`: a **disabled, pre-staged `<oidc-authentication>` stanza can be read but not written back.** It is the same empty-wrapper-vs-schema mechanism you diagnosed for `<users>` in `e32c9328e`, one field over — and it lands on exactly the configuration `ca42b2455` added support for. Config sync fails on marshal, so every subsequent adapter/bridge/combiner change made through the UI applies in memory, returns success, and is lost on restart. Verified by execution; one-word fix.

Then two hygiene items (a dead schema line the revert left behind; a test whose name and assertion disagree in the way round 1 flagged) and two considerations. Nothing else is new exposure.

## One for the ticket, not the code

`ApiAuthenticationFeature.java:179` — the `AuthenticationFilter` body is wrapped in `catch (final Exception e) { log.error(…) }` with no `abortWith`, so if the authentication chain throws, the filter returns and the request proceeds with no security context installed. For `/refresh-token` that is fail-*closed* (the handler reads the principal → 401), so #6's fix is not weakened; for an endpoint that does not read the principal it would be fail-open. **I did not establish that any handler actually throws**, so this is a latent shape rather than a demonstrated bug — raising it only because #6 moves an endpoint onto this filter. Recorded on EDG-849.

## Verified clean — three hypotheses I refuted before reporting

- **Role-name case is a non-issue.** `x-roles-allowed` is lowercase, `ApiRoles` constants are uppercase, and `<edge-role>` is stored verbatim — but `ApiPrincipal.hasRole` and `isValidEdgeRole` both compare `equalsIgnoreCase`, so a mapping written `<edge-role>Admin</edge-role>` is harmless.
- **"The built-in admin survives ONLY in the fully-implicit branch" holds.** The LDAP branch never calls `setUserList`, but `ApiConfigurationServiceImpl` initialises `userList` to an empty `CopyOnWriteArrayList`, so nothing stale is left for `getFirstUse()` to disclose.
- **The no-source error message's advice actually works.** It suggests `<enabled>false</enabled>`, and the branch keys on `getUsernameAuthentication() == null` — but the whole block sits under `if (usernameAuthEnabled)`, so it is reached only when the stanza is present *and* enabled.

Round 1's verified-clean list still holds on this head: `isLiteralLoopbackHost`, `checkIssuerMatch`'s never-return-on-lenient shape, and the refresh-token `x-roles-allowed` → `@RolesAllowed` wiring.

hivemq-edge/src/main/resources/config.xsd
marregui: Agreed on the revert, and the reasoning is right — the non-empty rule is contextual (`<enabled>` and `<ldap>`), so it cannot live in a context-free schema. Thanks for the 289-IT detail; that is the same mechanism as the new must-fix on `OidcAuthenticationEntity`.

One leftover though. The revert took out the inner `<xs:element name="user" minOccurs="1">` — the half that did the work — but this line still carries `minOccurs="0"` -> `minOccurs="1"` from `9f0fbc4bf`. That is the original no-op: a single-branch choice around an element that matches the empty sequence.

Verified by execution against this head's `config.xsd`:

```
empty <users></users>, local auth disabled      ACCEPT
empty <users></users>, NOTHING else stated      ACCEPT
empty <users/> self-closing, nothing else       ACCEPT
one <user>                                      ACCEPT
```

Given the decision is that the rule lives in code, the honest diff here is `minOccurs="0"` — i.e. no change to this line at all. Left as is, the next reader sees a tightened occurrence constraint sitting next to your (correct) comment at `ApiConfigurator.java:139-143` saying the rule "is not enforced away in the schema", and repeats this whole analysis to discover the line does nothing.

hivemq-edge/src/test/java/com/hivemq/configuration/reader/ApiConfiguratorTest.java
marregui: The rewrite is better, but it does not reach the path it is named for. `emptyUsersElement_isAccepted` uses `usernameAuth(false)`, so it lands in the `else` at `ApiConfigurator.java:167` (`setUserList(List.of())`) and never reads `<users>` at all — the assertion holds for a reason unrelated to `<users>` being empty.

The path actually under discussion is `<users></users>` with **no** `<username-authentication>`: `ApiConfigurator.java:151` -> `setUserList(DEFAULT_USERS)` -> `FrontendResourceImpl.java:200` serves `admin`/`hivemq` to an unauthenticated `GET /api/v1/frontend/configuration`.

I am not reopening the scope call — you ruled it is the first-boot branch and that separating absent from present-empty is a JAXB entity change, and that is defensible. But since it is now a deliberate accepted behaviour, a test pinning it makes it visible rather than something rediscovered as finding #2 next round. Related note on the ticket about the disclosure primitive itself.

hivemq-edge/src/main/java/com/hivemq/configuration/reader/ApiConfigurator.java
marregui: Confirmed — the comment at `:139-143` now describes what the code actually does (JAXB collapsing absent and present-empty, and the rule living here rather than in the schema). Accurate post-revert. Resolvable.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
marregui: Confirmed fixed in `f4cc782b6`. `isSupportedRoleClaimShape` gets the subtle case right too: an empty list returns `true` (`allMatch` over an empty stream), so a genuinely role-less user is not misreported as a wrong shape — which is the distinction the fix exists to make. Seven unit tests including the nested-object, mixed-array and number cases. Resolvable.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
marregui: Confirmed fixed — the javadoc now describes strict-only mapping with no verbatim mode, on the method that implements its removal. Resolvable.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
marregui: Confirmed fixed — DEBUG on the success path, WARN kept for the three denial branches. One related point on a sibling branch (claim *values* logged at WARN) raised as a separate comment. Resolvable.

hivemq-edge/src/main/resources/config.xsd
marregui: Confirmed fixed in `ca42b2455` — verified by execution: a disabled OIDC stanza without `<role-mappings>` ACCEPTs, and a present-but-empty `<role-mappings>` REJECTs by the schema. Both as intended.

Worth noting this is what surfaced the new must-fix on `OidcAuthenticationEntity`: the configuration this now admits at read time cannot be written back, because `roleMappings` is initialised to an empty list and marshals to `<role-mappings/>`. Resolvable once that is addressed.

hivemq-edge/src/main/java/com/hivemq/configuration/entity/api/oidc/OidcAuthenticationEntity.java
marregui: Confirmed fixed — both pairs are `<role-mapping>` now, matching the XSD and the `@XmlElement` name. Resolvable.

hivemq-edge/src/main/java/com/hivemq/configuration/entity/api/oidc/OidcAuthenticationEntity.java
mschoenert-hivemq: Fixed in a9fa3bd3d. Confirmed by a round-trip test first — reading a disabled <oidc-authentication> (no <role-mappings>) then re-marshalling failed with cvc-complex-type.2.4.b: element 'role-mappings' is not complete, exactly as you traced: the roleMappings field defaulted to an empty ArrayList, so JAXB emitted an empty <role-mappings></role-mappings> that the schema then rejected on write. Same empty-wrapper mechanism as <users>, one field over.

Rather than the one-word field fix alone, I moved the cardinality rule out of the schema entirely (it's the cleaner resolution for both this and the <users> class):

roleMappings now defaults to null (matching its siblings extraScopes/idTokenSigningAlgorithms), so an empty/absent list is omitted on marshal instead of emitting the empty wrapper.
The XSD inner <role-mapping> is relaxed to minOccurs="0", so a present-but-empty <role-mappings> no longer fails validation.
"Enabled OIDC requires ≥1 mapping" (finding #5) now lives solely in OidcConfiguration.fromEntity, gated on <enabled> — verified an enabled stanza with empty mappings is still rejected, by code.
The round-trip test is committed (disabled stanza reads and writes back with <role-mappings> omitted). Thanks — this was a real silent-data-loss path, and the reproduction made that concrete.

hivemq-edge/src/main/resources/config.xsd
mschoenert-hivemq: Done in 1332f84dc — and you were more right than the minimal fix. You're correct that the revert left 9f0fbc4bf's outer minOccurs="1" in place (master has minOccurs="0"), so the line was a meaningless diff reading as a live constraint next to the "rule lives in code" comment.

Rather than just restore the master value, I removed the redundant single-branch <xs:choice> entirely — <user> is now a plain minOccurs="0" maxOccurs="unbounded" element. A single-branch choice is semantically identical to its lone element, and AdminApiEntity binds <users> via @XmlElementWrapper + @XmlElementRef (it sees the <user> children, not the choice), so marshalling is unaffected.

Given how the <role-mappings> write-back bit us, I didn't want to assert equivalence — I proved it: a full read + write-back round trip through the real ConfigFileReaderWriter (both sides validate against config.xsd) for every <users> shape — empty, one-user-with-roles, multiple — all green. No behaviour change; empty <users> still accepted.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
mschoenert-hivemq: Fixed in 6b62c0aca — and you're right on both the harm and the inconsistency. With role-claim-name pointed at email/preferred_username (the exact class of mistake the diagnostic exists to catch), this branch was reached with the claim present and non-empty, so every denied login wrote the user's identifier to the log at WARN. And it contradicted the wrong-shape branch twelve lines down, which already logs only the value's type.

The fix went a bit wider than this one line: we swept every log call in the OIDC path and applied one rule — a WARN diagnostic may carry counts (size()), config-supplied strings (the claim name, a <role-mappings> key), and IdP claim names (schema, not personal data), but never token values. Three lines were logging token values:

this branch — dropped the value list from WARN (count + claim name remain); values now at DEBUG;
the no-roles denial — dropped the subject (sub is a user identifier); the reason is already logged above;
the lenient near-miss warning — now names only the config mapping key that was nearly matched, not the token-supplied IdP role.
The sibling OIDC files (OidcStateStore, OidcCallbackPage) were checked too — they log no token content. Thanks; this was the right call, and it turned into a clean invariant for the whole path.

hivemq-edge/src/test/java/com/hivemq/configuration/reader/ApiConfiguratorTest.java
mschoenert-hivemq: Addressed in a9fa3bd3d, though in the opposite direction from the suggestion — because that commit removed the schema rule this test was named for. When I moved the "enabled OIDC needs ≥1 mapping" rule out of the schema and into fromEntity (option 2 for the write-back fix), the inner <role-mapping> went to minOccurs="0" — so there's no longer a schema rejection to assert on; asserting on it would guard a rule we deliberately deleted.

Instead the test is now enabledOidcWithEmptyRoleMappings_isRejectedByTheCode, with a comment stating the rule lives in fromEntity, not the schema. So the name and the assertion agree — the rejection is the code path, and the test says so. Your "name and assertion should agree" point stands; it's just resolved by renaming to match where the rule actually is now, rather than by asserting on a schema constraint that no longer exists.

hivemq-edge/src/test/java/com/hivemq/configuration/reader/ApiConfiguratorTest.java
mschoenert-hivemq: Done in 2e0e05bd3 — both halves. You're right that emptyUsersElement_isAccepted used usernameAuth(false), so it landed in the local-auth-off branch and never read <users> — passing for a reason unrelated to the element being empty. Renamed it to emptyUsersElement_withLocalAuthOff_isAcceptedAndUnused and reworded the comment to say exactly that.

And added the pin you asked for: emptyUsersElement_withNothingElse_yieldsTheBuiltInAdminAccount — <users></users> with no <username-authentication>, <ldap>, or <oidc> reaches the fully-implicit branch and installs the built-in admin (same as an absent <users>; JAXB collapses absent and present-empty to an empty list). So the first-boot default is now documented as a deliberate decision rather than something to be rediscovered as finding #2 next round. Not reopening the scope call — this just makes the accepted behaviour visible, as you suggested.

hivemq-edge/src/main/java/com/hivemq/api/config/OidcConfiguration.java
```diff
@@ -161,36 +164,36 @@ public OidcConfiguration(
...
+        final List<String> scopes = configuredScopes == null ? List.of() : List.copyOf(configuredScopes);
 
-        // Role mappings define two distinct modes. An absent <role-mappings> element is verbatim mode:
-        // the claim values are used as Edge roles directly. A present element (the XSD guarantees it is
-        // non-empty) is strict mode: only mapped IdP roles produce an Edge role. Keys are stored and
-        // matched literally — no trimming or case-folding — so a stray space or wrong case is an honest
-        // mismatch the operator can see in the logs, not a silent, surprising match.
+        // Role mappings are required: only an explicitly mapped IdP role produces an Edge role, and an
+        // unmapped value grants nothing. The <role-mappings> element must be present with at least one
+        // <role-mapping> (enforced by the XSD; re-checked here so a programmatic entity cannot bypass it).
```
marregui: `a9fa3bd3d` deleted this XSD rule — this check is now the only enforcement of #5, not a re-check. Worth saying so before someone reads it as removable.

```suggestion
        // Role mappings are required: only an explicitly mapped IdP role produces an Edge role, and an
        // unmapped value grants nothing. An enabled OIDC must declare at least one <role-mapping>, and this
        // check is the ONLY enforcement of that: the schema deliberately allows an absent or present-but-empty
        // <role-mappings> so a disabled, mapping-less stanza can still be written back (EDG-849).
```

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
```diff
@@ -331,46 +332,138 @@ public OidcServiceImpl(
...
+     *   <li><b>claim present, some roles mapped</b> — the login succeeds. A warning is emitted <em>only when
+     *       some IdP roles were dropped</em> (unmapped), which may be a missing {@code <role-mapping>};</li>
+     *   <li><b>claim present, non-empty, none mapped</b> — denied; the IdP roles are listed so the operator
+     *       can see they are all unmapped;</li>
+     *   <li><b>claim present but empty</b> — denied; the user genuinely carries no roles;</li>
+     *   <li><b>claim absent</b> — denied; the configured {@code role-claim-name} is not in the token, which is
+     *       likely (but not certainly) a misconfiguration, since some IdPs omit an empty claim entirely.</li>
+     * </ul>
+     * IdP <em>role</em> values are logged (they are the names the operator maps, not personal data); other
+     * claim <em>names</em> are logged in the absent case to help spot the right claim, never their values.
```
marregui: Still describes the pre-`6b62c0aca` behaviour — values aren't listed at WARN any more, and "not personal data" is the premise the commit reversed. Third doc-drift on this branch after F4: worth grepping a rule's prose in the same pass that moves the rule.

```suggestion
     *   <li><b>claim present, non-empty, none mapped</b> — denied; the number of values and the claim name are
     *       logged, so the operator can see they are all unmapped;</li>
     *   <li><b>claim present but empty</b> — denied; the user genuinely carries no roles;</li>
     *   <li><b>claim absent</b> — denied; the configured {@code role-claim-name} is not in the token, which is
     *       likely (but not certainly) a misconfiguration, since some IdPs omit an empty claim entirely.</li>
     * </ul>
     * Logging invariant: a WARN carries counts, config-supplied strings (the claim name, a mapping key) and
     * claim <em>names</em> — never claim <em>values</em>, which are token content and may be user identifiers
     * when {@code role-claim-name} is misconfigured. Values are available at DEBUG.
```

marregui (approved):
**Round 3, `2e0e05bd3`.** All six round-2 items closed. Approving.

N1 and N2 both went past the minimal change, and both were the better call. Moving the cardinality rule out of the schema also closes a variant I hadn't raised — a hand-written empty `<role-mappings>` under a disabled stanza — and puts it on the same footing as the `<users>` decision. Removing the redundant `<xs:choice>` and proving equivalence beats restoring one attribute and asserting it; I checked both heads' `config.xsd` through Xerces and every `<users>` shape is identical, with the two `role-mappings` rows flipping as intended. The new round-trip test drives the real marshaller with the schema attached, which discharges my round-2 caveat.

Two inline suggestions, prose only — commit them directly, no further round from me.

**Don't merge yet:** Estefania's green is against `eb74ac445` and `6b62c0aca` rewrote three log lines her suite asserts on. Re-run needed against this head; delta on EDG-849.

hivemq-edge/src/main/java/com/hivemq/api/auth/oidc/OidcServiceImpl.java
mschoenert-hivemq: Applied in 3002cfcee.

And you're right on the precise point: the old javadoc leaned on "these values aren't personal data," which was only ever true if role-claim-name was configured correctly — the exact thing R.13 says we can't assume. The fix doesn't make that premise true; it stops depending on it. By not logging any claim values at WARN, we've made it so we can't accidentally leak personal information there even when the claim name is misconfigured — the invariant sidesteps the premise rather than resting on it. The javadoc now states that invariant instead of the old justification.

On the drift itself — fair, and worth an apology: this is more than once now that a comment outlived the rule it described. You have every right to flag it. I'll grep a rule's prose in the same pass that moves the rule, so the comment and the code stay in step.

hivemq-edge/src/main/java/com/hivemq/api/config/OidcConfiguration.java
mschoenert-hivemq: Applied in 3002cfcee. You're right this one matters more than a comment tidy: with the XSD rule gone in a9fa3bd3d, this checkArgument is the only thing still enforcing #5, and the old "re-checked here" wording would have read as an invitation to delete it — reopening the finding. The comment now says it's the sole enforcement and why the schema deliberately allows an empty <role-mappings>.

marregui (approved):
LGTM

PR #433, opened 2024-05-22
refactor(22362): Improve lazy loading experience with Susepnse

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/22362/details/

The PR adds a `Suspense` component to every router's `Outlet`, to make the lazy loading of the content more fluid 

### Before
screenshot-localhost_3000-2024 05 22-14_08_24 (1)

### After

screenshot-localhost_3000-2024 05 22-14_07_46

PR #1707, opened 2026-07-31
EDG-835: Add OPC UA Alarms & Conditions support

mschoenert-hivemq:
## Description (the why)

OPC UA Alarms & Conditions ([EDG-835](https://linear.app/hivemq/issue/EDG-835/implement-opc-ua-alarms-and-conditions-support-in-edge)) — the condition tag end to end: subscribe, publish, acknowledge, and recover the alarm picture after a reconnect.

A condition is not a value. The server owns a state machine, and fires an event when it changes; the event is a *transition report* carrying a snapshot of the resulting state. Edge is a conduit: it forwards those transitions north and relays transition requests south, without modelling or storing the state machine.

> **Base is `master`.** [#1705](https://github.com/hivemq/hivemq-edge/pull/1705) (EDG-845) has been merged and squashed, so this is no longer a stacked PR and the diff is against master directly.
>
> *This description was corrected against the implementation on 2026-08-13, after code review v06 — the kind count, the command shape, the method count, the refresh trigger and the overflow routing had all drifted. If you read an earlier version, those five are what changed.*

## What it does

**Three new tag kinds.** A tag definition now carries a **`kind`** — how the node is observed — and a **`type`** — what structure its northbound output has. `kind: VALUE` is the default and behaves exactly as before.

**Northbound, `kind: CONDITION`.** Subscribes as an event-mode monitored item on the *notifier* its events arrive on, and the filter picks its own condition out of that notifier's traffic by `ConditionId`. Its `type` (`AlarmConditionType`, `ExclusiveLevelAlarmType`, …) decides the fields it publishes — a level alarm publishes its limits, a plain alarm does not — and is verified against the device's `HasTypeDefinition` when the tag is subscribed.

**Northbound, `kind: EVENT_SUBSCRIPTION`.** A query against a notifier: it names the notifier directly and describes which of its traffic to accept, so one tag delivers events from many conditions at once. Three narrowing predicates, each independently optional and combined with `And` — `sourceNode`, `conditionNode`, `filterType`. Omitting all three is legitimate and means the firehose. Northbound only, and refused twice: the write schema declares an object accepting nothing, and the write path rejects the tag outright.

**Northbound, `kind: REFRESH`.** Requests a `ConditionRefresh` on demand, and is where the server's **subscription-wide refresh control events** surface — `RefreshStart`, `RefreshEnd`, `RefreshRequired`. Those three are copied to every notifier item in the subscription (OPC 10000-9 §4.5), which is why one tag sees them all and why at most one is allowed, enforced at startup.

It is **not** a general loss monitor. `EventQueueOverflowEventType` is delivered only to the monitored item whose own queue filled (OPC 10000-4 §7.22.3), so a REFRESH tag sees its own item's overflow and no other. A condition or query tag losing transitions is reported as a WARN adapter event naming that tag and as `METRIC_SUBSCRIPTION_EVENT_QUEUE_OVERFLOW_COUNT` — not on the refresh tag's topic.

**Southbound.** Writing to the tag requests a state transition. All fifteen per-tag methods are exposed — Edge does not decide which of the server's own operations an operator may invoke. `method` is a string naming one of them; there is no numeric form.

**Both schemas.** The read schema is the declared type's field set; the write schema is the fixed command. Neither needs a device, since both follow from the declaration. This is the first use of the `writeSchema` that #1705 added to `DataPointSchema`.

**Refresh after a reconnect.** Transitions are events — fired and forgotten — so a condition that went active while Edge was disconnected has already emitted its transition and the server cannot re-send it. `ConditionRefresh` asks the server to synthesise a fresh report for every retained condition, and it runs after every connect and every reconnect.

## Eight things that decide the design

**`kind` and `type` answer different questions, and conflating them was the original mistake.** The definition first had one field called `type` holding `VALUE`/`CONDITION`/`EVENT_SUBSCRIPTION`, and a second called `conditionType` holding the alarm type — plus, for query tags, a third called `projectionType` naming the published shape. So the *same* concept ("what shape does this publish?") had one name on a condition tag and another on a query tag. Now `kind` says how the node is observed and `type` says what its northbound output looks like, for both. `projectionType` is retired, and `filterType` is purely a predicate that never influences the published shape. The schema rules then fall out of the kind rather than needing to be stated per tag:

| kind | read schema | write schema |
| --- | --- | --- |
| `VALUE` | derived from the device (eventually from `type`, verified) | same |
| `CONDITION` | `schema(type)` | the transition command `{method, eventId?, comment?, duration?, selectedResponse?}` |
| `EVENT_SUBSCRIPTION` | `schema(type)` | forbidden |
| `REFRESH` | the base event fields | `{method: "REFRESH"}` |

This is a **config-surface change**: `"type": "CONDITION"` becomes `"kind": "CONDITION"`, and what was `conditionType` is now `type`. Nothing outside the OPC-UA module referenced either name.

**`queueSize` means something different for an event item, and the value we were sending was the worst one.** OPC 10000-4 §7.21 gives two tables. For a *data* item `1` means "a single entry, queuing effectively disabled" — coherent for a sampled value that will be sampled again. For an *event* item `1` means "the Server returns **the minimum queue size the Server requires**". Every item was configured from the same `serverQueueSize` (default 1), so condition tags were asking for the smallest event queue the server would grant — on the one path where bursts are normal, since a notifier carries every condition beneath it and `ConditionRefresh` re-reports every retained condition at once. Measured against the embedded server: **ten transitions fired into one publishing cycle, one arrives; with a queue of 64, all ten arrive.** Hence a separate `eventQueueSize` (new config field, default 64), the parameters moved inside the item-type switch, and `getRevisedQueueSize()` compared against what was asked so a server granting less says so per tag.

**The declared type drives the schema *and* the select clause.** One source of truth: a schema promising `HighLimit` cannot coexist with a subscription that never selects it.

**A condition is not an event notifier, so the item goes above it.** `ConditionType` defines no `EventNotifier` attribute, and OPC 10000-4 is explicit that events are only available from nodes whose `SubscribeToEvents` bit is set. So the monitored item goes on a notifier and the condition is picked out of its traffic by the filter — `ConditionId`, which *is* the condition node's own NodeId, and not `SourceNode`, which is the ConditionSource (the sensor the alarm is *about*, shared by every alarm on that sensor). The notifier is taken from the tag's `notifierNode` if declared, else found by walking `HasEventSource`/`HasNotifier` upward, bounded at ten steps; if neither yields one the tag is inactive, with no implicit fallback to the Server object — that would silently widen a tag from "this condition" to "everything this server emits".

**Reconciliation is keyed by the tag, not by the node.** Milo's `userObject` slot is never touched by the SDK — not cleared by `reset()`, not sent to the server — so each notification and each monitored item carries its own identity. Node id was never an identity here: the configuration enforces unique tag *names* only, so two tags may share a definition, and a condition tag's item is created on the notifier above it rather than on the condition itself.

**The methods do not share a signature.** Three take `(EventId, Comment)`, ten take no arguments, `TimedShelve` takes a duration, `Respond` takes the index of a dialog's chosen answer, and the shelve methods live on the condition's nested `ShelvingState` rather than on the condition — falling back to the ConditionId, which §5.8.17 requires every server to accept. So the command is `{method, eventId?, comment?, duration?, selectedResponse?}` with only `method` required, and the adapter re-checks per method before calling.

**A mistyped tag fails alone.** Verification is total: a browse failure, a timeout and an interrupt all answer "do not subscribe" rather than throwing. This runs inside adapter start, where an escaping exception would abort the whole sequence and silently prevent every later adapter from starting.

**A reconnect takes two shapes, and only one recreates anything.** If the subscription transfer fails, the subscription is rebuilt and the refresh rides on re-establishing the monitored items — ordering matters, because the refresh makes the server fire a burst immediately, and asking before the items exist would lose it silently. If the transfer *succeeds*, nothing is recreated and that hook never runs, which is the common case for a brief network drop and precisely when the alarm picture is most likely to have moved unseen. Milo has no `onTransferSucceeded`, so the session activity listener reports later activations as reconnects.

## Two specification audits, and what they changed

After the implementation was complete it was audited twice against OPC 10000-4/-5/-9 — 22 findings in the first pass, 23 in the second. **All HIGH and MEDIUM findings are closed.** The audits changed the published payload in several ways a reviewer should know about, and each fix below is one commit.

**Payload corrections — these change what consumers see:**

| what | why |
|---|---|
| `namespaceIndex` is the integer it is named for | it held the whole parseable node id (`"ns=2;s=Boiler1.Temp"`) for every index except 1 — a 2023 survival from a `reversibleMode` flag dropped in 2025 |
| `LimitState` published on `ExclusiveLimitAlarmType` | its one Mandatory member (Table 96) was missing, so the standard level alarm said an alarm was active but not **which limit** tripped |
| a withheld field is `null`, with `unavailableFields` saying why | §7.22.3 lets a server substitute a StatusCode for any value; published as-is, `Severity` became a status object and a state's `id` vanished |
| `LocalTime` selected | the one field on `BaseEventType` no configuration could reach, since no type adds it |
| `method` is a string only | the integer form was Edge's own invention, promised by no schema, and `RefreshCommand` rejected what `ConditionUpdate` accepted |

**Behaviour corrections:**

- **A vendor condition type is now placed by asking the server.** OPC 10000-9 §5.5 says to expect vendor subtypes; Edge rejected them outright, because the type-hierarchy walk ran only in Edge's own table. It now follows `HasSubtype` upward until it reaches a standard ancestor.
- **Two tags with the same definition both subscribe.** Reconciliation keyed by node id silently merged them — two tags in, one monitored item out, on both the value and condition paths.
- **The subscription is recorded on every path that reports success.** A partial monitored-item sync returned `true` without recording it, so every later refresh — connect-time, reconnect, and on-demand — silently did nothing for the life of the connection.
- **A failed subscription transfer rebuilds off Milo's delivery thread.** That callback runs on the subscription's single delivery queue, so rebuilding inline stopped every notification including the keep-alives the health check reads — which then fired a reconnect against the recovery still running.
- **Three ways a condition refresh was silently lost:** a guard that could wedge shut permanently, a reconnect arriving before its hook was wired, and a timeout reported to the operator as `"verification failed: null"`.
- **Rejected select-clause entries are named.** The server answers per field at subscribe time and the answer was discarded, so "this field can never arrive" and "this device does not implement it" were indistinguishable nulls.

- [x] A `CONDITION` tag subscribes as an event item and publishes transitions
- [x] Payload carries `EventId` (the token needed to acknowledge), `Message`, `Severity`, `ActiveState`, `AckedState`
- [x] `EventId` is declared `contentEncoding: base64` on both schemas, so the encoding is stated rather than inferred
- [x] An event item's queue is deep enough to hold a burst: several transitions in one publishing cycle all publish
- [x] The published fields follow the declared condition type, not a fixed list
- [x] A condition tag publishes only its own condition's transitions
- [x] Two condition tags sharing one notifier both subscribe and both publish
- [x] Two tags with the *same definition* both subscribe
- [x] A condition tag whose notifier cannot be resolved is dropped alone, with a logged adapter event
- [x] All fifteen methods reachable, `RESPOND` included; per-method argument validation before the call
- [x] Read and write schemas both generated, and different from each other
- [x] A declared type that the device does not satisfy drops that tag only
- [x] A supertype declaration is accepted as a valid widening
- [x] A **vendor** subtype of a standard type is resolved to its standard ancestor and subscribes
- [x] A tag with no `kind` still behaves as a `VALUE` (back compatibility)
- [x] An `EVENT_SUBSCRIPTION` tag with no narrowing delivers every condition its notifier carries
- [x] Each narrowing predicate works alone, and two combined are accepted by the server as one `And`
- [x] An `EVENT_SUBSCRIPTION` tag is unwritable — declared in the schema and refused by the write path
- [x] The retained conditions are re-reported after every connect and every reconnect, including the one that keeps its subscription
- [x] No refresh is requested when no tag is an event item of any kind — a query or refresh tag is reason enough — and a server that refuses one still delivers live transitions
- [x] A `REFRESH` tag publishes the three refresh control events, and its own item's `EventQueueOverflow`; another tag's overflow is reported as an adapter event and a metric against that tag
- [x] `LocalTime` and `LimitState` are published with the shapes the schema promises
- [x] A field the server withholds is `null`, and `unavailableFields` names it with the reason

## Testing

All green on the composite; `spotlessCheck` clean.

The event subscription tests had to run against a real server rather than over the filter builder, and that is not a preference. A `ContentFilter` is a **flat array**, not a tree — a boolean operator refers to other elements *by index* in the same array — so a builder unit test could only assert the shape we intended, never that the server accepts it.

Most run against an embedded OPC UA server, because the decisive parts cannot be checked locally: the server validates the event filter, and it rejects a method call whose id is not a component of the object. Two findings came only from that — a type-level method node id is **not** callable as the *ObjectId* (§5.7.3 forbids the type node there), and a monitored item is notified of events *generally*, so without a where clause a condition tag publishes unrelated conditions.

That is narrower than it first looked, and the difference matters for the hidden-instance fallback: a type-level id is perfectly callable as the **MethodId** so long as the ObjectId stays the ConditionId, which OPC 10000-4 §5.12.2.2 Table 59 permits generally and Part 9 repeats per method only for `Enable`/`Disable`. That is the fallback shipped here for all fifteen methods, and it is asserted at the Call level with mocks rather than end to end — the embedded server exposes its condition instances, so it cannot exercise the server model the fallback exists for.

The harness was corrected before the notifier change, not after, so the control experiment means something. `TestNamespace` had made conditions event notifiers in their own right, set `SourceNode` to the condition itself, and given events random NodeIds — three fictions that between them made subscribing on the condition and filtering on `SourceNode` look correct. With the harness spec-faithful, subscribing on the condition fails all five subscription tests; under the old harness it passed all five.

**Every audit fix was control-checked the same way** — the guarantee removed, the test confirmed to fail, then restored. Two of those checks changed what the fix was:

- The reconciliation test passed against the old code until the probe was rewritten to measure item *count* rather than absence of an exception — the old code **merged** two tags rather than throwing, which is what the audit had claimed.
- A timing assertion on `onTransferFailed` passed against the inline version too, because a local server answers in milliseconds. Replaced by an assertion on *which thread* runs the rebuild, which is the property itself rather than a proxy for it.

### Notes for reviewers

`TestNamespace` needed `onEventItemsCreated` to register event items with the server's `EventNotifier`. Without it a subscription is accepted as `Good` and then silently never fed.

The same file now models conditions the way the spec does: they hang off an area notifier by `HasEventSource` rather than being notifiers themselves, their events carry the condition's own NodeId, and `SourceNode` points at a distinct ConditionSource. A harness that is wrong in the same direction as the code under test will confirm anything.

The isolation test fires the foreign alarm in a publishing cycle of its own. Fired back-to-back with one of ours it would be dropped by a server queue of depth 1 regardless of the filter, and the test would pass for the wrong reason.

**Sibling branches:** `hivemq-edge-adapter-sdk` carries a pointer branch at the same commit as its EDG-845 branch, so the composite resolves a coherent set. `hivemq-edge-test` carries two commits of its own on [PR #433](https://github.com/hivemq/hivemq-edge-test/pull/433) — `eventQueueSize` is a new property in the adapter's generated config schema and `ConfigSchemaIT` snapshot-tests it, and `OpcUaTeardownErrors` allow-lists the two ERROR logs a teardown interrupting a start now produces (see that PR for why they sit at the base classes).

## Non-Goals

- **Vendor condition subtypes** — *no longer a non-goal*; see the audit section. A vendor type is resolved to its standard ancestor.
- **Event queue overflow reporting** — *no longer a non-goal*; the loss is reported as a WARN adapter event naming the affected tag and as `METRIC_SUBSCRIPTION_EVENT_QUEUE_OVERFLOW_COUNT`, and a `REFRESH` tag publishes its own item's `EventQueueOverflowEventType`. A per-tag MQTT surface remains [EDG-856](https://linear.app/hivemq/issue/EDG-856/surface-opc-ua-event-queue-overflow-as-a-tag) — overflow is per monitored item, so no single tag can stand in for the others.
- **`GetGroupMemberships`** — returns data rather than requesting a transition, so a southbound write is the wrong shape for it
- **`eventId` in MQTT 5 Correlation Data** — belongs to the general correlation model, designed separately
- **Bounding the `TimedShelve` duration** — deliberately left to the server; capping it would be Edge second-guessing a rule the server must enforce
- **Retrying a verification timeout** — a tag whose verification times out is dropped until restart. The operator message now says so plainly rather than reading `"verification failed: null"`; how many retries and on what schedule is a separate question
- **Removing the duplicated rebuild in `onTransferFailed`** — the adapter's reconnect path already rebuilds everything this method does by hand. The blocking it caused is fixed; the duplication is [EDG-878](https://linear.app/hivemq/issue/EDG-878/opc-ua-subscription-recovery-runs-on-milos-delivery-thread-blocking)


mschoenert-hivemq:
Composite build [#3](https://jenkins.cicd.pd.hmq.dev/job/HiveMQ4/job/edge-composite/job/feature%252FEDG-835%252Fopcua-alarms-and-conditions/3/) is **SUCCESS** — 6666 tests, 0 failed, 23 skipped (13m32s).

The three condition integration tests pass against the embedded OPC UA server in CI, including `whenAnotherConditionFires_thenTheTagDoesNotPublishIt`, which is the one that exercises the where clause and fails when it is removed.

### Temporary sibling branches — delete after EDG-845 merges

The composite resolves sibling repositories by matching branch name, falling back to master. Because this work sits on top of EDG-845 and that is not yet on master, the build needs a name-matched branch in every repository EDG-845 touches, or it resolves those two to master and fails to compile.

| repository | `feature/EDG-835/opcua-alarms-and-conditions` |
|---|---|
| `hivemq-edge` | this PR |
| `hivemq-edge-adapter-sdk` | `2ea072d` — same commit as the EDG-845 branch |
| `hivemq-edge-test` | `db3707b` — same commit as the EDG-845 branch |

The latter two are pointers only, carrying no commits of their own. Once EDG-845 lands on master they become redundant and should be deleted.

caoccao (changes requested):
https://linear.app/hivemq/document/sam-code-review-v01-2df6d1c88404

The most important findings are:

1. **Event queue overflow notifications are always lost.** OPC UA delivers `EventQueueOverflowEventType` only to the monitored item whose queue overflowed. The implementation drops that event from every non-`REFRESH` item and incorrectly expects a separate `REFRESH` item to receive it. An overloaded alarm subscription can therefore lose transitions without any MQTT indication or operator event.
2. **The advertised northbound schemas do not describe the emitted JSON.** Fifteen selected standard fields have the wrong schema type, including `Comment`, which is inherited by every condition type. A valid non-null comment is emitted as an object while the schema says string. In addition, `AudibleSound` is omitted completely.
3. **A lost-wakeup race can skip the required reconnect refresh.** `onSessionActive()` and `setOnReconnect()` use two independent atomics plus a volatile callback. A specific, valid interleaving leaves `missedReconnect=true` after the callback has already been installed, with nobody left to invoke it. The retained-condition picture can remain stale after reconnect.

The PR is substantial and has useful separation between event filtering, condition commands, type verification, notifier discovery, schema generation, and JSON conversion. Its tests cover many happy paths and several failure paths. Nevertheless, the three High findings affect the core safety contract of alarms and conditions, and several Medium findings create silent data loss, schema incompatibility, or incomplete recovery. The PR should not merge until the High findings and the lifecycle/data-contract Medium findings are resolved.

marregui (approved):
Approving so the merge timing is yours. One thing I'd fix first, but I don't want to hold the PR for it.

Reviewed `af7e55b75`, not `31e192a1c` — your two newest commits are in scope. Full write-up in `EDG-835-FEEDBACK-AUG-13.md`.

### Fix before merge (my one blocker, your call on timing)

`6c0e699c3` regresses failed-start status. The compensating `setConnectionStatus(DISCONNECTED)` in `OpcUaProtocolAdapter.stop():398` runs inside `ProtocolAdapterWrapper`'s failed-start cleanup, which sets `ERROR` and then deliberately declines to overwrite it (`wasError`, `ProtocolAdapterWrapper:650/:674`). Your line overwrites it one frame earlier:

1. `wrapper.start():438` → `clearShuttingDown()`, so status writes are live.
2. `startNorthbound()` → `invokeStart` → `output.failStart(...)` (two REFRESH tags, unparseable config, any of them).
3. `invokeStart():866` rethrows; `startNorthbound()` catches, FSM → `Error`, `setErrorConnectionStatus(e, msg)`. Status is `ERROR`.
4. `start()` calls `stopNorthbound()`, `wasError = true`, invokes the adapter's `stop()`.
5. `stop():398` writes `DISCONNECTED`. `shuttingDown` is false here — `markShuttingDown()` only runs from `wrapper.stop()` — so it lands.
6. `stopNorthbound():674` then declines to write, exactly because `wasError`. The preservation has already been undone.

A mis-configured OPC UA adapter now shows "disconnected" instead of "failed". `lastErrorMessage` survives, so the text is still reachable through the API, but the status the UI colours is wrong — and it bites hardest on the duplicate-REFRESH rule, whose whole justification is that the message is unmissable.

**Suggested fix:** drop the line. The wrapper already publishes it, with the `wasError` case handled. `OpcUaStaleStatusTest` can't catch this — all four cases drive `OpcUaClientConnection` directly, so no wrapper, no `shuttingDown`, no `wasError`. One test through `ProtocolAdapterWrapper` — fail a start, assert `ERROR` — would pin it.

Two notes on the same commit. Both compensating writes (`stop():398`, `destroy():657`) are inert on the wrapper path anyway, because `markShuttingDown()` at `ProtocolAdapterWrapper:532` runs first and `setConnectionStatus` returns early while that flag is set — nothing is broken by it, but the commit message's "what changed is who reports it" isn't true in production. And v06 finding 1 as described isn't reachable: `adapter.destroy()` has exactly one call site (`ProtocolAdapterWrapper:549`), inside `stop(true)`, after `adapter.stop()` cleared `opcUaClientConnection` and completed `conn.stop()` synchronously; and a config change builds a new wrapper with a fresh `ProtocolAdapterStateImpl` (`ProtocolAdapterManager:543`), so an old connection's late write can't reach the replacement's slot. Keep the gate as an invariant — it just didn't need the compensating writes that cost the regression.

### Closed from the 2026-08-12 round

Scope (the ticket was amended today — §2/§3/§4 now match the implementation), `stop()` blocking (`markClosed()` before the monitor is exactly the shape asked for), `unavailableFields` on both sides, and refresh coalescing.

The `RefreshEnd`-release retry in `af7e55b75` is the best thing in the PR. You put the three mandatory reasons through one coordinator, found that the requeue was draining on the completion's own stack — three retries as three refusals in the same millisecond — and closed it with the server's own signal rather than a longer timeout. That's the standard.

One correction to the javadoc at `OpcUaClientConnection:80`: the residual stop wait isn't "one outstanding server call" but the remaining calls for the tag being verified — `verifyCondition` makes up to three sequential `.get(10_000)`s (`OpcUaSubscriptionLifecycleHandler:1847, 1862, 1880`) and the `abandoned` check is between tags (`:454`), not between those three. ~30s worst case, once. Fine as a bound, wrong as written.

### Your call / Martin's

Duplicate REFRESH failing the whole adapter. Your valid-vs-verifiable rationale at `:266-279` answers the consistency objection I raised last round against `NotifierResolver` — I withdraw it. What I'd still argue is narrower: if it's a validity rule it belongs where validity is decided, so the user is refused at save time rather than getting an adapter that stores fine and won't start. Not a blocker.

### Still open, low

`OpcUaTagVerificationPerKindTest` is still reflection on `verify` (`:117`) asserting only presence or absence of an Optional. And `CapturingTagStreamingService:41` still flattens every batch with `dataPoints::addAll`, so §3.2 — one message per transition — can't be pinned by any integration test. That second one is the one I'd still spend time on: a stated end-state guarantee with no test that fails when it's removed.

The manual southbound refresh is still outside the coordinator in the other direction — a user's write to a REFRESH tag during a mandatory refresh gets a bare `Bad_RefreshInProgress` with no retry (`OpcUaProtocolAdapter:910-914`). Follow-up.

### EDG-845 is not broken — checked, not assumed

The entire diff against master is inside `modules/hivemq-edge-module-opcua`: zero lines in `TagSchemaCreationOutputImpl`, `TagSchemaDirection`, `ProtocolAdaptersResourceImpl`, the OpenAPI spec, the adapter SDK or the frontend. `createTagSchema` populates the fourth component of `DataPointSchema` (`OpcUaProtocolAdapter:1001-1002`), which is EDG-845 §4 step 2 — A&C being the first populator is what that field exists for. `ConditionSchemas.unwritableSchema()` is `{"type":"object","additionalProperties":false}` rather than the empty schema, which matters because per EDG-845 §5 an empty schema is DataHub's minimal schema and validation is skipped for it entirely.

One thing I chased and cleared: `eceab0374` moves ByteString from `STRING` to `BINARY`, adding `contentEncoding: base64` to the southbound document — and after EDG-845 that document is the DataHub write-validation contract, so an added assertion there would silently change which writes are accepted. It isn't one. networknt 1.5.9 asserts `contentEncoding` under draft-07 but not under 2019-09 (I ran both), and `SchemaJsonRepresentation` stamps `https://json-schema.org/draft/2019-09/schema`. `JsonToOpcUAConverterUtil:131` already required base64 besides.

EDG-845's core tests re-run green on this branch: `TagSchemaCreationOutputImplSchemaBuilderTest`, `ProtocolAdaptersResourceImplTest`, `SimulationProtocolAdapterTest`.

### Specification delivery

Against the amended §3, verified in source: four kinds selected by `kind`, `type` separate and naming the published shape, 22 standard condition types, 15 methods including `RESPOND`, five command fields including `selectedResponse`, automatic refresh whenever any event item exists rather than only a condition tag (`OpcUaSubscriptionLifecycleHandler:1165-1169`), manual `{"method":"REFRESH"}` (`OpcUaProtocolAdapter:814-817`), the three control events published only on a REFRESH tag with overflow routed per-item to an adapter event plus `METRIC_SUBSCRIPTION_EVENT_QUEUE_OVERFLOW_COUNT` (`:1036-1043`), separate NB/SB schemas per tag, and device verification for every kind that has something to verify.

Two things I can't demonstrate from here: §3.2's one-message-per-transition, for the harness reason above, and anything black-box against a real A&C server — that's EDG-889.

On the reconnect refresh I checked the premise rather than only noting the missing test. Milo 1.1.6 fires `onSessionActive` from `SessionFsmFactory:698-716` on the transition into `Active`, and `PublishingManager:73` resumes publish requests against the reactivated session with the same subscription ids — there's no separate transfer step in between, so `onSessionReactivated` does run with the subscription live. Where it didn't survive, the refresh returns `Bad_SubscriptionIdInvalid` and `onTransferFailed` drives a rebuild whose `established()` refreshes again. The design is sound; what's missing is the end-to-end proof, and that's Estefania's in EDG-889.

### What I ran

Head `af7e55b75`, unmodified, Temurin 25. `:hivemq-edge-module-opcua:test` from the `hivemq-edge` root: **624 tests, 0 failures, 1 skipped** — up from 570 two rounds ago and 609 at v06's head. Plus the three EDG-845 core test classes. Jenkins composite **build 99 is green**; v06 listed build 96 as red and treated it as a blocker, and that's resolved at this head.

mschoenert-hivemq:
Miguel's point on the failed-start status is spot on. The compensating `setConnectionStatus(DISCONNECTED)` in `stop()` lands before `stopNorthbound()` reaches its `success && !wasError` check, so it overwrites the `ERROR` the wrapper deliberately preserves — a mis-configured adapter ends up reading "disconnected" rather than "failed". Dropping the line is right: the wrapper already publishes `DISCONNECTED` on an ordinary stop and withholds it on failed-start cleanup, which is exactly the behaviour we want.

Not critical enough to hold anything up, though — `lastErrorMessage` still carries the text, so the reason is reachable even while the status colour is wrong. **Sam, your call when to do it**, before merge or straight after.

mschoenert-hivemq:
On duplicate REFRESH tags: I agree with Miguel that they should be rejected earlier. If it is a validity rule — decidable by reading the configuration, no device involved — then it belongs where validity is decided, and a user should be refused at save time rather than saving something that stores cleanly and then will not start.

That is a larger change than it looks, though: it means the rule has to live where configuration is validated rather than where the adapter starts, and that is a different surface with its own error reporting. **So a quick follow-up rather than part of this PR.** The current behaviour is not wrong, only later than ideal.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/condition/EventPropagationVerifier.java
```diff
@@ -0,0 +1,236 @@
...
+            return new Result.OutsideHierarchy();
+        }
+        return first instanceof final Result.Unverified unverified ? unverified : (Result.Unverified) second;
+    }
+
+    private static @NotNull CompletableFuture<Result> walkUpwards(
+            final @NotNull OpcUaClient client, final @NotNull List<NodeId> origins, final @NotNull NodeId notifier) {
+        final Set<NodeId> visited = new LinkedHashSet<>(origins);
+        return walkUpwards(client, origins, notifier, visited, 0, null);
+    }
```
marregui, line 158: **Blocker.** A clean browse that returns no parents is read as proof of a mismatch.

`parentsOf` browses inverse `HasEventSource`. When that comes back Good with zero references, `walkUpwards` takes the empty-frontier branch and `finished(null)` returns `OutsideHierarchy` — a complete negative — so `verifyQueryNarrowingNode` reports the tag unsubscribable and drops it.

Sequence: a server that exposes alarms but models no area hierarchy, i.e. no `HasEventSource` above the machine node, which is a very common shape. The user configures an `EVENT_SUBSCRIPTION` tag with `node = ns=0;i=2253` and `sourceNode = <machine node>`. The browse succeeds and returns nothing, so the tag is dropped with an adapter event saying the notifier "is neither the node itself nor an ancestor in its HasCondition/HasEventSource/HasNotifier propagation hierarchy". Today that subscription is created and the events are delivered.

The Server object receives every event a server produces regardless of how the tree is modelled — which this branch already relies on: `OpcUaSubscriptionLifecycleHandler:1316` places every REFRESH item on `NodeIds.Server` "because that is the root of the notifier hierarchy", and `OpcUaEventSubscriptionIT:404-414` says so in as many words. `TestNamespace` is the tell: `areaNotifier()` needed an explicit inverse `HasNotifier` to `NodeIds.Server` for the walk to work, and `addEventNotifier(name, null)` — the constructor the five new ITs use for their roots — does not get one. No test covers `i=2253` as the query notifier.

The walk itself is good and I want to be clear I am not asking for it back: upward from the narrow node rather than down from the notifier, `HasNotifier` covered by `includeSubtypes=true`, specification and compatibility paths combined so either proves reachability, and `browseAll` throwing on a bad `BrowseResult` status so a permissions refusal lands in `Unverified` instead of a rejection. It is this one reading.

```suggestion
    private static @NotNull CompletableFuture<Result> walkUpwards(
            final @NotNull OpcUaClient client, final @NotNull List<NodeId> origins, final @NotNull NodeId notifier) {
        // The Server object is the root notifier: every event a server produces reaches it whether or not the
        // HasEventSource tree above the narrowing node is modelled or browsable. This branch already relies on
        // that -- OpcUaSubscriptionLifecycleHandler:1316 places every REFRESH item on NodeIds.Server for
        // exactly this reason -- so a walk that cannot reach it is evidence about the server's modelling
        // rather than about the query.
        if (NodeIds.Server.equals(notifier)) {
            return CompletableFuture.completedFuture(new Result.Reachable());
        }
        final Set<NodeId> visited = new LinkedHashSet<>(origins);
        return walkUpwards(client, origins, notifier, visited, 0, null);
    }
```

Worth considering separately, and not a blocker: "browsed cleanly, no `HasEventSource` parent at all" may deserve `Unverified` rather than `OutsideHierarchy` in general. A node with no ancestry is a server that does not model the hierarchy, which is the case the `Unverified` branch exists for, rather than proof of a mismatch.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaProtocolAdapter.java
```diff
@@ -753,6 +1059,176 @@ public void createTagSchema(
...
+        }
+    }
+
+    /** Resets attempt-local readiness and publishes CONNECTING as one generation-owned transition. */
+    private void publishConnectingFrom(final @NotNull OpcUaClientConnection source) {
+        connectionOwnershipLock.lock();
+        try {
+            if (opcUaClientConnection.get() == source && !stopped) {
+                browseReadiness.set(ReadinessStatus.WARMING_UP);
+                protocolAdapterState.setConnectionStatus(ProtocolAdapterState.ConnectionStatus.CONNECTING);
```
marregui, line 1166: **Blocker, and not fixable in this PR** — raised on EDG-835 for Martin.

`CONNECTING` has no route out of the product. `AdapterStatusModelConversionUtils:52-57` enumerates `DISCONNECTED`, `CONNECTED`, `ERROR` and `STATELESS`, and sends everything else to `default -> UNKNOWN`; `Status.connection` in `ext/hivemq-edge-openapi.yaml:8002-8007` has no `CONNECTING` member at all. So `GET /adapters/{id}/status` answers `UNKNOWN` for the whole connect window, and the frontend paints that red — `status-utils.spec.ts:145` pins `UNKNOWN` to `#E53E3E` and `status-mapping.utils.ts` folds it into `INACTIVE`.

`UNKNOWN` was previously unreachable. Nothing in `main` published `CONNECTING` before this commit — this line is the only site — and `ProtocolAdapterStateImpl` starts at `DISCONNECTED`. It is a new value on a REST contract, produced by one protocol only.

The window is not short either. It now spans the bounded metadata build, up to three ten-second verification phases per condition tag, and — after `3b3e48f28` — up to two more per event-subscription tag. A ten-tag A&C adapter can sit red for minutes where it previously showed `CONNECTED`.

To be explicit about what I am *not* asking for: the readiness model is right, and deleting `OpcUaConditionUpdateIT.awaitClientUsable()` is the proof — `CONNECTED` finally means usable. Keep it. What it needs is `CONNECTING` added to `Status.ConnectionEnum` and the switch, the spec regenerated and the frontend client regenerated, which is cross-repo and a decision rather than a fix. Note it also cannot be undone by publishing `DISCONNECTED` instead: that puts `startNorthbound():586-591`'s optimistic promotion straight back.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaClientConnection.java
```diff
@@ -206,31 +407,176 @@ synchronized boolean start(final ParsedConfig parsedConfig) {
...
+        if (closed.get()) {
+            task.cancel(true);
+        } else {
+            Thread.ofVirtual().name("opcua-metadata-preparation-" + adapterId).start(task);
+            if (closed.get()) {
+                task.cancel(true);
+            }
         }
+
+        final long timeoutMs = config.getConnectionOptions().connectionTimeoutMs();
```
marregui: **Blocker, also a decision rather than a fix** — raised on EDG-835.

Two things here, and neither is A&C-specific. This is the start path of *every* OPC UA adapter, including the northbound-only ones that have nothing to do with this ticket.

`warmUpBrowseMetadata` used to be a fire-and-forget `CompletableFuture` whose failure was caught, logged at WARN, and left the adapter connected with the browse endpoint answering 503. It now throws `IllegalStateException`, and `prepareClientForUse` turns that into `publishStatus(ERROR)`, a closed client and `return false` — the attempt fails and enters exponential backoff up to five minutes, indefinitely. `client.getDataTypeTree()` recursively browses the DataType hierarchy; a server that denies browse on those nodes to the configured identity, or simply answers slowly, previously cost the customer the browse endpoint and now costs them the adapter: no tags, no data, `ERROR` on the tile.

And the deadline is `connectionTimeoutMs`, whose `@ModuleConfigField` reads "Timeout for establishing connection to OPC UA server" and whose `numberMin` is `2 * 1000`. A customer who set it to 2s to fail fast against unreachable servers — a supported, documented value — has now also capped their recursive data-type-tree build at 2s, and their working adapter stops connecting after an upgrade. Nothing in the field's description tells them that.

The cancellation machinery itself is right, and I checked it rather than assuming: the `FutureTask` published before the thread starts, the `closed.get()` recheck on both sides, `markClosed()` reading the task without the start monitor. Same two-sided handshake as `subscriptionHandler`, and it holds. It is the failure policy and the borrowed setting.

Keeping the ordering change is worth it — hydrating before tags go live is correct. Either keep the pre-existing policy now that preparation runs earlier, or give it its own bounded, documented setting. If it does stay fatal it needs a release-note line, because it changes which existing deployments connect.

hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterWrapper.java
```diff
@@ -539,6 +539,9 @@ public synchronized boolean stop(final boolean destroy) {
         // Step 3 & 4: Stop connections (southbound first, then northbound)
         final boolean southboundSuccess = stopSouthbound();
         final boolean northboundSuccess = stopNorthbound();
+        // Adapter callbacks are blocked from the moment shutdown starts. Publish the wrapper-owned terminal
+        // status explicitly after teardown; failed-start cleanup does not take this path and keeps ERROR.
+        protocolAdapterState.completeShutdown();
```
marregui, line 544: Not a change request. This is the right fix in the right place, and the wrapper-driven case in `DuplicateRefreshTagTest` is exactly what was missing last round.

Two cross-adapter consequences worth a paragraph in the squash message, since `completeShutdown()` is core and every adapter goes through this line:

- a stopped stateless adapter (File, Simulation) now reports `DISCONNECTED` where it kept `STATELESS`. Restart is unaffected — the adapter republishes `STATELESS` inside `invokeStart`, before `startNorthbound()`'s `DISCONNECTED` guard is read.
- a stop whose `stopNorthbound()` failed now reports `DISCONNECTED` anyway, where it previously kept the last live status. `stop()` still returns `false`, so the caller can tell.

Separately: all eight commits today have empty message bodies. The reasoning is in the code comments, which is fine, but that leaves the squash message as the only place the change in status ownership gets recorded — and it is the part of this PR that touches other adapters.

marregui (changes requested):
Fourth round, head `3b3e48f28`. `:hivemq-edge-module-opcua:test` — **650 tests, 0 failures, 1 skipped**, up from 624. `ProtocolAdapterWrapperTest` and `com.hivemq.edge.modules.adapters.impl.*` run separately and green, since this round touches core rather than only the module. Spotless clean from the composite.

The blocker from the last round is closed, and closed in the right place: `stop()` no longer speaks for the adapter, the wrapper owns the terminal status through `completeShutdown()`, and `DuplicateRefreshTagTest` drives a real `ProtocolAdapterWrapper` and asserts the `ERROR` survives failed-start cleanup. That is the test that would have caught it in the first place. `6b5d952b7`'s token-bound retry timer is a defect you found in your own fix; `3c609a554` answered a javadoc correction by fixing the code instead, so the stop bound really is ten seconds now. Both remaining test gaps are closed — `OpcUaTransitionMessageBoundaryTest` pins §3.2 end to end through `TagManager`, which is the one I said I would still spend time on.

I also went looking for the lock-order inversion the new `connectionOwnershipLock` usually brings and did not find one: `reconnectLock` is always taken first, `stop()` and `reconnect()` release ownership before calling `conn.stop()`, and `setConnectionStatus`'s listener cannot re-enter because nothing in `main` ever registers one.

Three blockers left, and all three are in the three commits that were not asked for. One is two lines and carries a suggestion. The other two are behaviour changes that reach past this ticket — the start path of every OPC UA adapter, including northbound-only ones — and the call on both is Martin's rather than yours or mine; raised on EDG-835 rather than here.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaProtocolAdapter.java
```diff
@@ -753,6 +1059,194 @@ public void createTagSchema(
...
+     * rebuilds the tree. Reporting browse-ready on the strength of an attempt that did not manage it would be
+     * the same class of untruth as the optimistic {@code CONNECTED} this readiness model exists to remove.
+     */
+    @VisibleForTesting
+    void publishReadyFrom(final @NotNull OpcUaClientConnection source) {
+        connectionOwnershipLock.lock();
+        try {
+            if (opcUaClientConnection.get() == source && !stopped) {
+                final boolean browseReady = source.hasBrowseMetadata();
+                browseReadiness.set(browseReady ? ReadinessStatus.BROWSE_READY : ReadinessStatus.WARMING_UP);
```
marregui, line 1193: **Blocker.** Browse readiness never re-opens, so a metadata failure costs browse for the life of the connection — and the operator is told the opposite.

`browseReadiness` is written in exactly two places: `publishConnectingFrom:1130` sets `WARMING_UP`, and this line sets it once from `source.hasBrowseMetadata()`. That flag is itself latched once, at `OpcUaClientConnection:516`. `isBrowseReady():742` is a bare read of the latch. Nothing else touches either.

Sequence — a server slow to serve, or denying browse on, the DataType hierarchy, which are the two cases `8b4fd88ed`'s own message names:

1. `getDataTypeTree()` exceeds `min(connectionTimeoutMs, 30s)` → `continueWithoutBrowseMetadata` → WARN, `true`.
2. Here: `CONNECTED`, `browseReadiness` stays `WARMING_UP`. Data path fine, tags subscribed, events flowing.
3. `DeviceTagBrowsingResourceImpl:134` → **503 `AdapterBrowseNotReady` with `Retry-After: 2`**, for the rest of the connection.
4. The WARN, the adapter event and this method's javadoc all say the tree will be rebuilt by "the next browse, southbound write or reconnect" and browse will come back. A browse cannot: it is refused at `:134` before `browser.browse()` is ever reached, so the one action the operator is told to retry is the one that is structurally incapable of clearing the condition. A southbound write does rebuild Milo's tree, and does not touch `browseReadiness`. Only a new connection clears it, and a healthy `CONNECTED` adapter has no reason to make one.

So the trade this makes is not "ERROR plus backoff" → "degraded but recovering". It is "ERROR plus backoff, **which retried**" → "CONNECTED, browse permanently dead, **never retried**", while `Retry-After: 2` tells the client it is transient. `OpcUaClientConnectionTest` and `OpcUaAtomicStatusPublicationTest:200` pin the stuck state as intended, and nothing covers recovery, because there is none.

Worth placing against EDG-894: this is the unfixed half of P1. Estefania's report says the 503 "stays that way indefinitely… is never retried (a forced reconnect does not revive it)". `ca62bbf91` fixed the trigger they hit. The stuck latch is still here, and `8b4fd88ed` adds a second way to reach it — now on every OPC UA adapter, not only condition-tag ones.

Two ways out, and either closes it for me:

- **Minimum, prose only.** Correct the three messages to say the adapter must reconnect, and drop `Retry-After` on this path — a `Retry-After` that will never come good is worse than no header. Suggestions on the two message sites below.
- **Better.** Have `isBrowseReady()` ask the connection rather than read a latch, so a tree rebuilt by a southbound write actually re-opens browse.

I am not asking for the ordering to change — Martin ruled that out for this branch on EDG-894, and you kept it. This is only about what happens after the fallback fires.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaClientConnection.java
```diff
@@ -206,31 +428,216 @@ synchronized boolean start(final ParsedConfig parsedConfig) {
...
+
+    /**
+     * Reports metadata the connection could not hydrate, and carries on connecting anyway.
+     * <p>
+     * WARN rather than ERROR, and no status change, because nothing has failed that the adapter cannot do:
+     * tags will still be verified and subscribed, events and values will still flow, and the tree will be
+     * rebuilt by the first browse or southbound write that wants it. What the operator loses is a browse
+     * endpoint that answers immediately, and the message says so — including the likeliest cause, since a
+     * recursive browse of the type hierarchy is exactly the sort of thing a locked-down server account is
+     * refused.
```
marregui: Part of the blocker above — the javadoc is where the claim starts. "Rebuilt by the first browse or southbound write that wants it" is true of Milo's tree and false of the browse endpoint, and the two read as the same sentence.

```suggestion
     * WARN rather than ERROR, and no status change, because nothing has failed that the adapter cannot do:
     * tags will still be verified and subscribed, and events and values will still flow. Milo will rebuild
     * the tree for the first southbound write or schema request that wants it, but that does not restore
     * browse: {@code OpcUaProtocolAdapter} decides browse readiness once, from {@link #hasBrowseMetadata()}
     * at CONNECTED, and nothing promotes it afterwards — so the browse endpoint answers 503 until the
     * adapter reconnects. The message says so, including the likeliest cause, since a recursive browse of
     * the type hierarchy is exactly the sort of thing a locked-down server account is refused.
```

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaClientConnection.java
```diff
@@ -206,31 +428,216 @@ synchronized boolean start(final ParsedConfig parsedConfig) {
...
+                        + "hierarchy to the configured identity, or is slow to serve it, will keep reporting this.",
+                message,
+                adapterId,
+                cause);
+        eventService
+                .createAdapterEvent(adapterId, PROTOCOL_ID_OPCUA)
+                .withMessage(message + " for adapter '" + adapterId
+                        + "'. The adapter is connected; browse is not ready until the metadata is built.")
+                .withSeverity(Event.SEVERITY.WARN)
+                .fire();
```
marregui, line 681: The two operator-facing halves of the same claim. Worth re-running spotless after taking it — I have kept inside 120 columns but have not had the formatter over it.

```suggestion
        log.warn(
                "{} for adapter '{}'. The adapter stays connected and its tags are unaffected, but the browse "
                        + "endpoint reports not-ready for the rest of this connection: readiness is decided once, "
                        + "at CONNECTED, and nothing promotes it afterwards. Reconnecting the adapter is what "
                        + "retries it. A server that denies browsing the data-type hierarchy to the configured "
                        + "identity, or is slow to serve it, will keep reporting this.",
                message,
                adapterId,
                cause);
        eventService
                .createAdapterEvent(adapterId, PROTOCOL_ID_OPCUA)
                .withMessage(message + " for adapter '" + adapterId
                        + "'. The adapter is connected; browse stays not-ready until the adapter reconnects.")
                .withSeverity(Event.SEVERITY.WARN)
                .fire();
```

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaClientConnection.java
```diff
@@ -206,31 +428,216 @@ synchronized boolean start(final ParsedConfig parsedConfig) {
...
+            return continueWithoutBrowseMetadata(
+                    "OPC UA metadata preparation did not finish within " + timeoutMs + " milliseconds", e);
+        } catch (final InterruptedException e) {
+            task.cancel(true);
+            Thread.currentThread().interrupt();
+            if (closed.get()) {
+                quietlyCloseClient(client, false, serviceFaultListener, activityListener);
+                return false;
+            }
+            return continueWithoutBrowseMetadata("OPC UA metadata preparation was interrupted", e);
```
marregui: Not a blocker, and narrow — but this arm re-sets the interrupt flag at `:531` and then returns `true`, so the connect sequence carries on **on an interrupted thread**. Everything downstream that blocks — tag verification's futures, subscription creation, a lock acquisition — throws immediately, and the adapter fails the attempt anyway, further along and with a less legible cause than the interrupt it already knew about.

The other three arms are right to continue: cancellation, timeout and execution failure are all statements about the metadata build. An interrupt is a statement about *this thread*, and the honest answer to it is to abandon the attempt, whether or not `closed` is set. Cheapest version is to treat it like the `closed` branch two lines up: `quietlyCloseClient(...)` and `return false`.

No suggestion block because the surrounding lines fall outside the diff hunk.

hivemq-edge/src/main/java/com/hivemq/api/model/adapters/AdapterStatusModelConversionUtils.java
```diff
@@ -45,15 +45,27 @@ public class AdapterStatusModelConversionUtils {
                 .message(protocolAdapterWrapper.getErrorMessage());
     }
 
+    /**
+     * Translates an SDK connection status into the one the API publishes.
+     *
+     * <p>Every member is named rather than left to the default, and the default is kept only for a value the
+     * SDK might add after this is compiled. {@code CONNECTING} was previously unreachable in the product —
```
marregui: The javadoc and the code disagree about the mechanism the javadoc exists to justify: it says the default "is kept", and the diff removes it. Removing it is the right call and the commit message argues it well — the doc just hasn't caught up.

Worth stating the other half while it is being fixed, because `ProtocolAdapterState.ConnectionStatus` lives in `hivemq-edge-adapter-sdk`, a separately published artifact. An exhaustive switch expression over an enum from another jar compiles a synthetic throwing default, so a constant added to the SDK without recompiling Edge now throws `MatchException` out of `GET /adapters/{id}/status` where it previously answered `UNKNOWN`. The composite builds both together so this is a build-integrity assumption rather than a live risk, but it is exactly the scenario the current wording claims to be protecting against.

```suggestion
     * <p>Every member is named and the default arm removed, so a constant added to the SDK enum is a compile
     * error here rather than a silent downgrade. (The trade: a constant added to the SDK jar without
     * recompiling Edge throws {@code MatchException} from this method instead of answering {@code UNKNOWN} —
     * an assumption that Edge and the SDK are built together, which the composite guarantees.)
     * {@code CONNECTING} was previously unreachable in the product —
```

marregui (changes requested):
Fifth round, head `f480b2f94`. `:hivemq-edge-module-opcua:test` on Temurin 25 — **662 tests, 0 failures, 1 skipped**, up from 650. `AdapterStatusModelConversionUtilsTest` 10/0. Spotless clean on both modules from the composite, and `openApiSpecCopy` leaves `ext/hivemq-edge-openapi.yaml` untouched, so the staleness gate passes.

All three blockers from round 4 are addressed. `f480b2f94` is the Server-object exemption, at both entry points and before any browse, with `butAnOrdinaryNotifierWithNoAncestryIsStillARejection` pinning that review-08's sibling-mismatch rejection survives it — that is exactly the shape I asked for, and leaving the broader `Unverified` reading out was the right call. `ca62bbf91` is the better of the two fixes: `carriesNoDiagnostic` covers both protocol-empty shapes, the decode is contained per item, and the fence around both reporters means the next unreadable diagnostic costs a log line instead of the adapter. Estefania has since confirmed P1/P2 black-box at this head.

One blocker left, on `8b4fd88ed`. The fallback is right and I am not asking for it back — but browse readiness is a latch that never re-opens, so the adapter parks in a state three separate messages promise it will recover from and it cannot. Details inline; the minimum fix is the messages, and if you take that I will approve on it rather than hold the branch over which of the two fixes you prefer.

Two smaller things inline, plus a note on `ext/`: both behaviour changes want a release-note line, which their own commit messages call for.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/condition/EventPropagationVerifier.java
marregui: Closed by `f480b2f94`, with the suggestion taken as written and applied to both entry points. `butAnOrdinaryNotifierWithNoAncestryIsStillARejection` pins the part I cared about most — that carving out the Server object does not soften review-08's sibling-mismatch rejection. Resolving.

The non-blocking half I raised at the end — whether a clean zero-parent browse should be `Unverified` rather than `OutsideHierarchy` in general — is deliberately not in this commit and I am not asking for it here.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaProtocolAdapter.java
marregui: Decided and closed by `99d1ef9cb`: `CONNECTING` became a real status rather than the readiness alignment coming back out. The conversion switch is exhaustive, the spec member is in both the source and the bundled copy (`openApiSpecCopy` produces no diff), and `ConnectionStatusBadge` picked up the half that would have been worse than the badge colour — the object-literal lookup returning `undefined` and throwing on the next property access. Resolving.

Two residuals, neither on this thread: a one-line javadoc correction on `AdapterStatusModelConversionUtils` in this round's review, and the release note that `Status.connection` gaining a member wants. The decision itself I have taken to Martin on EDG-835, since it is his to ratify rather than mine to accept.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaClientConnection.java
marregui: Decided by `8b4fd88ed`: a metadata failure is non-fatal again, bounded at `min(connectionTimeoutMs, 30s)`, and no new configuration knob. That answers what this thread asked, and the reasoning in the commit message is right — nothing at connect time consumes the tree, and `getNamespaceTable()` performs no I/O. Resolving this one.

What the fallback leaves behind is a separate finding rather than a residual of this one, so it is on its own thread against `OpcUaProtocolAdapter:1158`: browse readiness is latched once at CONNECTED and never promoted, so the recovery the new WARN describes cannot happen and the endpoint answers 503 with `Retry-After: 2` for the life of the connection.

marregui (approved):
left some comments, looking good to me

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaProtocolAdapter.java
caoccao: Fixed in `88321f829`, taking the second option rather than the prose one — the 503 was the symptom, and correcting the messages would have left the endpoint dead.

The health check now retries the build. It is already the periodic "is this connection still all right" pass, and a build the server refused at connect time is exactly what comes good later when a permission is granted or a slow server finishes starting. Readiness is promoted through the same ownership lock as every other readiness write, so a build finishing after the adapter has moved on cannot reopen browse for a session nobody holds.

You are right that this is the unfixed half of EDG-894 P1, and it wanted more care than it first looked. Three things came out of building it:

- **`join()` is uninterruptible.** Milo's `NonBlockingLazy` waits for an in-progress `getDataTypeTree()` with `CompletableFuture.join()`, so a rehydration arriving while a southbound write is building cannot be cancelled. Releasing the guard when a *wait* expired let the next interval start a second worker against the same server, with the first no longer reachable by teardown. The slot is now claimed before the thread starts and cleared by the thread itself, so "a retry is running" is a fact about the worker rather than about a timer.
- **The deadline had to leave the caller.** Checking it when `retryBrowseMetadata` next happens to run looked equivalent and is not: the health-check interval reaches five minutes while the deadline can be two seconds. It is now a one-shot timer started with the worker, carrying that worker's identity and no shared flag, so a late timer can neither cut short a later attempt nor consume its interrupt.
- **Backoff.** A permanently restricted account would otherwise buy a complete recursive browse every interval forever. One free retry — that is the transient case, worth recovering within an interval — then a minute doubling to a half-hour cap, jittered so adapters against one server do not line up after a shared outage, and clamped after the jitter so the cap means what it says. It does not give up, because the conditions that cause this are the ones that come good.

`Retry-After: 2` is deliberately unchanged and I would rather flag it than fix it here: it lives in `DeviceTagBrowsingResourceImpl`, shared by every adapter, and making it reflect OPC UA's next attempt means adding a method to `BulkTagBrowser` in the SDK. The expensive part is now rate-limited where it belongs; the header costs a client a boolean read.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaClientConnection.java
caoccao: Fixed in `88321f829`, though not with the suggested text — and the difference is worth flagging rather than hiding.

Your wording is right for the prose-only route: browse stays shut until the adapter reconnects. Since the sibling thread ended with browse actually recovering, that sentence would have been wrong again in the other direction. The javadoc now says what reopens it — `retryBrowseMetadata`, on the health-check interval and with backoff — and keeps the part your suggestion was really about: that Milo rebuilding its tree for a southbound write does **not** restore the endpoint, because readiness is a separate decision.

It also now records the second thing the wording has to get right, which I only learned by getting it wrong in the opposite direction: this runs before the subscription, the context and `CONNECTED`, so it cannot assert a connection the attempt has not made yet.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaClientConnection.java
caoccao: Fixed in `88321f829`. Same caveat as the javadoc thread: the substance is taken, the words are not, because browse now recovers and "until the adapter reconnects" would be a fresh inaccuracy.

Both messages now say the adapter retries the build on later health checks, backing off while it keeps failing, and reopens browse when one succeeds — and both keep the sentence that mattered most in your version: **retrying the browse itself does not**, because it is refused before it runs. That was the part sending an operator at the one action that could not help.

The event message changed in a second way you did not ask for and I think you would want. It fires from inside `prepareClientForUse`, before the subscription, the context and `CONNECTED`, so "The adapter is connected" was asserting something the attempt had not earned and might never earn — subscription creation can still fail, or a stop can land — leaving an event claiming a healthy connection beside a status of `ERROR`. It is now conditional: *"The connection attempt continues; if it succeeds, browse stays not-ready until a later health check rebuilds the metadata."*

Spotless run, as you suggested.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/OpcUaClientConnection.java
caoccao: Taken, in `88321f829`, and the cheap version you described is what it does: close the client and return `false`, whether or not `closed` is set. Your framing is the reason — cancellation, timeout and execution failure are statements about the build, which the connection does not need; an interrupt is a statement about this thread.

Following it up turned out to matter more than the arm itself, in two ways.

**The order of the two lines is load-bearing.** `quietlyCloseClient` ends in `OpcUaClient.disconnect()`, which is `disconnectAsync().get()`. With the flag already restored, that `get()` throws `InterruptedException` at once, Milo wraps it as a `UaException`, and the class logs "Failed to disconnect" over a teardown that never happened — so the session was still closing while the attempt returned and a retry was scheduled. Cleanup now runs while the caught `InterruptedException` has the flag clear, and the flag is restored in a `finally`.

**The obvious test does not discriminate.** Asserting an empty connection context was true of the old behaviour too, and waiting for the server session to close passes either way, because the close does eventually complete — it just was not awaited. The assertion that separates them is the absence of that `Failed to disconnect` ERROR, which is also the thing an operator would have seen. With the ordering reverted it reproduces exactly:

```
[the disconnect must be awaited, not abandoned to finish behind a retry]
  [ERROR] Failed to disconnect: java.lang.InterruptedException
```

hivemq-edge/src/main/java/com/hivemq/api/model/adapters/AdapterStatusModelConversionUtils.java
caoccao: Taken as written, in `1b3907b62` — kept as its own commit since it is a correction to `99d1ef9cb` rather than to the browse work in the rest of this round.

You were right that the javadoc described the mechanism the same commit removed, and right that the `MatchException` trade is the half worth stating: it is exactly the scenario the old wording claimed to be protecting against, so it belongs in the record rather than in someone's stack trace. Wording is yours, with the note that the composite is what makes it a build-integrity assumption rather than a live risk.

caoccao (approved):

PR #1709, opened 2026-08-01
EDG-811: Bound the message expiry setters, normalize the "no expiry" markers at the copy boundary

mschoenert-hivemq:
## Description (the why)

A message expiry interval is a duration in seconds — except that internally "no expiry" has been spelled at least four different ways, none of which is a duration and none of which converts to the others:

| constant | value | where it comes from |
|---|---|---|
| `MAX_EXPIRY_INTERVAL_DEFAULT` | 2³² | what both PUBLISH decoders clamp an absent property to; the default `<message-expiry>` |
| `MESSAGE_EXPIRY_INTERVAL_NOT_SET` | `Long.MAX_VALUE` | `PUBLISHFactory`'s default — every internally generated publish |
| `NorthboundMapping.DEFAULT_MESSAGE_EXPIRY` | `JS_MAX_SAFE_INTEGER` (2⁵³−1) | the API/config layer |
| `TTL_DISABLED` | −1 | retained-message TTL |

Two independent defects came out of that, both fixed here.

**1. The copy boundary rejected the markers.** `PublishBuilderImpl.fromPublish` copied the source publish's interval straight into the *public, bounded* setter, so a publish that had no expiry arrived at `checkMessageExpiryInterval` as `Long.MAX_VALUE` and threw `Message expiry interval 9223372036854775807 not allowed`. That is how the bidirectional adapter's dead-letter repost silently stopped happening (observed on Edge 2026.12-SNAPSHOT, once per poll): the exception was swallowed by `validateAndWrite`'s outer catch and the message never reached `$invalid/`.

**2. The encoder truncated out-of-range values.** `encodeIntProperty` tested only for equality with the caller's default, so any *other* out-of-range value reached `writeInt((int) value)` and kept its low 32 bits — 2⁴⁰ encoding as **0** ("expire immediately"), `Long.MAX_VALUE` as 4294967295.

Linear: EDG-811

## What changed

Validation and normalization are two different jobs, done in two different places.

**The public setter stays bounded.** `PluginBuilderUtil.checkMessageExpiryInterval` accepts `0 .. min(configuredMaximum, MAX_EXPIRY_INTERVAL_DEFAULT)` and nothing else. No value — however it is spelled — gets past `<message-expiry>` through a public extension SDK setter. Two deliberate deltas from the previous behaviour:

- **Zero is now accepted.** The SDK javadoc documents only a *negative* interval as throwing, MQTT 5 permits zero, and `Mqtt5PublishDecoder` reads it without a lower-bound check. Rejecting it was itself a live instance of this bug class: a client publishing with expiry 0 and failing its southbound schema hit the same exception on the repost.
- **The four-byte cap is enforced explicitly.** `MqttConfigurator` bounds `<message-expiry>` at 2³², but only on the `config.xml` path — `MqttConfigurationServiceImpl.setMaxMessageExpiryInterval` is unchecked. Without the cap a programmatically raised maximum would let a value with no MQTT representation into the domain model.

**The copy boundary normalizes first, and bypasses the setter.** `PublishBuilderImpl.fromPublish(PUBLISH)`, `PublishBuilderImpl.fromComplete(...)` and `RetainedPublishBuilderImpl.fromComplete(...)` ask `PluginBuilderUtil.isCopyableMessageExpiryDuration(...)` whether they were handed a real four-byte duration. If yes it goes through the bounded setter as before; if no — any marker, any spelling, present or future — the canonical `MESSAGE_EXPIRY_INTERVAL_NOT_SET` is assigned to the field directly, and `build()` resolves it to the operator's configured maximum. This mirrors the guard `RemoteMqttForwarder` already applies at the same kind of boundary, and it is why the fix does not need to enumerate the four constants.

**The encoder never truncates.** `intPropertyEncodedLength` and `encodeIntProperty` share one `shouldEncodeIntProperty` predicate, so the declared property length and the bytes actually written cannot drift apart. (They did once on this branch, producing malformed packets and widespread timeouts.) The equality half of the predicate is kept because `MAXIMUM_PACKET_SIZE` passes a *genuine* default — a legal value the receiver reconstructs from the spec — where omission is a size optimisation. Packet-size and session-expiry behaviour are unchanged.

### Behaviour, per boundary

| input | public setter, `<message-expiry>` = 10 | public setter, default max (2³²) | copy boundary, `<message-expiry>` = 10 |
|---|---|---|---|
| `0 .. 10` | accepted | accepted | passed through unchanged |
| `11 .. 2³²−1` | **rejected** | accepted | **rejected** — copying is not a way around the bound |
| `2³²` (`MAX_EXPIRY_INTERVAL_DEFAULT`) | **rejected** | accepted — it *is* the maximum | → no expiry → `build()` = 10 |
| `Long.MAX_VALUE` (`MESSAGE_EXPIRY_INTERVAL_NOT_SET`) | **rejected** | **rejected** | → no expiry → `build()` = 10 |
| `JS_MAX_SAFE_INTEGER`, 2⁴⁰ | **rejected** | **rejected** | → no expiry → `build()` = 10 |
| `TTL_DISABLED` (−1) | **rejected** | **rejected** | → no expiry → `build()` = 10 |

"No expiry" resolving to the configured maximum during `build()` is deliberate and is **not** a fallback that loses information: it is the rule the rest of the broker already applies. `Mqtt5PublishDecoder` clamps an absent Message Expiry Interval property to `maxMessageExpiryInterval()`, `Mqtt3PublishDecoder` sets it unconditionally, and the extension SDK documents `<message-expiry>` as this builder's default. A publish built here therefore behaves exactly like the equivalent publish from a client. Left at the default 2³² the encoder omits the property again, so the common case still goes out as a genuine "no expiry".

- [x] `PublishBuilderImpl.fromPublish` and the shared expiry-copy boundary treat "not a real duration" as unset rather than validating it as a numeric bound.
- [x] The three (four) expiry "infinity" constants are reconciled at that boundary, without migrating any of them.
- [x] A real duration above the operator's configured maximum is **still** rejected — through the setter and through a copy.
- [x] No internal marker can pass a public SDK setter and become a no-expiry message under a finite `<message-expiry>`.
- [x] No value above the four-byte range can reach `writeInt` and be truncated.
- [x] Packet-size and session-expiry encoding are unchanged.
- [x] Regression: a rejected southbound write for a no-expiry message publishes the `$invalid/` repost without throwing — see the companion PR below.

## Non-Goals

- **Migrating the four constants to one.** The copy boundary makes them behave identically; unifying them touches config, domain and API models and belongs in its own change.
- `ModifiableOutboundPublishImpl.getMessageExpiryInterval()` returns `Optional.of(Long.MAX_VALUE)` for the sentinel while `PublishPacketImpl` and `ModifiablePublishPacketImpl` return `Optional.empty()`. Pre-existing, unrelated to this fix, and changing it changes behaviour for existing publishes.
- `NorthboundMapping.messageExpiryInterval` has no consumer in main source and appears to have no effect. Pre-existing; worth its own ticket.

## Test plan

Full `hivemq-edge` unit suite green; `:hivemq-edge:check` (Spotless + ErrorProne + NullAway + tests) green.

**Setter contract, all five shared call sites** — `PublishBuilderImpl`, `RetainedPublishBuilderImpl`, `WillPublishBuilderImpl`, `ModifiablePublishPacketImpl`, `ModifiableOutboundPublishImpl`: both markers are rejected under a finite maximum, real durations above the maximum are rejected, zero is accepted, negatives are rejected, and 2⁴⁰ / `JS_MAX_SAFE_INTEGER` are rejected even when `setMaxMessageExpiryInterval` was raised past the four-byte range.

**Build-through, not just the setter** — `test_message_expiry_default_marker_cannot_be_built_into_a_no_expiry_publish` follows 2³² under a ten-second maximum all the way to `build()`, which is where the earlier revision of this PR produced a publish that `MessageExpiryHandler` never counts down and the encoder never emits.

**Copy boundary** — `fromPublish` / `fromComplete` normalize `MESSAGE_EXPIRY_INTERVAL_NOT_SET`, `MAX_EXPIRY_INTERVAL_DEFAULT`, `JS_MAX_SAFE_INTEGER`, 2⁴⁰ and `TTL_DISABLED`; real durations stay bounded; a marker copied under a ten-second maximum comes out as ten seconds, not as a marker.

**Byte level** — `Mqtt5PublishEncoderTest` asserts the exact wire bytes for `2³²−1` (written in full) and for four out-of-range values (property absent, declared property length matching the bytes written, payload intact).

**Mutation-checked.** Reintroducing the marker exemption in the validator fails 7 tests across all five call sites; reintroducing the raw `messageExpiryInterval(publish.getMessageExpiryInterval())` copy fails the `ReactiveWriter` regression with the original `IllegalArgumentException`; reintroducing the encoder length/write desync fails 3 of the 4 out-of-range byte-level cases.

## Companion PR

[hivemq/hivemq-edge-commercial-modules#359](https://github.com/hivemq/hivemq-edge-commercial-modules/pull/359) — **merge this PR first**; that one's repost test fails against a core build without this fix.

The end-to-end acceptance criterion is covered there: `ReactiveWriterTest` now drives a queued no-expiry `PUBLISH` through a rejecting `DataHubService.validatePublish(...)` and asserts the captured `Publish` lands on `$invalid/<topic>` with the expiry resolved to the configured maximum, that the failure metric fires, the queue entry is removed, polling continues, exactly one repost is issued, and a message already on `$invalid/` is not reposted recursively.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/extensions/services/builder/PluginBuilderUtil.java
+++ b/hivemq-edge/src/main/java/com/hivemq/extensions/services/builder/PluginBuilderUtil.java
@@ -17,6 +17,7 @@ package com.hivemq.extensions.services.builder;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.hivemq.configuration.entity.mqtt.MqttConfigurationDefaults.MAX_EXPIRY_INTERVAL_DEFAULT;
 
 import com.google.common.base.Preconditions;
 import com.hivemq.extension.sdk.api.packets.general.Qos;
@@ -42,8 +43,28 @@ public class PluginBuilderUtil {
         return !validateUTF8 || !Utf8Utils.hasControlOrNonCharacter(stringToValidate);
     }
 
+    /**
+     * EDG-811: the message expiry interval has two ranges, and only one of them is a duration.
+     * <ul>
+     *     <li>{@code 0 .. 2^32-1} — a real duration, checked against the configured maximum and written to
+     *     the wire as the four-byte Message Expiry Interval property.</li>
+     *     <li>{@code >= 2^32} — "no expiry". MQTT expresses this by omitting the property, so there is no
+     *     wire value to bound; several such markers exist internally ({@code MAX_EXPIRY_INTERVAL_DEFAULT},
+     *     {@code MESSAGE_EXPIRY_INTERVAL_NOT_SET}, {@code JS_MAX_SAFE_INTEGER}) and all mean the same thing.</li>
+     * </ul>
+     * Validating the second range as if it were a duration is what broke the bidirectional adapter's
+     * dead-letter repost: copying a publish that had no expiry threw, because the "absent" marker is
+     * {@code Long.MAX_VALUE} and therefore far above any configured maximum.
+     * <p>
+     * Note the resulting asymmetry: an operator who configures a one-hour maximum rejects a client asking
+     * for two hours but accepts one asking for no expiry at all. That matches MQTT, where an absent property
+     * is the protocol default and already outlives any configured bound.
+     */
     public static void checkMessageExpiryInterval(
             final long messageExpiryInterval, final long maxMessageExpiryInterval) {
+        if (messageExpiryInterval >= MAX_EXPIRY_INTERVAL_DEFAULT) {
+            return;
+        }
         checkArgument(
                 messageExpiryInterval <= maxMessageExpiryInterval,
                 "Message expiry interval %s not allowed. Maximum = %s",
--- a/hivemq-edge/src/main/java/com/hivemq/extensions/services/builder/PluginBuilderUtil.java
+++ b/hivemq-edge/src/main/java/com/hivemq/extensions/services/builder/PluginBuilderUtil.java
@@ -18,8 +18,6 @@ package com.hivemq.extensions.services.builder;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.hivemq.configuration.entity.mqtt.MqttConfigurationDefaults.MAX_EXPIRY_INTERVAL_DEFAULT;
-import static com.hivemq.mqtt.message.publish.PUBLISH.MESSAGE_EXPIRY_INTERVAL_MAX;
-import static com.hivemq.mqtt.message.publish.PUBLISH.MESSAGE_EXPIRY_INTERVAL_NOT_SET;
 
 import com.google.common.base.Preconditions;
 import com.hivemq.codec.encoder.mqtt5.UnsignedDataTypes;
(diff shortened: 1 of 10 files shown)


caoccao (approved):

hivemq-edge/src/main/java/com/hivemq/extensions/services/builder/PluginBuilderUtil.java
```diff
@@ -42,19 +45,67 @@ public static boolean isValidUtf8String(final @NotNull String stringToValidate,
...
+     * <p>
+     * Internally "not a real duration" has been spelled at least four ways —
+     * {@link PUBLISH#MESSAGE_EXPIRY_INTERVAL_NOT_SET}, {@code MAX_EXPIRY_INTERVAL_DEFAULT} (what the decoders
+     * clamp an absent property to), {@code TTL_DISABLED}, and {@code JS_MAX_SAFE_INTEGER} from the northbound
+     * mapping defaults. A copy boundary cannot know which one it is being handed, and once the configured
+     * maximum is finite every one of them fails {@link #checkMessageExpiryInterval} — which is exactly the
+     * exception the bidirectional adapter's dead-letter repost hit. So the test is "is this an MQTT four-byte
+     * duration?" rather than an enumeration of markers, and anything that is not one collapses to the
+     * canonical marker at the caller. This mirrors the guard {@code RemoteMqttForwarder} already applies when
+     * converting a PUBLISH for a bridge client.
```
marregui, line 100: `isCopyableMessageExpiryDuration` is stricter than the guard it cites, in the direction that matters here. `RemoteMqttForwarder.convertPublishForClient:519-521` bounds only from above (`<= PUBLISH.MESSAGE_EXPIRY_INTERVAL_MAX`); this one bounds both ends via `isUnsignedInt`. They diverge on negatives — and `TTL_DISABLED` (-1) is one of the four markers this very javadoc enumerates five lines up.

Worth being exact, because the forwarder also targets `Mqtt5Publish.builder()` — the client library's validator, not this one — so it is a sixth site, outside the five the ticket enumerated. Whether a `TTL_DISABLED` publish actually reaches it, and whether the client library rejects a negative interval, I could not confirm (decompiling `hivemq-mqtt-client-1.3.17` was inconclusive). So I am not claiming a bridge bug — only that the sentence overstates the match.

```suggestion
     * canonical marker at the caller. This is the stricter form of the guard {@code RemoteMqttForwarder}
     * applies when converting a PUBLISH for a bridge client, which bounds only from above.
```

hivemq-edge/src/test/java/com/hivemq/extensions/services/builder/WillPublishBuilderImplTest.java
```diff
@@ -61,6 +63,35 @@ public void setUp() throws Exception {
         willPublishBuilder = new WillPublishBuilderImpl(service);
     }
 
+    @Test
+    public void test_message_expiry_no_expiry_markers_are_rejected_under_a_finite_maximum() {
```
marregui, line 67: These cover the setter, which is right. The copy boundary next to it is the one left untreated.

`WillPublishBuilderImpl.fromComplete:189` still does `this.messageExpiryInterval = messageExpiryInterval.orElse(NOT_SET)` — direct field assignment, no validation and no `copyMessageExpiryInterval` — while `build():300` resolves only `NOT_SET`. `PublishPacketImpl.getMessageExpiryInterval():148-154` maps only `NOT_SET` to `Optional.empty()`, so `2^32`, or a real duration above the configured maximum, arrives here and lands in the built packet unbounded.

Same reachability caveat I gave this in round 3 — it needs the maximum to change between the publish being created and being copied — so Low, and not a blocker. But it is now the only one of the three copy boundaries without the treatment, and the argument you make for the modifiable packets ("no `build()` step that resolves a marker") applies here too: `build()` resolves one of the four markers, not all of them. Either give it `copyMessageExpiryInterval` or say in the PR body why it is out of scope.

marregui (approved):
Approving. The corrective commit closes both round-3 blockers, and CI is green on this head, which it was not on the previous one.

- **The validator has no special cases left**, all five setters are pinned, and both copy boundaries are covered through `build()`. `copyMessageExpiryInterval` bypassing the setter is safe — `messageExpiryInterval(long)` only validates and assigns, so nothing is skipped. I also checked that no internal caller leaned on the removed exemption: the only non-builder `.messageExpiryInterval(...)` callers are `NorthboundMapping`'s own builder and `RemoteMqttForwarder`, and the latter targets the MQTT client library's builder, not this one.
- **The cap constant.** You took that head-on rather than leaving it implicit, and `test_message_expiry_default_marker_is_accepted_while_it_is_the_configured_maximum` pins *why* the pre-existing `test_max_message_expiry_value_validation` stays green. That is better than what I asked for.

Three residual items, none of which holds the branch — take or leave:

1. The `RemoteMqttForwarder` precedent is cited more strongly than it holds. Inline suggestion below; the same sentence is in the PR description.
2. `WillPublishBuilderImpl.fromComplete:189` is now the only copy boundary without the treatment. Inline note below.
3. The encoder change means a publish carrying `MESSAGE_EXPIRY_INTERVAL_NOT_SET` used to go on the wire as an explicit `4294967295` and now carries no property at all. The PR body has the fact in the "what was wrong" section; it is a visible change for existing subscribers and worth a release-note line.

The last substantive item — the repost's exception safety in `ReactiveWriter` — belongs to #359, so I have put it there rather than here.

One process note: the only review on this PR is an APPROVE on `323a168c3` from 2026-08-04, now three commits stale and predating both corrective commits. Worth re-requesting review so that is not what carries it through branch protection.

PR #1710, opened 2026-08-03
feat(workload): retained-identity replay verb (harness gap from testsuite PR #1558 round 2)

estefania-hivemq:
## Summary
- Adds the `retain <node>` / `emit datapointretained <node> <value>` verb pair to the workload control channel, closing the harness gap Miguel identified in the testsuite review: a plain `emit datapoint` resolves the id at injection time, so after a tag remove/recreate the injection arrives as a fresh synthetic node (`resolved=false`) rather than the stale identity the recreate tests are named for.
- `retain` snapshots the live `Node` instance while it is still resolvable (journals `RETAIN node=<id> captured=<bool>`); `datapointretained` emits with that exact retained instance (`retained=true`), and **misses loudly** (`MISS no-retained-identity`, no emission) when nothing was retained — a silent fallback would recreate the synthetic-node trap.
- `WorkloadProtocolAdapter.BUILD` bumped to `wl-2026-08-03-v2b10` per the token discipline.

## Test Plan
- [x] Module unit tests: retained replay delivers the SAME instance (`isSameAs`) after the id is pruned; the no-retain path journals the miss and emits nothing; full module suite green.
- [ ] Testsuite side (rides testsuite PR #1558): `RecreatedTagWithSameNameRejectsOldNodeCallbacksTest` re-enabled on the new verbs, premise-proved via `retained=true`.

### DIFF
--- a/modules/hivemq-edge-module-workload/src/main/java/com/hivemq/edge/adapters/workload/WorkloadControlChannel.java
+++ b/modules/hivemq-edge-module-workload/src/main/java/com/hivemq/edge/adapters/workload/WorkloadControlChannel.java
@@ -97,9 +97,12 @@ public final class WorkloadControlChannel {
      * Identities captured by the {@code retain <node>} command while they were still resolvable. Lets a test emit a
      * callback with the ORIGINAL node instance after the tag has been removed/recreated — the stale-identity replay
      * that a plain {@code emit datapoint} cannot produce, because the live resolver builds a fresh synthetic node
-     * once the id has been pruned.
+     * once the id has been pruned. STATIC and adapter-id-keyed on purpose: a recreate constructs a NEW adapter (and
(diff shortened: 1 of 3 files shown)


estefania-hivemq:
e2e validation on the testsuite side is green: RecreatedTagWithSameNameRejectsOldNodeCallbacksTest re-enabled on these verbs, premise-proven (retained=true) and passing — the engine correctly rejects the genuine stale-identity replay. Note the first run caught that per-instance retained state dies at the recreate (journaled MISS) — hence the static adapter-id-keyed registry in 26153321e. Testsuite PR #1558 now depends on this PR being in the epic build (retain proof + WL_BUILD v2b11 guards fail loudly without it), so this should merge first or together.

marregui (approved):
LGTM

PR #1713, opened 2026-08-06
EDG-862: Extract the bridge reconnect schedule into BridgeReconnectDelay

mschoenert-hivemq:
**Motivation**

[EDG-862](https://linear.app/hivemq/issue/EDG-862/replace-the-toxiproxy-bridge-tests-backoff-sleeps-with-waits-on-the)

The bridge's reconnection schedule lived inline in `BridgeMqttClient`'s disconnect listener — four private constants and a dozen lines of arithmetic. Nothing could assert it, and no test could vary it.

That second part had a cost. The Toxiproxy bridge tests break a bridge's connection and wait for it to come back, and under exponential backoff that second wait is bounded not by how quickly the bridge *can* reconnect but by how much backoff the outage accumulated. So the tests waited out the schedule with fixed 30-second sleeps, which is what made them the two longest classes in the suite. The test-side change is hivemq/hivemq-edge-test#427 and depends on this one.

**Changes**

`BridgeReconnectDelay` now owns the schedule:

- `baseDelayMs(attempts, mode)` is pure, so the sequence can be asserted exactly
- `nextDelay(attempts)` draws its own jitter and returns a `Delay(mode, baseDelayMs, jitterMs)`, so the caller logs the parts separately without knowing randomness is involved
- a `CONSTANT` mode retries at a fixed 1s with no jitter, selected by the system property `hivemq.bridge.reconnect.mode`

The switch is deliberately defended: not reachable from configuration, an unrecognised value throws rather than silently reverting to production behaviour, and the mode is logged on every reconnection attempt. A test whose switch failed to apply should fail, not flake.

**Behaviour change — the plateau moves from 120s to 128s**

`MAX_DELAY_MS` (120s) and `MAX_BACKOFF_EXPONENT` (10 → 1024s) were redundant: the former clamped at exponent 7, so the latter was unreachable and read as dead. A reader would reasonably delete whichever looked unused — and deleting the exponent cap is the dangerous one. Java shifts a long by only the low six bits of its operand, so an uncapped exponent turns negative at 54 consecutive failures and wraps back to a 1s delay at 64, inverting the backoff into the hot retry loop it exists to prevent. Around 100 minutes of continuous outage reaches that, so it is not theoretical.

`MAX_BACKOFF_EXPONENT = 7` now expresses both the policy and the guard, at the cost of only reaching ceilings that are powers of two seconds. Jitter already made "2 minutes" a range (120–150s) rather than a figure, and at seven consecutive failures 8 seconds is immaterial. `BRIDGES.md` updated to match.

**Acceptance criteria**

- [x] Reconnect delays follow 1, 2, 4, 8, 16, 32, 64, then 128s for every further attempt
- [x] The schedule is unaffected by attempt counts large enough to overflow the shift
- [x] `CONSTANT` returns exactly 1000ms with no jitter, at any attempt count
- [x] An unrecognised `hivemq.bridge.reconnect.mode` value fails loudly
- [x] The mode is read per call, so a test setting it cannot impose its schedule on others sharing the JVM
- [x] 13 unit tests pass; all `com.hivemq.bridge.*` tests pass; `spotlessCheck` clean

**Non-goals**

- Fixing the bridge defects the Toxiproxy tests reproduce
- Making the schedule configurable by deployments

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/bridge/BRIDGES.md
+++ b/hivemq-edge/src/main/java/com/hivemq/bridge/BRIDGES.md
@@ -131,9 +131,9 @@ When the bridge connection to the remote broker is lost:
 
 1. **Connection Monitoring**: The `BridgeMqttClient` detects disconnection via MQTT client listeners
 2. **Automatic Reconnection**: Exponential backoff reconnection strategy is triggered
-   - Initial delay: 1 second
-   - Maximum delay: 2 minutes
-   - 25% jitter to prevent thundering herd
+   - Initial delay: 1 second, doubling with each consecutive failure
+   - Ceiling: 128 seconds, reached after seven failures and held for every attempt after that
+   - Up to 25% jitter on top, to prevent thundering herd
 3. **Message Buffering**: Messages continue to be queued
 
 ### Two-Level Queueing System
--- a/hivemq-edge/src/main/java/com/hivemq/bridge/BridgeService.java
+++ b/hivemq-edge/src/main/java/com/hivemq/bridge/BridgeService.java
@@ -48,6 +48,21 @@ import org.slf4j.LoggerFactory;
 public class BridgeService {
     private static final @NotNull Logger log = LoggerFactory.getLogger(BridgeService.class);
 
+    /**
+     * How long {@link #updateBridges} waits for a bridge to stop before giving up on it and
+     * disconnecting its client the hard way.
+     * <p>
+     * Generous because a bridge that is mid-reconnect or draining a queue should be allowed to
+     * finish rather than be cut off; the forced disconnect below exists for the case where it never
+     * will.
+     */
+    static final long STOP_TIMEOUT_SECONDS = 30;
+
+    /**
+     * How long the forced disconnect that follows a stop timeout is itself given.
+     */
+    static final long FORCED_DISCONNECT_TIMEOUT_SECONDS = 5;
+
     private final @NotNull MessageForwarder messageForwarder;
     private final @NotNull BridgeMqttClientFactory bridgeMqttClientFactory;
     private final @NotNull ExecutorService executorService;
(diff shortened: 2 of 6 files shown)


mschoenert-hivemq:
## Status — blocked by the GitHub Actions incident, not by this change

**Verification is complete and green.** The composite Jenkins build passed on this branch (builds [#1](https://jenkins.cicd.pd.hmq.dev/job/HiveMQ4/job/edge-composite/job/feature%252FEDG-862%252Freconnect-delay-schedule/1/) and #3: 6688 tests, 0 failed), and a `CI Check` run passed in full on commit `ca1e87b5` — `check-for-changes` success, `check-backend / check` success, `Test Results` success.

What is blocking the merge is [GitHub incident qcvjkzcs7j74](https://www.githubstatus.com/incidents/qcvjkzcs7j74) (GitHub Actions, ongoing at time of writing). Two distinct symptoms hit this PR:

1. **Runners not picking up jobs** — several `check` jobs sat idle for exactly 15 minutes and were cancelled with no steps recorded. This part has since recovered.
2. **Webhook throttling** — GitHub reported processing ~15% of webhooks, so pushes were not triggering workflow runs. Commits `ca1e87b5` and `474dd095` both produced no `pull_request` run at all.

The passing run on `ca1e87b5` was triggered by `workflow_dispatch`, and branch protection still reported the required `check` context as *"Expected — Waiting for status to be reported"* — it appears to want the result from the `pull_request` event specifically, so a manual dispatch does not satisfy it.

**Auto-merge (squash) is armed**, so this will land on its own once a `pull_request`-triggered run reports green on the head commit.

### If you are picking this up

- If the head commit has no checks, push any commit (or wait for webhook throughput to recover) to get a `pull_request` run.
- **Merge order matters:** this PR must merge **before** hivemq/hivemq-edge-test#427. That PR's `BridgeReconnectSupport` imports `BridgeReconnectDelay`, which is added here — merging the test PR first will break `master`.

PR #357, opened 2024-04-08
refactor(20712): update the edit flow for the Scripts

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/20712/details/

This PR replicate the editing flow of the `Schemas` (see #334 ) to the `Scripts`. The same states and transitions are applied to the elements of the editing form (name, version and description + source ). 

```mermaid
stateDiagram-v2
    [*] --> DRAFT: create
    [*] --> VERSION_N: load

    DRAFT --> DRAFT: select type
    DRAFT --> DRAFT: edit content
    DRAFT --> DRAFT: create new name

    DRAFT --> VERSION_LATEST: select name
    VERSION_N:::namedSchema --> VERSION_LATEST: select name
    
    VERSION_N --> DRAFT: create new name
    VERSION_LATEST:::namedSchema --> DRAFT: create new name
    
    VERSION_LATEST --> VERSION_N: select version
    VERSION_N --> VERSION_N: select version

    VERSION_LATEST --> MODIFIED: edit content
    VERSION_N --> MODIFIED: edit content
    
    DRAFT --> [*]: publish
    VERSION_N --> [*]: publish
    VERSION_LATEST --> [*]: publish
    MODIFIED --> [*]: publish

  classDef namedSchema fill:#f96,color:black
```

The PR also adds the `description` property to the form.

### Before
screenshot-localhost_3000-2024 04 08-11_35_06

### After
screenshot-localhost_3000-2024 04 08-11_34_26

screenshot-localhost_3000-2024 04 08-11_34_39

PR #431, opened 2024-05-22
fix(18935): Fix the verification of connectivity limits on handles

vanch3d:
See https://hivemq.kanbanize.com/ctrl_board/57/cards/18935/details/

The PR fixes a bug with the computation of connecting edges to a handle, allowing correct restrictions on certain connections (e.g. Topic Filter's handles can only have one target)

The PR also fixes the rendering of the icon when dragging a connection from a handle

The PR finally fixes a bug when deleting nodes, whereas connected edges were not automatically deleted and stayed in the data store (not displayed but contributing to the connections count of a node)

### Before 
screenshot-localhost_3000-2024 05 22-10_07_21

screenshot-localhost_3000-2024 05 22-10_06_33

### After 
screenshot-localhost_3000-2024 05 22-10_05_02

screenshot-localhost_3000-2024 05 22-10_05_19

PR #1714, opened 2026-08-10
EDG-882: stop the periodic clean-up from clearing live bridge and sampler queues

marregui:
## What this fixes

A bridge forwarder queue is named `$FORWARDER::<bridgeId>-<digest>/<topic>`, where the digest is an MD5 of the subscription's filters and destination in **standard Base64** — an alphabet that includes `/`. Roughly one subscription in three therefore has a queue name with more than one slash and no way to tell from the string which one separates the owner from the topic.

The periodic clean-up assumed the first one did:

```java
final SharedSubscription sharedSubscription = SharedSubscriptionServiceImpl.splitTopicAndGroup(sharedQueue);
final ImmutableSet<SubscriberWithQoS> sharedSubscriber = topicTree.getSharedSubscriber(
        sharedSubscription.getShareName(), sharedSubscription.getTopicFilter());
if (sharedSubscriber.isEmpty()) {
    localPersistence.clear(sharedQueue, true, bucketIndex);
}
```

For an affected configuration that split lands inside the digest, so the job asks the topic tree about a share name that has never existed, is truthfully told nobody owns it, and clears a **live** bridge queue — every message buffered while the remote broker was unreachable, deleted, with no error and no metric. Reported by Maibornwolff on 2026.10 and 2026.11 and reproduced from their configuration.

This is a steady state, not a race. A given queue is visited every few minutes: `INTERVAL_BETWEEN_CLEANUP_JOBS_SEC` is 4, but each tick handles one bucket-and-persistence pair, there are four persistences, `CLEANUP_JOB_PARALLELISM` is 1, and the bucket count defaults to twice the processor count — about four minutes per bucket on an eight-core node. Every sweep that reaches an affected queue empties it again.

The same line had a second victim: payload sampling names its queue `$SAMPLER::<topic>/<topic>`, so for any hierarchical topic the first slash is again the wrong one and every sample was wiped on every sweep.

## What changed

**Ownership is asked, not parsed.** `ClientQueuePersistenceImpl.cleanUp` delegates to `isOrphaned`, which tries three readings — a registered bridge forwarder, the ordinary first-slash split, a sampler queue whose topic is sampled right now — and clears only when none finds an owner. The forwarder question is answered by a reference-counted index in `MessageForwarderImpl` keyed by the exact queue id, so no parsing happens. The sampler question is answered by `SamplingService.extractSampledTopic`, which uses the fact that both halves of `<topic>/<topic>` have equal length, so the shape is decidable in three checks.

**Queue names are unchanged, deliberately.** Re-encoding the digest into a slash-free alphabet is a smaller change and was rejected: it renames every persisted queue, so on upgrade the messages already in the old names become unreachable. That trades a message-loss bug for a message-loss upgrade. Every change here keeps the names byte-identical to what a 2026.11 node wrote.

**The clean-up job was not the only site that parsed a queue name.** `PublishDistributorImpl` decided a shared queue belonged to a bridge with `client.startsWith(FORWARDER_PREFIX)` and resolved which subscription's limits to apply with `contains`. A share name is the client's to choose, so a subscription to `$share/$FORWARDER::<a live forwarder id>/t` took the bridge branch: an ordinary client got a bridge's queue limit and, against a `persist=false` bridge, its QoS silently rewritten to 0. The `contains` match separately let one bridge whose id merely appears inside another's queue id answer for it. Both are now asked of the owning registry and anchored respectively. The original ticket said the housekeeping job was the only affected site; it was not, and the ticket has been corrected.

**"Nobody owns it" no longer means two different things.** The registry is legitimately empty during node start, before any bridge configuration has been applied, and during shutdown, after the bridge hook has un-registered every forwarder while the clean-up is still scheduled. Both are gated (`hasAppliedBridgeConfiguration`, and a `ShutdownHooks` check inside the reclamation loop). A third mechanism, `MessageForwarder.reserveQueues`, is ownership without a forwarder: taken at the top of every start and before every restart's stop, released once the replacements register. It is what lets a bridge that *fails* to start keep its backlog while the operator corrects the configuration.

**A colliding configuration is refused instead of silently losing a subscription.** The digest joins sorted filters with an empty separator, so `["ab","c"]` and `["a","bc"]` collide; both forwarders started and the second took the first's queues out of the index. `BridgeMqttClient.verifyForwarderIdsAreUnique` now throws before anything is registered, naming both subscriptions, their destinations and the shared id. The bridge fails to start; the node and every other bridge come up normally; nothing is discarded.

**A SUBSCRIBE can no longer take a live internal queue.** `IncomingSubscribeService` refuses a shared subscription that collides with a forwarder or sampler queue that exists at that moment. Shared subscribers take turns, so without this a client could be handed a bridge's own messages and they would never be forwarded. The `$FORWARDER::`/`$SAMPLER::` namespace stays open otherwise — reserving it outright was implemented, found to break a client's legal `$share/$SAMPLER::customer/alerts` subscription, and reverted.

**And a client that got there first is taken off it.** The refusal above cannot cover the one window that matters — a client that subscribed to the group *before* the bridge existed, when there was nothing to collide with. `MessageForwarderImpl` takes the group back when the forwarder registers, and it looks under **every** split of the queue id rather than the forwarder's own. The broker stores a shared subscription by splitting at the first slash after `$share/`, which for a slash-bearing digest falls inside the digest: the client sits on a topic-tree node the forwarder's own split never visits, while the queue it actually drains — keyed off the concatenated string — is byte-identical to the bridge's. The removal string does not depend on which split found it, since group and filter rejoin to the same queue id however it was cut.

**Restarts that nobody asked for (EDG-884).** `LocalSubscription.equals` compared a lazily-computed digest, so an unchanged file always compared as changed and the reload restarted the bridge in the queue-discarding mode. Filters and excludes are now canonicalised at construction so a reorder is not a change, and `MqttBridge.equals` compares the subscription lists as multisets for the same reason one level up. A REST update was expressed as remove-then-add, which the bridge subsystem read as "this bridge is gone" and answered with `internalStopBridge(active, clearQueue = true, retain = List.of())` — an empty retain list clears **every** queue of the bridge, including the subscriptions the request never mentioned. `BridgeExtractor.replaceBridge` makes it one transition, so it is classified as an update and keeps the surviving forwarders' queues. `persist` also joined `equals`, which `hashCode` had always included: once an update stopped being a remove-then-add, that comparison is what decides whether the bridge restarts.

**The configuration write-back stops damaging the operator's file.** Any REST write of any subsystem rewrites `config.xml` from the in-memory model, and that write dropped `<queue-limit>`, reordered topic filters, added a `<client-id>` nobody wrote, and materialised `${ENV:...}` placeholders — putting a bridge password on disk in plain text. Each of those makes the next reload see a changed bridge and restart it. It also opened the file before a schema-validating marshal, and opening truncates, so a rejected value could leave an emptied `config.xml` while the REST call answered 200. All four are preserved now, the document is rendered in full before the file is opened, and the rendered document is written beside `config.xml` and moved onto it rather than written into it — a crash or a full disk between open and close used to leave the operator with a file cut off mid-element and a node that would not start.

**A credential supplied through `${ENV:...}` stays out of that file.** The placeholders are collected from the *parsed* configuration rather than from its bytes, so what is stored is the value JAXB will hold and marshal. A numeric character reference, CRLF padding or a CDATA section all reach the restore as the value the marshaller writes, where a collector working on the source span searched the document for a string that was never in it, found nothing, and let the credential through. Where the restore genuinely cannot decide, it no longer completes the write quietly: an unrestorable credential is written as `${ENV:EDGE_UNRESTORED_SECRET}`, which no variable sets, so the node carries on with the value it already holds and the next start stops and names it; a credential still sitting in one of the file's credential-bearing elements, or a placeholder the collector could not account for at all, refuses the write and leaves `config.xml` untouched.

**And the replacement is no wider than the file it replaces.** It carries the original's owner, group, mode and any ACL, applied while it is still owner-only and verified before the move. Failing to reproduce the mode, the group or the list aborts the write rather than proceeding — those decide who *else* can read the file. The owner is the exception: changing it is privileged everywhere this runs, so it falls back to the account that wrote the file, with a warning, rather than making a configuration installed by one account and served by another unwritable. The rolling backup goes through the same sequence, having previously been a plain copy that carried the mode and left the group to whatever the process had. A file created by `FileWriter` takes the process umask and the creating process's group, so preserving the mode alone still hands group-read of every secret in the configuration to a different set of principals — on every REST write to any subsystem.

**QoS 0 sampler queues are bounded (EDG-885).** The ten-message limit was never applied to QoS 0 and the per-queue cap is skipped for shared queues, so a sampled QoS 0 topic grew until it exhausted the node-wide QoS 0 budget — a quarter of heap, shared with every QoS 0 consumer on the node. `applyMaxToQos0` is threaded through the persistence contract, defaulted to the historical behaviour and passed true only for sampler queues. Rotating that ring is no longer counted as a dropped message, which had made the one metric an operator watches for real loss useless.

## What review rounds 4 and 5 changed

Nine commits on top of the above, all in the configuration write path, and all with the same shape: a check
that asserted how something is *written down* rather than what it means.

- **The survivor sweep and the unlocatable-span check ask the parsed document's values, not its text.**
  Searching serialised XML with `contains` read markup as content: a password of `admin` was "still present"
  in any file with an `<admin-api>` element, and every configuration write was refused from then on. The
  sweep additionally asks only of credential-bearing spans, because the restore cannot move a value into a
  span it never touched — so the same string in `<username>` or `<host>` is the operator's own text, and
  refusing over it made a bridge whose password equalled its host name unwritable.
- **Access-control lists are judged by the access they grant.** Both ACL checks demanded that the list read
  back equal the list set, which is a claim about a file store's representation rather than about who can
  read the file — and the only platform that path exists for is the one nobody in the review loop can run.
  They compare per principal over ALLOW and DENY now, so reordering, merged entries and inheritance flags
  pass while an extra grant or a lost denial does not.
- **The replacement is only written where the node could have written the file in place.** Replace-by-move
  needs permission on the directory, so a write-protected `config.xml` was being replaced anyway, and
  quietly changing owner in the process.
- **A placeholder span is located on an element that carries attributes**, and placeholder tokens are
  accounted for per variable rather than by a total that could go negative and refuse every write.
- **What cannot be done is said.** A variable that resolves to nothing is reported before the write that
  drops its placeholder; a value containing `<` or `&` — which cannot be spliced into the file as text at
  all — names the variable responsible when the parse fails, and never its value; and a refused write states
  the consequence rather than logging a message-less stack trace: the node is correct, `config.xml` is
  stale, a restart would lose the change.

Verified by 5138 unit tests, a 36-test integration slice over the config and bridge paths, and 30 external
assertions against the built distribution driven through the REST API — including 160 concurrent writes
followed by a restart from the configuration those writes produced.

## What this does not fix

`stopSampling` still has **no production caller**, so a topic sampled once stays sampled for the life of the node. The method itself was broken — it passed `null` as the share name, so the removal silently did nothing — and that is fixed, but the release path needs a DELETE endpoint with a watcher token or an expiring lease, both of which change the API surface. **EDG-885 should stay open after this merges.**

Also deliberately deferred: the canonical/versioned forwarder id with a queue-name migration; relaxing the config schema so a username can be stored without a password (today the pair is dropped, costing one spurious restart); and a client that takes a *sampler's* group, which costs sample fidelity rather than customer messages.

Two more from the review rounds. The access-control path is proven for every shape a file store can return, but **nothing here executes on Windows** — every workflow in this repo is `ubuntu-latest`, and closing that needs a capability-gated test class plus one Windows job, which is a CI change with its own cost and its own ticket. And the rule that decides which elements hold a credential is a five-suffix list (`password`, `secret`, `passphrase`, `token`, `credentials`); every decision on that path turns on it, and whether anything in the schema or in an adapter configuration falls outside it has not been surveyed.

## Delivery unit

Merges together with **hivemq/hivemq-edge-test#431** and **hivemq/hivemq-edge-commercial-modules#357**.

`ClientQueueLocalPersistence` gained a parameter, and the direction of the `default` is what decides whether a mismatched pair is survivable. It sits on the **new** signature and delegates to the old one. A default protects callers of the signature it replaces and core has none of those left; what needs protecting is implementors, because `ModuleLoader` reads those jars from `HIVEMQ_HOME/modules` at run time and an older module implements only what existed when it was built. The other way round — the shape this branch carried until review — a core-only upgrade that leaves an old module in place fails with `AbstractMethodError` on **every queued publish**, not only on a sampled QoS 0 one, which on a file-native node is the whole message path.

As it stands an old module keeps working and gives up only the QoS 0 sampler bound, which is the previous behaviour, and `ClientQueueLocalPersistenceProvider` probes the loaded implementation once at start-up and says plainly what a stale module costs. `ClientQueueModuleCompatibilityTest` pins the three cases: a module predating the parameter, a current one, and one implementing half the contract.

The three still ship together — #357 carries the Xodus fixes this ticket depends on, and only that pairing gives a file-native node the bounded ring — but a partial upgrade is now a logged degradation rather than an outage.

## Scope note for reviewers

This reaches outside the bridge subsystem — `ConfigFileReaderWriter`, `EnvVarUtil`, `PublishDistributorImpl`, `IncomingSubscribeService`, `BridgeUtils`. Each of those either destroys data on its own or undoes something this branch fixes (a lossy write-back re-triggers the restart EDG-884 exists to prevent), and several were made reachable or worse by this branch. Called out here rather than left to be discovered in the diff.

47 files: 23 under `src/main`, 24 under `src/test`.

## Evidence

Four adversarial review rounds, run requirements-first rather than against a spec derived from the code, confirmed 38 defects. Rounds 1–2 found 14; round 3, run specifically over the fixes those rounds produced, found 14 more — six of them introduced by the earlier fixes, the worst being the truncating `config.xml` write; round 4 and a findings audit found 10 survivors, including `MqttBridge.equals` omitting `persist`, which this branch's own REST change had made reachable.

Three rounds of external review on top of that — 10 findings, then 22, then two more raised against the fixes for the second. Both of the last two were on the configuration write-back this branch introduced: the placeholder restore compared a raw XML source span against post-marshal output, so a character reference, CRLF padding or a CDATA section each put a credential on disk; and the replacement carried the file's mode but not the principals that mode refers to. Both are closed here.

Round 3 also produced a fix that round 4 had to revert: moving `writeConfigWithSync` outside the `BridgeExtractor` monitor closed a lock-order inversion and opened a worse one, where the config watcher could revert an operator's REST change and clear the queues the reverted configuration no longer held. It is back in master's shape with the reasoning recorded in the code, and the pre-existing inversion is written up for its own ticket.

Every fix is mutation-checked: reverted, suite re-run, the named test observed to fail and only that test. One ordering (per-bridge metrics cleared before the stop future completes) is intra-method and says so in a comment rather than claiming a test.

5131 core unit tests, 254 commercial, 1161 in the `hivemq-edge-test` suite, spotless clean. Integration coverage includes the customer's own slash-bearing digest in both persistence modes, a node restart with a file-native backlog, orphan reclamation asserted in both modes so a "fix" that simply never reaps forwarder queues would fail, colliding subscriptions, the config-file reload and REST update paths, a client that takes a forwarder's shared-subscription group, and the eviction of one stored under any split of a slash-bearing queue id.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -33,6 +33,7 @@ javassist = "3.32.0-GA"
 jaxb4-bind = "4.0.5"
 jaxb4-impl = "4.0.9"
 jctools = "4.0.6"
+jimfs = "1.3.1"
 jersey = "4.0.2"
 jose4j = "0.9.6"
 json = "20250517"
@@ -127,6 +128,7 @@ javassist = { module = "org.javassist:javassist", version.ref = "javassist" }
 jaxb4-bind = { module = "jakarta.xml.bind:jakarta.xml.bind-api", version.ref = "jaxb4-bind" }
 jaxb4-impl = { module = "org.glassfish.jaxb:jaxb-runtime", version.ref = "jaxb4-impl" }
 jctools = { module = "org.jctools:jctools-core", version.ref = "jctools" }
+jimfs = { module = "com.google.jimfs:jimfs", version.ref = "jimfs" }
 jersey-container-jdk-http = { module = "org.glassfish.jersey.containers:jersey-container-jdk-http", version.ref = "jersey" }
 jersey-hk2 = { module = "org.glassfish.jersey.inject:jersey-hk2", version.ref = "jersey" }
 jersey-media-json-jackson = { module = "org.glassfish.jersey.media:jersey-media-json-jackson", version.ref = "jersey" }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -33,8 +33,8 @@ javassist = "3.32.0-GA"
 jaxb4-bind = "4.0.5"
 jaxb4-impl = "4.0.9"
 jctools = "4.0.6"
-jimfs = "1.3.1"
 jersey = "4.0.2"
+jimfs = "1.3.1"
 jose4j = "0.9.6"
 json = "20250517"
 json-path = "2.10.0"
@@ -128,11 +128,11 @@ javassist = { module = "org.javassist:javassist", version.ref = "javassist" }
 jaxb4-bind = { module = "jakarta.xml.bind:jakarta.xml.bind-api", version.ref = "jaxb4-bind" }
 jaxb4-impl = { module = "org.glassfish.jaxb:jaxb-runtime", version.ref = "jaxb4-impl" }
 jctools = { module = "org.jctools:jctools-core", version.ref = "jctools" }
-jimfs = { module = "com.google.jimfs:jimfs", version.ref = "jimfs" }
 jersey-container-jdk-http = { module = "org.glassfish.jersey.containers:jersey-container-jdk-http", version.ref = "jersey" }
 jersey-hk2 = { module = "org.glassfish.jersey.inject:jersey-hk2", version.ref = "jersey" }
 jersey-media-json-jackson = { module = "org.glassfish.jersey.media:jersey-media-json-jackson", version.ref = "jersey" }
 jersey-media-multipart = { module = "org.glassfish.jersey.media:jersey-media-multipart", version.ref = "jersey" }
+jimfs = { module = "com.google.jimfs:jimfs", version.ref = "jimfs" }
 jose4j = { module = "org.bitbucket.b_c:jose4j", version.ref = "jose4j" }
 json-path = { module = "com.jayway.jsonpath:json-path", version.ref = "json-path" }
 json-schema-inferrer = { module = "com.github.saasquatch:json-schema-inferrer", version.ref = "json-schema-inferrer" }
--- a/hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
@@ -73,6 +73,13 @@ public class MessageForwarderImpl implements MessageForwarder {
     private final @NotNull Set<String> notEmptyQueues;
     private final @NotNull Map<String, MqttForwarder> forwarders;
     private final @NotNull Map<String, Set<String>> queueIdsForForwarder;
+    /**
+     * How many registered forwarders own each queue ID. Exactly the multiset union of the values of
+     * {@link #queueIdsForForwarder}, maintained so {@link #isForwarderQueue(String)} answers in
+     * constant time instead of scanning every forwarder.
+     */
+    private final @NotNull Map<String, Integer> forwarderQueueRefs;
+
     private final @NotNull ExecutorService executorService;
     private final @NotNull Lock pollLock;
     private volatile boolean polling;
@@ -92,6 +99,7 @@ public class MessageForwarderImpl implements MessageForwarder {
         this.notEmptyQueues = new ConcurrentSkipListSet<>();
         this.forwarders = new ConcurrentHashMap<>(0);
         this.queueIdsForForwarder = new ConcurrentHashMap<>(0);
+        this.forwarderQueueRefs = new ConcurrentHashMap<>(0);
         this.pollLock = new ReentrantLock();
         final int threadCount = InternalConfigurations.BRIDGE_MESSAGE_FORWARDER_POOL_THREADS_COUNT.get();
         this.executorService =
@@ -137,6 +145,24 @@ public class MessageForwarderImpl implements MessageForwarder {
         return FORWARDER_PREFIX + forwarderId + "/" + topic;
     }
 
+    /**
+     * Claims one reference on each queue ID. Every {@code retain} is paired with exactly one
+     * {@link #release(Set)} of the same set — the value a {@code put} displaces, or the value the
+     * final {@code remove} returns — which is what keeps the counts exact under any interleaving.
+     */
+    private void retain(final @NotNull Set<String> queueIds) {
+        for (final String queueId : queueIds) {
+            forwarderQueueRefs.merge(queueId, 1, Integer::sum);
+        }
+    }
+
+    /** Drops one reference on each queue ID, removing the entry when the last owner lets go. */
+    private void release(final @NotNull Set<String> queueIds) {
+        for (final String queueId : queueIds) {
+            forwarderQueueRefs.computeIfPresent(queueId, (key, count) -> count == 1 ? null : count - 1);
+        }
+    }
+
     @Override
     public void addForwarder(final @NotNull MqttForwarder mqttForwarder) {
         final String forwarderId = mqttForwarder.getId();
@@ -156,10 +182,33 @@ public class MessageForwarderImpl implements MessageForwarder {
             queueIdsBuilder.add(createQueueId(forwarderId, topic));
         }
         final ImmutableSet<@NotNull String> queueIds = queueIdsBuilder.build();
-        // register the queue IDs before anything else can observe the queues: the periodic clean-up
+        // Ownership is registered before anything else can observe the queues: the periodic clean-up
         // clears forwarder queues it finds unowned, so a pre-existing persisted queue must never be
-        // visible while its forwarder is mid-registration
-        queueIdsForForwarder.put(forwarderId, queueIds);
+        // visible while its forwarder is mid-registration.
+        //
+        // The statement order below is load-bearing. None of these four rules can be pinned by a test
+        // without a seam into this class -- they are intra-method orderings only observable from
+        // another thread -- so they are enforced here and in review. Getting any of them wrong loses
+        // customer messages silently.
+        //
+        //   O0  ownership is claimed before the topicTree.addTopic loop, or a persisted queue becomes
+        //       pollable while still reading as unowned.
+        //   O1  retain() runs before the put. forwarderQueueRefs -- not queueIdsForForwarder -- is the
+        //       map isForwarderQueue reads, so it, not the put, is the moment registration takes
+        //       effect. Publishing first opens a window in which the clean-up clears a live queue.
+        //   O2  retain(new) runs before release(superseded). For a queue in both sets the count would
+        //       otherwise go 1 -> absent -> 1, and a sweep landing in that instant clears a queue that
+        //       is live, registered and in the current sweep set.
+        //   O3  see removeForwarder.
+        //
+        // Re-registration takes the difference rather than skipping: one forwarder ID does not imply
+        // one queue set, because the ID embeds a digest over the filters joined with an empty
+        // separator, so {"ab","c"} and {"a","bc"} share an ID with different queue sets.
+        retain(queueIds);
+        final Set<String> superseded = queueIdsForForwarder.put(forwarderId, queueIds);
+        if (superseded != null) {
+            release(superseded);
+        }
         for (final String topic : mqttForwarder.getTopics()) {
             topicTree.addTopic(
                     clientId,
@@ -294,7 +343,14 @@ public class MessageForwarderImpl implements MessageForwarder {
                 }
             }
         }
-        queueIdsForForwarder.remove(forwarderId);
+        // O3: drop the registration first, release its queues second. The reverse order would let the
+        // reference count reach zero while the forwarder is still published and still polling, so a
+        // clean-up sweep in that window would clear a live queue. The set the map returns -- not
+        // mqttForwarder.getTopics() -- is what was actually registered, and is what must be released.
+        final Set<String> removedQueueIds = queueIdsForForwarder.remove(forwarderId);
+        if (removedQueueIds != null) {
+            release(removedQueueIds);
+        }
         final MqttForwarder removed = forwarders.remove(forwarderId);
         if (removed != null) {
             removed.stop();
@@ -368,14 +424,18 @@ public class MessageForwarderImpl implements MessageForwarder {
         });
     }
 
+    /**
+     * Whether any registered forwarder owns this queue.
+     * <p>
+     * A reference count rather than a queue-ID-to-owner map: a share name may contain '/', so
+     * concatenating it with the topic filter is not injective and two forwarders can in principle mint
+     * the same queue ID. A plain map would drop the entry when the first of them unregisters, and the
+     * periodic clean-up would then clear a queue the second still owns. Counting owners is exact for
+     * every input without depending on that argument.
+     */
     @Override
     public boolean isForwarderQueue(final @NotNull String queueId) {
-        for (final Set<String> queueIds : queueIdsForForwarder.values()) {
-            if (queueIds.contains(queueId)) {
-                return true;
-            }
-        }
-        return false;
+        return forwarderQueueRefs.containsKey(queueId);
     }
 
     @Override
--- a/hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
@@ -723,6 +723,16 @@ public class MessageForwarderImpl implements MessageForwarder {
             // single writer and the answer can have changed by now: removeForwarder releases ownership
             // before it prunes, so re-checking here is what keeps an entry for a retired queue out of the
             // set for good (EDG-882 review v02, R2-06).
+            //
+            // isForwarderQueue is true for a queue that is only *reserved* -- held for a bridge that is
+            // starting or that failed to start -- and that is deliberate, both here and at the caller:
+            // a publish arriving mid-start must be queued and noted, or the hand-over would drop it.
+            // The accepted cost is that a bridge which never starts leaves its queue ids in this set for
+            // the life of the node. Bounded by the bridge's configured subscriptions, one string each,
+            // and it costs nothing to poll: pollForBuffer iterates registered forwarders, so a set entry
+            // nobody owns is never visited. The alternative -- deriving the forwarder from the queue id
+            // to see whether it is registered -- is the queue-name parsing this whole ticket exists to
+            // remove, so it is not worth having (EDG-882 QA, 2026-08-25).
             if (!isForwarderQueue(queueId)) {
                 return null;
             }
@@ -783,11 +793,32 @@ public class MessageForwarderImpl implements MessageForwarder {
 
     private void checkBuffersAfterLock() {
         if (notEmptyQueues.isEmpty()) {
-            if (log.isTraceEnabled()) {
-                log.trace("No queues to poll, ending polling cycle");
+            // Under the lock, and only when nothing asked for another pass. checkBuffers answers
+            // "already polling" by setting pollAgain and returning, so between the emptiness check above
+            // and clearing the flag there is a window in which another thread adds a queue, sees polling
+            // still true, sets pollAgain -- and then this cleared polling without ever looking at it.
+            // The wake-up was lost: the queue sits in notEmptyQueues with no cycle running, and nothing
+            // polls it until the next unrelated message or a reconnect. On a bridge whose last message
+            // before a quiet period lands in that window, that is a message that simply waits.
+            //
+            // This mirrors the end of the poll callback below, which has always done it this way.
+            // Pre-existing (EDG-882 QA, 2026-08-25); the branch made these paths busier.
+            pollLock.lock();
+            try {
+                if (!pollAgain) {
+                    if (log.isTraceEnabled()) {
+                        log.trace("No queues to poll, ending polling cycle");
+                    }
+                    polling = false;
+                    return;
+                }
+                pollAgain = false;
+            } finally {
+                pollLock.unlock();
             }
-            polling = false;
-            return;
+            if (log.isTraceEnabled()) {
+                log.trace("pollAgain was set while the queue set read empty, polling again");
+            }
         }
 
         final int forwarderCount = forwarders.size();
--- a/hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
@@ -403,14 +403,15 @@ public class MessageForwarderImpl implements MessageForwarder {
                 }
             }
         }
-        // O3: drop the registration first, release its queues second. The reverse order would let the
-        // reference count reach zero while the forwarder is still published and still polling, so a
-        // clean-up sweep in that window would clear a live queue. The set the map returns -- not
-        // mqttForwarder.getTopics() -- is what was actually registered, and is what must be released.
-        final Set<String> removedQueueIds = queueIdsForForwarder.remove(forwarderId);
-        if (removedQueueIds != null) {
-            release(removedQueueIds);
-        }
+        // O3: unpublish and stop the forwarder first, drop its ownership second. Ownership is what the
+        // periodic clean-up reads, so releasing it while the object is still in `forwarders` and still
+        // polling leaves a live queue reading as unowned -- a sweep landing in that window clears a
+        // queue that is being filled and drained as it does so. The reverse order is safe: for the
+        // instant between the stop and the release the queue reads as owned by a forwarder that has
+        // gone, which costs one clean-up cycle and no messages.
+        //
+        // The set the map returns -- not mqttForwarder.getTopics() -- is what was actually registered,
+        // and is what must be released.
         final MqttForwarder removed = forwarders.remove(forwarderId);
         if (removed != null) {
             removed.stop();
@@ -422,6 +423,10 @@ public class MessageForwarderImpl implements MessageForwarder {
             }
         } else {
             log.warn("Attempted to remove forwarder '{}' but it was not found in active forwarders", forwarderId);
+        }
+        final Set<String> removedQueueIds = queueIdsForForwarder.remove(forwarderId);
+        if (removedQueueIds != null) {
+            release(removedQueueIds);
         }
     }
 
--- a/hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
@@ -392,26 +392,54 @@ public class MessageForwarderImpl implements MessageForwarder {
             final @NotNull Set<String> queueIds,
             final @NotNull List<String> subscribedTopics) {
         log.error("Registration of forwarder '{}' failed, undoing it", forwarderId);
+        // A try per phase, not one around all three. Sharing a single try meant a stop() that threw --
+        // an ordinary outcome, since it drains buffers through persistence futures -- skipped the
+        // topic-tree removal and the poll-set prune, leaving this forwarder subscribed to topics it can
+        // no longer serve. Ownership was released either way in the finally, so a queue was never
+        // stranded; the subscriptions were (EDG-882 review v03, R3-04).
         try {
             if (forwarders.remove(forwarderId, mqttForwarder)) {
                 mqttForwarder.stop();
             }
-            for (final String topic : subscribedTopics) {
-                topicTree.removeSubscriber(clientId, topic, shareName);
-            }
-            notEmptyQueues.removeAll(queueIds);
         } catch (final Throwable rollbackFailure) {
             // Reported, never rethrown: the caller must see why the registration failed, not why the
             // clean-up after it did.
             log.error(
-                    "Undoing the failed registration of forwarder '{}' did not complete", forwarderId, rollbackFailure);
-        } finally {
+                    "Stopping forwarder '{}' while undoing its failed registration failed",
+                    forwarderId,
+                    rollbackFailure);
+        }
+        for (final String topic : subscribedTopics) {
+            try {
+                topicTree.removeSubscriber(clientId, topic, shareName);
+            } catch (final Throwable rollbackFailure) {
+                log.error(
+                        "Removing forwarder '{}' from the topic tree for topic '{}' while undoing its failed registration failed",
+                        forwarderId,
+                        topic,
+                        rollbackFailure);
+            }
+        }
+        try {
+            notEmptyQueues.removeAll(queueIds);
+        } catch (final Throwable rollbackFailure) {
+            log.error(
+                    "Pruning the poll set of forwarder '{}' while undoing its failed registration failed",
+                    forwarderId,
+                    rollbackFailure);
+        }
+        try {
             // Last, and unconditionally: ownership is what the periodic clean-up reads, so releasing it
             // before the forwarder is unpublished would expose a live queue, and not releasing it at all
             // would keep a queue nobody drains alive for the life of the node.
             if (queueIdsForForwarder.remove(forwarderId, queueIds)) {
                 release(queueIds);
             }
+        } catch (final Throwable rollbackFailure) {
+            log.error(
+                    "Releasing the queue ownership of forwarder '{}' while undoing its failed registration failed",
+                    forwarderId,
+                    rollbackFailure);
         }
     }
 
@@ -579,16 +607,39 @@ public class MessageForwarderImpl implements MessageForwarder {
         // registration passed to addTopic.
         final Set<String> registeredQueueIds = queueIdsForForwarder.get(forwarderId);
 
+        // Every phase below runs, whatever the ones before it did. A teardown that stops half way leaves
+        // this forwarder's queue IDs in the ownership index with no forwarder behind them: the periodic
+        // clean-up then reads those queues as owned for the life of the node -- so they are never
+        // reclaimed -- and addForwarder refuses a replacement under the same id, because putIfAbsent
+        // still finds the abandoned claim. A bridge that cannot be restarted and storage that cannot be
+        // freed, out of one exception during a stop. The failures are collected and rethrown at the end
+        // so the caller still learns, but they no longer decide how much of the teardown happens
+        // (EDG-882 review v03, R3-04).
+        //
+        // The ordering the phases run in is unchanged and still load-bearing: unpublish and stop before
+        // ownership is dropped, prune the poll set last. See the note above the removal below.
+        final List<Throwable> teardownFailures = new ArrayList<>();
+
         for (final String topic : mqttForwarder.getTopics()) {
-            topicTree.removeSubscriber(clientId, topic, FORWARDER_PREFIX + forwarderId);
+            try {
+                topicTree.removeSubscriber(clientId, topic, FORWARDER_PREFIX + forwarderId);
+            } catch (final Throwable t) {
+                teardownFailures.add(t);
+                log.error("Removing forwarder '{}' from the topic tree for topic '{}' failed", forwarderId, topic, t);
+            }
             final String queueId = createQueueId(forwarderId, topic);
             if (clearQueue) {
-                final var qPersistence = queuePersistence.get();
-                if (qPersistence != null) {
-                    qPersistence.clear(queueId, true); // clear up queue
-                    if (log.isTraceEnabled()) {
-                        log.trace("Cleared queue '{}' for forwarder '{}'", queueId, forwarderId);
+                try {
+                    final var qPersistence = queuePersistence.get();
+                    if (qPersistence != null) {
+                        qPersistence.clear(queueId, true); // clear up queue
+                        if (log.isTraceEnabled()) {
+                            log.trace("Cleared queue '{}' for forwarder '{}'", queueId, forwarderId);
+                        }
                     }
+                } catch (final Throwable t) {
+                    teardownFailures.add(t);
+                    log.error("Clearing queue '{}' of forwarder '{}' failed", queueId, forwarderId, t);
                 }
             }
         }
@@ -612,12 +663,23 @@ public class MessageForwarderImpl implements MessageForwarder {
         // permanent leak is the wrong direction for this ticket.
         final MqttForwarder removed = forwarders.remove(forwarderId);
         if (removed != null) {
-            removed.stop();
-            if (log.isInfoEnabled()) {
-                log.info(
-                        "Forwarder '{}' removed and stopped, total active forwarders: {}",
+            // RemoteMqttForwarder.stop() drains its buffers through reset callbacks that wait on
+            // persistence futures and rethrow an ExecutionException as a RuntimeException, so this
+            // throwing is an ordinary consequence of a persistence hiccup, not a contrived case.
+            try {
+                removed.stop();
+                if (log.isInfoEnabled()) {
+                    log.info(
+                            "Forwarder '{}' removed and stopped, total active forwarders: {}",
+                            forwarderId,
+                            forwarders.size());
+                }
+            } catch (final Throwable t) {
+                teardownFailures.add(t);
+                log.error(
+                        "Stopping forwarder '{}' failed; its ownership and poll state are still being released",
                         forwarderId,
-                        forwarders.size());
+                        t);
             }
         } else {
             log.warn("Attempted to remove forwarder '{}' but it was not found in active forwarders", forwarderId);
@@ -642,6 +704,17 @@ public class MessageForwarderImpl implements MessageForwarder {
         for (final String queueId :
                 requireNonNullElseGet(registeredQueueIds, () -> queueIdsOf(forwarderId, mqttForwarder.getTopics()))) {
             notEmptyQueues.remove(queueId);
+        }
+
+        // Reported only once every phase above has run. The caller still sees that the teardown was not
+        // clean -- BridgeService logs it and carries on to the next forwarder -- but by this point the
+        // index is in the state it would have reached anyway, so nothing is left claiming a queue.
+        if (!teardownFailures.isEmpty()) {
+            final RuntimeException aggregate = new IllegalStateException(String.format(
+                    "Forwarder '%s' was removed and its queue ownership released, but %d step(s) of the teardown failed",
+                    forwarderId, teardownFailures.size()));
+            teardownFailures.forEach(aggregate::addSuppressed);
+            throw aggregate;
         }
     }
 
--- a/hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
@@ -17,7 +17,9 @@ package com.hivemq.bridge;
 
 import static com.hivemq.configuration.service.InternalConfigurations.FORWARDER_POLL_THRESHOLD_MESSAGES;
 import static com.hivemq.configuration.service.InternalConfigurations.PUBLISH_POLL_BATCH_SIZE_BYTES;
+import static java.util.Objects.requireNonNullElse;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.util.concurrent.FutureCallback;
@@ -45,6 +47,7 @@ import dagger.Lazy;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -566,10 +569,16 @@ public class MessageForwarderImpl implements MessageForwarder {
                     clearQueue);
         }
 
+        // Read before anything is torn down: what was actually registered under this id, which is what
+        // has to be released and what has to be pruned from notEmptyQueues. mqttForwarder.getTopics() is
+        // only what the caller believes it registered, and the two can differ (EDG-882 review v02,
+        // R2-05). The topic-tree removal below still uses the caller's topics, because that is what the
+        // registration passed to addTopic.
+        final Set<String> registeredQueueIds = queueIdsForForwarder.get(forwarderId);
+
         for (final String topic : mqttForwarder.getTopics()) {
             topicTree.removeSubscriber(clientId, topic, FORWARDER_PREFIX + forwarderId);
             final String queueId = createQueueId(forwarderId, topic);
-            notEmptyQueues.remove(queueId);
             if (clearQueue) {
                 final var qPersistence = queuePersistence.get();
                 if (qPersistence != null) {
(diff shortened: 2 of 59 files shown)


hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
```diff
@@ -363,6 +368,16 @@ public void messageAvailable(final @NotNull String queueId) {
         });
     }
 
+    @Override
+    public boolean isForwarderQueue(final @NotNull String queueId) {
+        for (final Set<String> queueIds : queueIdsForForwarder.values()) {
```
caoccao: The performance of this algo is poor. I suggest a whole new rewrite.

hivemq-edge/src/main/java/com/hivemq/sampling/SamplingService.java
```diff
@@ -70,9 +71,34 @@ public void stopSampling(final @NotNull String topic) {
         localTopicTree.removeSubscriber(clientId, topic, null);
     }
 
+    /**
+     * The queue backing a sampled topic. The share name is {@code $SAMPLER::<topic>}, so the topic
+     * appears twice and, when it contains a '/', the share-name boundary is not the first slash.
+     */
+    public static @NotNull String createQueueId(final @NotNull String topic) {
+        return SAMPLER_PREFIX + topic + "/" + topic;
```
caoccao, line 191: This algo is a bit hacky. I'd like to know why this is the only or the best solution.

caoccao (changes requested):
I hold some concerns.

caoccao (changes requested):
https://linear.app/hivemq/document/sam-code-review-v01-1822d89349b7

caoccao (changes requested):
https://linear.app/hivemq/document/sam-code-review-v02-4e971af43adc

caoccao:
No—the direct `prefix-${ENV:PW}` case is fixed, but the implementation remains vulnerable.

### High: XML normalization bypasses the fail-closed check

The collector stores the raw XML source span:

```java
final String literal = matcher.group(2);
final String value = renderOrNull(literal);
```

Later, it compares that raw representation with JAXB’s normalized output:

```java
if (result.contains(first.value()) ||
        result.contains(escapeXmlText(first.value()))) {
    throw new UnrecoverableException(false);
}

continue; // marshalled secret is written
```

XML parsing can change the representation without changing the value. For example:

```xml
<password>prefix&#45;${ENV:PW}</password>
```

The collector records `prefix&#45;s3cr3t`, but JAXB writes `prefix-s3cr3t`. Neither fallback check matches, so the code reaches `continue` and writes the secret.

CRLF normalization has the same effect:

```xml
<password>\r\n  ${ENV:PW}\r\n</password>
```

XML converts CRLF to LF before JAXB marshals it, again bypassing the fallback.

Relevant code: [[EnvVarUtil.java](https://github.com/hivemq/hivemq-edge/blob/8327d448b2cada68e17e643790cf1801f31ae48f/hivemq-edge/src/main/java/com/hivemq/util/render/EnvVarUtil.java#L176-L195)](https://github.com/hivemq/hivemq-edge/blob/8327d448b2cada68e17e643790cf1801f31ae48f/hivemq-edge/src/main/java/com/hivemq/util/render/EnvVarUtil.java#L176-L195) and [[the fallback](https://github.com/hivemq/hivemq-edge/blob/8327d448b2cada68e17e643790cf1801f31ae48f/hivemq-edge/src/main/java/com/hivemq/util/render/EnvVarUtil.java#L298-L346)](https://github.com/hivemq/hivemq-edge/blob/8327d448b2cada68e17e643790cf1801f31ae48f/hivemq-edge/src/main/java/com/hivemq/util/render/EnvVarUtil.java#L298-L346).

### High: valid XML forms are still invisible to the regex collector

This pattern cannot recognize CDATA or elements with attributes:

```java
Pattern.compile(
    "<([A-Za-z_][\\w.:-]*)>" +
    "([^<>]*\\$\\{ENV:[^}]*}[^<>]*)" +
    "</\\1>");
```

This valid configuration still leaks:

```xml
<password><![CDATA[prefix-${ENV:PW}]]></password>
```

The collector returns no placeholder, so restoration never runs. The attribute pattern also handles only double-quoted attributes, missing valid single-quoted ones.

### Verification

I ran the complete focused `ConfigWriteBackEnvVarTest` at commit `8327d448…` with three additional end-to-end cases:

- Numeric character reference: failed; secret written.
- CRLF padding: failed; secret written.
- CDATA: failed; secret written.

The other 17 submitted tests passed.

### Recommended correction

Use an XML-aware collector before environment substitution, storing normalized text/attribute values and a structural locator. Avoid comparing raw XML source spans against post-JAXB output.

As an immediate safety measure:

- Abort whenever a secret-bearing collected span cannot be restored; do not conditionally `continue`.
- Count all `${ENV:...}` tokens and abort a write if any token was not accounted for by the collector.

So I would still request changes on this commit.

caoccao:
This commit is partially correct. The forwarder teardown and in-memory sampler fixes look sound, but the configuration-permission fix still has a high-severity gap.

### High: replacement preserves mode bits, but not actual access control

The implementation copies only `Set<PosixFilePermission>`:

```java
final Set<PosixFilePermission> targetPermissions =
        posixPermissionsOf(target);

createPartialFile(partial, targetPermissions);
// write secrets
applyPermissions(partial, targetPermissions);
Files.move(partial, target, ...);
```

The temporary file retains its newly assigned owner and group. ACLs and other security attributes are also lost. Consequently, applying `0640` may grant read access to a completely different group than the original `0640` file.

I reproduced this end to end:

```text
expected group: everyone
actual group after write: staff
```

The mode remained `0640`, but the principals receiving group-read access changed. There is both:

- A disclosure window after `applyPermissions()` and before `move()`.
- A permanent access-control change after replacement.

On non-POSIX filesystems, `posixPermissionsOf()` returns `null`, so Windows-style ACLs are not preserved at all.

Relevant code: [[ConfigFileReaderWriter.java](https://github.com/hivemq/hivemq-edge/blob/b16adfbd1d74e8b96cec1d6031b253673303f9fe/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L448-L483)](https://github.com/hivemq/hivemq-edge/blob/b16adfbd1d74e8b96cec1d6031b253673303f9fe/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L448-L483).

Suggested fix:

1. Capture the target’s `PosixFileAttributes`, including owner and group—not only its mode.
2. While the partial remains `0600`, apply the intended owner/group and any ACLs.
3. Apply the final mode last, then verify all relevant access-control attributes before moving.
4. Use `AclFileAttributeView` on ACL-based filesystems, or abort replacement when equivalent protection cannot be reproduced.

### High: permission-read failures still fail open

```java
try {
    return Files.getPosixFilePermissions(path);
} catch (UnsupportedOperationException | IOException | SecurityException e) {
    return null;
}
```

`null` means:

```java
Files.createFile(partial); // process umask
...
applyPermissions(partial, null); // no-op
```

An `IOException` or `SecurityException` on an existing target is not equivalent to “the file has no permissions.” The write can proceed using the umask and replace the protected file.

Only a genuinely absent target or positively unsupported attribute view should produce “nothing to preserve.” Permission-inspection failures should abort before creating or writing the partial file.

### Other reviewed portions

- Forwarder teardown: correct for the reported failure. Every teardown phase now runs, ownership is released after the forwarder is unpublished/stopped, failures are aggregated, and `BridgeService` continues with subsequent forwarders.
- In-memory sampler admission: correct. It calculates prospective victims without mutation, rejects while retaining the old sample, and evicts only after admission succeeds.
- Commercial sampler: not changed by this commit, but the current commercial branch contains a separate corresponding fix (`169bbcb1`).

All submitted focused test classes passed. The additional group-preservation test failed exactly as described above.

Verdict: request changes on the configuration-permission portion; the R3-04 and core R3-05 fixes can be accepted.

hivemq-edge/src/main/java/com/hivemq/bridge/MessageForwarderImpl.java
marregui: Rewritten — the scan is gone.

`isForwarderQueue` is now a single `containsKey` against a reference-counted index keyed by the exact queue id: O(1), no iteration over forwarders and no parsing of the queue name.

```java
public boolean isForwarderQueue(final @NotNull String queueId) {
    return forwarderQueueRefs.containsKey(queueId);
}
```

The index is maintained by `addForwarder`/`removeForwarder` and by `reserveQueues`/`releaseReservedQueues`, reference-counted so a restart hand-over never passes through zero. That change did more than fix the cost: because ownership is now looked up by the exact id, nothing on this path splits a queue name at all — which is the defect the ticket is about.

hivemq-edge/src/main/java/com/hivemq/sampling/SamplingService.java
marregui: Fair question. Short answer: the queue name is not this PR's to choose, and given the name the parse is decidable rather than heuristic.

**Why the name is fixed.** `$SAMPLER::<topic>/<topic>` is what 2026.11 already writes to disk. Re-encoding it with an unambiguous separator renames every persisted sampler queue, so on upgrade the samples stored under the old names become unreachable. That is the same trade rejected for the forwarder digest, for the same reason — it swaps a bug for an upgrade that loses data.

**Why the parse is not a guess.** Both halves after the prefix are the same string, so the separator can only be at the exact midpoint. Three checks, no searching and no backtracking:

- the id starts with `$SAMPLER::`
- the remainder has odd length with `/` at its midpoint
- the two halves are equal

Anything not of that shape is rejected outright. It is pinned by `SamplingServiceTest`: near-misses differing only in the last character, topics that themselves look like other queue ids, and an exhaustive differential test against the previous implementation.

**The alternative, and why it is not here.** The other option is what the forwarders do — keep a registry of live queue ids and never parse. It would work; `isSamplerQueue` and the `isOrphaned` call site could both be served by a lookup. I did not take it because it is a refactor of the orphan-reclaim path late in the cycle for no behavioural change, and that path is the one this ticket exists to protect. If you would rather have the symmetry, say so and I will do it — it is contained.

marregui:
Both fixed, on `e20cfb40d`. The branch has also been refreshed onto master, which had moved.

### XML normalization bypass

You were right that the fail-closed check was the wrong instrument — it compared a raw XML source span against post-JAXB output. The collector no longer works on the file's bytes: it parses the configuration and collects element text and attribute values from the parsed document, so the value it stores is the value JAXB holds and marshals. That closes all three cases you reproduced, plus single-quoted attributes:

- **numeric character reference** — the parser resolves it, so the collected value matches what is written
- **CRLF padding** — normalized by the parser
- **CDATA** — the walk sees the section's content; the old pattern could not match it at all, so nothing was collected and the restore never ran

Both immediate measures are in as well:

- `restorePlaceholders` refuses the write when any `${ENV:...}` token in the file was not accounted for by the walk — including when the document could not be parsed at all, which previously returned "no placeholders" and would have written every resolved secret out in silence
- a final sweep over the finished document refuses if a collected credential is still present, independent of which branch produced it

One deliberate deviation from your suggestion: an unrestorable credential is written as `${ENV:EDGE_UNRESTORED_SECRET}` rather than as an empty element. Several secret-bearing elements are `nonEmptyString` in the schema, so an empty one makes the file unparseable on the next start — a worse failure, and one that names the schema rather than the cause. The placeholder makes the next start stop with `Environment Variable EDGE_UNRESTORED_SECRET ... is not set`.

### Configuration file access control

Also right, and reproduced. The replacement now carries `PosixFileAttributes` — owner, group and any ACL — not just the mode. Owner and group are set first while the file is still `0600`, then the ACL, then the mode, then a verification read before the move; a mismatch aborts the replacement. And `preservedAttributesOf` no longer fails open: an `IOException` or `SecurityException` reading an existing file's protections aborts before the partial file is created. Only a genuinely absent file, or a store with no such view, is "nothing to preserve".

### Tests

`ConfigWritePermissionsTest` gained your group-preservation case (skipped where the build account has only one group) and the fail-closed read case. `EnvVarUtilTest` gained one test per bypass plus the token-accounting and survivor-sweep refusals. Mutation-checked: reverting each of the three new guards fails exactly its own test and nothing else.

On this SHA: 5131 core unit, 254 commercial, 1161 integration, spotless clean.

### Still open — on the record rather than found later

- the persistence-contract compatibility test is in-JVM (`ClientQueueModuleCompatibilityTest`, legacy / current / half-migrated implementations) rather than a boot against the released 2026.11 jar. It proves the contract semantics, not "an old jar left in `modules/`"
- R2-06, R2-19 and R2-22 from v02 I have not re-verified on this branch
- no tickets exist yet for the deferrals (lock-order inversion, username-without-password). EDG-885 stays open as agreed
- four integration tests are flaky and unrelated to this branch: `SimpleModbusIT`, `PublishMessageExpiryIntervalMemoryIT`, `OpcUaMetadataIT`, `MySQLLifecycleIT`. Each failed once across two full runs, passed on retry, and passes in isolation — different tests each run. They need their own ticket

### Where your v02 gates landed

The branch was rebased before this, so your line numbers no longer resolve. R2-02 `68a5b1a2d` · R2-01 `b540ff8b2` · R2-03 `ad5eb2b4f` · R2-04 `05827a125` · R2-14/15/16/17 `f979a12c0` `22a942f71` `da948ef2b` · R2-21 `3a4ffc6c3` · one-liners R2-09/10/13 `ce0939408` · today's two `de8f9544d`.

caoccao:
Two issues remain at PR head `e20cfb40d`. The earlier XML-normalization and POSIX group-preservation problems are fixed.

### Findings

1. **High — ACL-only filesystems still expose the temporary configuration file**

On Windows/NTFS, `permissions()` is `null`, so [`[createPartialFile](https://github.com/hivemq/hivemq-edge/blob/e20cfb40d5d3e172ab94e2e90703976d259e34b4/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L627-L634)`](https://github.com/hivemq/hivemq-edge/blob/e20cfb40d5d3e172ab94e2e90703976d259e34b4/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L627-L634) creates the file with inherited/default ACLs:

```java
if (preserved.nothingToReproduce() || preserved.permissions() == null) {
    Files.createFile(partial);
}
```

Secrets are then written before the target ACL is applied. Additionally:

- The owner is restored only through `PosixFileAttributeView`, so it is not restored on ACL-only filesystems.
- Empty ACLs are discarded as `null`, although an empty ACL can be meaningful.
- [`[verifyPreservedAttributes](https://github.com/hivemq/hivemq-edge/blob/e20cfb40d5d3e172ab94e2e90703976d259e34b4/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L701-L729)`](https://github.com/hivemq/hivemq-edge/blob/e20cfb40d5d3e172ab94e2e90703976d259e34b4/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L701-L729) returns immediately without a POSIX view, so neither ACL nor owner is verified.
- On stores supporting both POSIX permissions and ACLs, the final `chmod` may change the ACL mask, but the ACL is not subsequently verified.

**Suggested fix:** create the partial file with a restrictive ACL atomically—or the target ACL itself—before writing. Restore the owner through `FileOwnerAttributeView`, preserve empty ACLs, and verify ACL and owner after all attributes have been applied.

2. **Medium — Common credential values incorrectly prevent every configuration write**

The final safety check uses whole-document substring matching:

```java
if (document.contains(value)
        || document.contains(escapeXmlText(value))
        || document.contains(escapeXmlAttribute(value))) {
    throw new UnrecoverableException(false);
}
```

See [`[refuseIfACredentialSurvived](https://github.com/hivemq/hivemq-edge/blob/e20cfb40d5d3e172ab94e2e90703976d259e34b4/hivemq-edge/src/main/java/com/hivemq/util/render/EnvVarUtil.java#L524-L545)`](https://github.com/hivemq/hivemq-edge/blob/e20cfb40d5d3e172ab94e2e90703976d259e34b4/hivemq-edge/src/main/java/com/hivemq/util/render/EnvVarUtil.java#L524-L545).

A password of `admin`, for example, is successfully replaced with `${ENV:PW}`, but the resulting XML still contains the unrelated element name `<admin-api>`. Consequently, `document.contains("admin")` rejects the write. Short values such as `a`, `1`, or `true` are even more likely to fail.

**Suggested fix:** inspect parsed XML text and attribute values and compare complete values, rather than searching the serialized XML—including tag names—with `String.contains`.

### Validation

I ran 61 focused tests:

- The new numeric-character-reference, CRLF, and CDATA cases pass.
- All submitted POSIX permission/group tests pass.
- The added `PW=admin` regression test fails with `UnrecoverableException`, confirming finding 2.
- The ACL-only path could not be executed on the current POSIX filesystem, but the unsafe ordering and missing verification are explicit in the code path.

marregui:
Both findings are fixed on `4c374e91b`. The pass they prompted turned up eight more of the same class in the same code; those are fixed too, and listed below so none of it arrives as a surprise in the diff.

The branch was refreshed onto master by **merge, not rebase**, so your line anchors survive this time.

### Finding 1 — the ACL-only path

- The replacement is restricted to its own owner while it is still empty, and the narrowing is proved before any content goes into it.
- Owner is read and restored through `FileOwnerAttributeView`, so it exists on a store that has no POSIX view.
- An empty access-control list is carried rather than discarded as "nothing to reproduce".
- The verification covers mode, group, owner **and** the list, and runs after the mode is applied — the `chmod`-rewrites-the-mask case you named.

**Two deviations, both deliberate, both yours to ratify.**

The partial is not created with the target's own list. A list naming a principal this process is not either locks the writer out of the file it just created or — where the target is readable by others — leaves the replacement open to them for the whole write, which is the defect itself. It is created, narrowed to its owner, verified, then written. The window that remains holds an empty file.

The checks no longer compare lists for equality. A store that returns a reordered, merged or flag-decorated list has not widened anything, and demanding identity would have refused **every** configuration write on any store that normalises — on the only platform this path exists for, in a way neither of us can reproduce. They ask whether the list grants anyone more than the reference does, per principal, over both ALLOW and DENY (losing a DENY widens access even though no ALLOW changed). The predicate is exercised directly against every shape a store might plausibly hand back in `AclComparisonTest`, which needs no file system at all.

### Finding 2 — the survivor sweep

Asked of the parsed document's element text and attribute values now, never of its markup. Your `PW=admin` case passes, and so does `<port>1883</port>` with a password of `1`.

I went further than "compare complete values", and this one is worth your attention: whole-value equality still refuses a bridge whose password variable resolves to its own host name — reproduced end to end before changing it. The sweep now asks only of credential-bearing spans, where the question stays containment rather than equality so `<password>prefix-s3cr3t</password>` is still caught. The argument is that the restore replaces a span with its own literal and cannot move a value into a span it never touched, so the same string in `<username>` is the operator's own text. The broad question survives in the branch where the span itself has vanished, because there the marshaller could have put the value anywhere.

### Found in the same pass, and fixed

- The rolling backup carried the configuration's **mode but not its group** — R3-07 exactly, in the sibling file the same write creates. It now goes through the same replace-by-move as `config.xml`, so the two cannot drift apart again.
- Replace-by-move needs permission on the *directory*, not the file, so a write-protected `config.xml` was being **replaced anyway**. Writing in place refused that; so does this now.
- `setOwner` failing aborted the whole write, so a configuration installed by one account and served by another (root-owned file, service user) could never be persisted. Owner is now best-effort with a warning; mode, group and list still fail closed.
- A span on an element carrying attributes was never located: a credential in one refused every write, a non-secret in one lost its placeholder. Not reachable through today's schema — the arbitrary XML in the file is unmarshalled into maps and loses attributes first — so it was a trap rather than a live defect.
- The token accounting subtracted a parsed-document count from a raw-text one and could go **negative**, and the guard refuses on "not zero". Counted per variable now, and two different oddities can no longer cancel to a clean bill of health.
- A variable resolving to the empty string dropped the operator's placeholder silently. It is reported before the write that will drop it.
- A credential containing `&` or `<` cannot be supplied through a variable at all — the value is spliced in as raw text before the parse. That is unchanged, deliberately (it is what lets a fragment bring in markup), but the parse failure now names the variable responsible, and never its value.
- Every deliberate refusal was a message-less `UnrecoverableException` stack under a generic "sync failed" line. It now states the consequence: the node is correct, `config.xml` is stale, a restart would lose the change. The REST caller still is not told — making it so decides what every configuration endpoint answers when the file cannot be written, which is a larger change than this one.

### Verified

- Unit: 5138 tests green, including the new `AclComparisonTest` (13), `ConfigBackupPermissionsTest` (8), `ConfigWriteAclPermissionsTest` (18), `ConfigWritePermissionsTest` (14), `ConfigWriteFailureReportingTest` (5). Every fix is mutation-checked: reverting each guard fails exactly its own tests and nothing else.
- Integration: the config and bridge slice (`BridgeApiIT`, `ConfigReloadIT`, `ConfigSchemaIT`, `BridgeReloadIT`, `BridgeRestUpdateQueueRetentionIT`, `BridgeConfigReloadQueueRetentionIT`, `BasicBridgeIT`, `DataCombiningFunctionalityFromConfigFileIT`) — 36 tests, `--no-build-cache`. The full 375-class suite is a CI job, not a local gate; CI is the check on it.
- External, against the built distribution: 30 assertions over 8 scenarios on a real node driven through the REST API — 60 writes with no credential reaching disk, the backup's group, the password-equals-host case, the escaped-token case, the empty-variable and ampersand diagnostics, a write-protected file refused with the file untouched and no backup taken, and 160 concurrent writes followed by a restart from the configuration those writes produced.

### Still open

- **Windows.** The decision logic is now proven for every shape a store can return, so a normalising NTFS no longer refuses writes; what nobody in this loop can execute is what NTFS actually returns. Every workflow in this repo is `ubuntu-latest`. A capability-gated test class plus one Windows job would close it — that is a CI change with a cost, so it wants its own ticket rather than this PR.
- `isSecretElement` is a five-suffix list (`password`, `secret`, `passphrase`, `token`, `credentials`) and every decision on this path turns on it. Whether any credential-bearing element in the schema or in an adapter configuration falls outside it is a survey I have not done.
- A placeholder written with a character reference is never resolved, and a write converts it into a live one. Pre-existing, low, and now reachable rather than blocked by a refusal.
- Unchanged from the last round: the persistence-contract compatibility test is in-JVM rather than a boot against the released 2026.11 jar; R2-06, R2-19 and R2-22 from v02 are not re-verified on this branch; the deferrals (lock-order inversion, username-without-password) still have no tickets; EDG-885 stays open; and `SimpleModbusIT`, `PublishMessageExpiryIntervalMemoryIT`, `OpcUaMetadataIT` and `MySQLLifecycleIT` are flaky and unrelated.

gradle/libs.versions.toml
```diff
@@ -33,6 +33,7 @@ javassist = "3.32.0-GA"
 jaxb4-bind = "4.0.5"
 jaxb4-impl = "4.0.9"
 jctools = "4.0.6"
+jimfs = "1.3.1"
```
caoccao: The order is wrong. It should be below `jersey`.

gradle/libs.versions.toml
```diff
@@ -127,6 +128,7 @@ javassist = { module = "org.javassist:javassist", version.ref = "javassist" }
 jaxb4-bind = { module = "jakarta.xml.bind:jakarta.xml.bind-api", version.ref = "jaxb4-bind" }
 jaxb4-impl = { module = "org.glassfish.jaxb:jaxb-runtime", version.ref = "jaxb4-impl" }
 jctools = { module = "org.jctools:jctools-core", version.ref = "jctools" }
+jimfs = { module = "com.google.jimfs:jimfs", version.ref = "jimfs" }
```
caoccao: The order is wrong. It should be below `jersey`.

caoccao (changes requested):
Not completely. At PR head `4c374e91b`, the environment-variable issue is fixed, but the ACL fix still has one security problem.

### High — ACL comparison ignores access-significant ordering and flags

[`[grantsNoMoreThan](https://github.com/hivemq/hivemq-edge/blob/4c374e91b496acb77554db6193359168b9a1af34/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L830-L873)`](https://github.com/hivemq/hivemq-edge/blob/4c374e91b496acb77554db6193359168b9a1af34/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L830-L873) merges ACL entries into permission sets:

```java
return contains(byPrincipal(reference, ALLOW), byPrincipal(candidate, ALLOW))
        && contains(byPrincipal(candidate, DENY), byPrincipal(reference, DENY));
```

This discards:

- ACE ordering.
- `INHERIT_ONLY` and other flags.

However, Java defines an ACL as an ordered list, and Windows evaluates ACEs sequentially. Therefore these lists can grant different access:

```text
Target:      DENY Alice READ, ALLOW Alice READ   // denied
Replacement: ALLOW Alice READ, DENY Alice READ   // potentially granted
```

The comparator reports them as equivalent. Microsoft explicitly documents that DACL order affects access checks, and Java exposes `INHERIT_ONLY` as meaning that an entry does not apply to the object itself. [[Oracle ACL documentation](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/nio/file/attribute/AclFileAttributeView.html)](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/nio/file/attribute/AclFileAttributeView.html), [[Microsoft DACL ordering](https://learn.microsoft.com/en-us/windows/win32/secauthz/order-of-aces-in-a-dacl)](https://learn.microsoft.com/en-us/windows/win32/secauthz/order-of-aces-in-a-dacl), [Java `INHERIT_ONLY`](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/nio/file/attribute/AclEntryFlag.html#INHERIT_ONLY).

This comparator is used both:

- [[Before writing credentials](https://github.com/hivemq/hivemq-edge/blob/4c374e91b496acb77554db6193359168b9a1af34/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L787-L828)](https://github.com/hivemq/hivemq-edge/blob/4c374e91b496acb77554db6193359168b9a1af34/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L787-L828), when deciding whether the partial file is sufficiently restricted.
- [[During final ACL verification](https://github.com/hivemq/hivemq-edge/blob/4c374e91b496acb77554db6193359168b9a1af34/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L1024-L1042)](https://github.com/hivemq/hivemq-edge/blob/4c374e91b496acb77554db6193359168b9a1af34/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L1024-L1042).

A false “equivalent” result can therefore permit writing configuration credentials into a file with broader effective access.

I added two temporary adversarial tests:

1. Reordering an `ALLOW` ahead of its corresponding `DENY`.
2. Removing `INHERIT_ONLY` from an `ALLOW`.

Both failed because `grantsNoMoreThan` returned `true`.

**Suggested fix:** do not reduce ACLs to unordered maps. Either require exact ordered ACL reproduction—which is the safest fail-closed behavior—or implement platform-aware effective-access comparison that preserves ACE ordering and access-relevant flags. For the pre-write owner-only check, accept only ACLs that grant current-object access solely to the partial file’s owner.

### What is fixed

The credential substring false positive is fixed correctly. The survivor sweep now examines parsed XML values only in credential-bearing elements, so passwords such as `admin` no longer collide with `<admin-api>` or unrelated values.

I ran 127 submitted focused tests covering environment restoration, POSIX permissions, ACL-only permissions, ACL comparison, and backup protection; all passed. The two additional ACL adversarial tests exposed the remaining issue.

gradle/libs.versions.toml
marregui: Moved below `jersey`, in this block and the other one. `12f8c7b39`

gradle/libs.versions.toml
marregui: Moved below `jersey`, in this block and the other one. `12f8c7b39`

marregui:
Both fixed on `12f8c7b39`.

### Finding 1 — order and `INHERIT_ONLY`

You are right on both, and the second one is an error of mine rather than a gap. The javadoc argued that inheritance flags "govern what a *directory* propagates to what is created inside it" — true of `FILE_INHERIT` and `DIRECTORY_INHERIT`, false of `INHERIT_ONLY`, which says the entry does not apply to the object carrying it. Removing it hands its principal access to the file that the reference never gave.

`grantsNoMoreThan` no longer merges entries into per-principal permission sets. It asks what each list **decides**:

- the list is walked in order, and the first entry that names a principal and mentions a permission settles that permission for that principal; no later entry can change it
- settlement is per permission, not per principal, so `DENY Alice READ` followed by `ALLOW Alice {READ, WRITE}` leaves Alice with the write and not the read
- entries carrying `INHERIT_ONLY` decide nothing about this file and take no part, and neither do audit and alarm entries, which grant nothing

The two conservative directions are unchanged in shape and now asked of that result: an `ALLOW` in effect the reference does not have widens, and losing a `DENY` the reference has in effect widens.

Your two cases refuse now. The reorder pair is asymmetric, which is the point of doing it this way: `DENY, ALLOW` → `ALLOW, DENY` is refused, while `ALLOW, DENY` → `DENY, ALLOW` is accepted, because a denial that an earlier allow has already made unreachable is worth nothing, and moving it somewhere it decides something is a narrowing.

I did not take the "require exact ordered reproduction" option you offered first. It is fail-closed on the wrong axis: it refuses every configuration write on any store that normalises, on the only platform this path exists for, which is v04 finding 2.1 back again. The other half of your suggestion is in as stated — the partial file is accepted before the credentials go into it only when its effective access on the object itself is its owner's alone.

### Finding 2 — `libs.versions.toml`

Moved below `jersey`, in both blocks.

### Verified

`AclComparisonTest` 13 → 20. Four cases that must still pass — a denial moved ahead of its allow is a narrowing, per-permission settlement, an `INHERIT_ONLY` entry adds nothing, adding the flag narrows — and three refusals: your reorder, your flag removal, and flag removal on an otherwise owner-only list.

Mutation-checked, and each guard is exactly as wide as its tests:

- drop the `INHERIT_ONLY` skip → the three `INHERIT_ONLY` tests fail and nothing else
- drop the ordering → `movingAnAllowAheadOfItsDenialGrantsMore` fails and nothing else

309 tests across `com.hivemq.configuration.reader.*` and `com.hivemq.util.render.*`, no failures. Spotless clean from the composite root.

### Still open, unchanged from the last round

- **Windows.** Every workflow in this repo is `ubuntu-latest`, so what NTFS actually returns is still unexecuted. A capability-gated test class plus one Windows job would close it; that is a CI change with a cost and wants its own ticket.
- `isSecretElement` is a five-suffix list and every decision on this path turns on it. Whether any credential-bearing element falls outside it is a survey I have not done.
- A placeholder written with a character reference is never resolved, and a write converts it into a live one. Pre-existing and low.
- R2-06, R2-19 and R2-22 from v02 are not re-verified on this branch; the deferrals (lock-order inversion, username-without-password) still have no tickets; EDG-885 stays open; and `SimpleModbusIT`, `PublishMessageExpiryIntervalMemoryIT`, `OpcUaMetadataIT` and `MySQLLifecycleIT` are flaky and unrelated.

Back to you for another round.

marregui:
The two items I listed as open now have tickets, so neither is riding on this PR:

- **EDG-947** — run the configuration-write file-protection tests on Windows in CI. Capability-gated class plus one Windows job; what NTFS actually returns gets recorded either way. https://linear.app/hivemq/issue/EDG-947
- **EDG-948** — survey every credential-bearing element in the schema and in the adapter configurations against the five-suffix name rule, and say for each whether the rule recognises it. https://linear.app/hivemq/issue/EDG-948

caoccao:
Not completely. The new commit `12f8c7b39` fixes same-principal ordering and `INHERIT_ONLY`, but one ACL security issue remains.

### High — ACL evaluation ignores group membership

[`[effective()](https://github.com/hivemq/hivemq-edge/blob/12f8c7b3902f39ff298d1a2596ffe59d16a91b79/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L873-L897)`](https://github.com/hivemq/hivemq-edge/blob/12f8c7b3902f39ff298d1a2596ffe59d16a91b79/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L873-L897) evaluates order independently for each `UserPrincipal`:

```java
settled.computeIfAbsent(entry.principal(), ...);
```

Windows evaluates ACEs against an access token containing both the user SID and all applicable group SIDs. Therefore entries naming different principals can settle the same user’s permission. [[Microsoft’s access-check documentation](https://learn.microsoft.com/en-us/windows/win32/secauthz/how-dacls-control-access-to-an-object)](https://learn.microsoft.com/en-us/windows/win32/secauthz/how-dacls-control-access-to-an-object) describes this token-based evaluation.

For Alice, who belongs to `Administrators`:

```text
Target ACL:
DENY  Alice          READ
ALLOW Administrators READ

Replacement ACL:
ALLOW Administrators READ
DENY  Alice          READ
```

The target denies Alice. The replacement’s group `ALLOW` can grant the requested access before Alice’s later `DENY` is reached.

The new implementation nevertheless considers these ACLs equivalent because it creates separate map entries for `Alice` and `Administrators`; both maps ultimately contain the same per-principal `ALLOW` and `DENY` sets.

This affects both:

- [[The pre-write decision](https://github.com/hivemq/hivemq-edge/blob/12f8c7b3902f39ff298d1a2596ffe59d16a91b79/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L811-L827)](https://github.com/hivemq/hivemq-edge/blob/12f8c7b3902f39ff298d1a2596ffe59d16a91b79/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L811-L827), where credentials may be written after a false “no wider” result.
- [[Final ACL verification](https://github.com/hivemq/hivemq-edge/blob/12f8c7b3902f39ff298d1a2596ffe59d16a91b79/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L1059-L1077)](https://github.com/hivemq/hivemq-edge/blob/12f8c7b3902f39ff298d1a2596ffe59d16a91b79/hivemq-edge/src/main/java/com/hivemq/configuration/reader/ConfigFileReaderWriter.java#L1059-L1077).

I added this adversarial test temporarily; it failed because `grantsNoMoreThan(candidate, reference)` returned `true`.

**Suggested fix:** do not claim semantic equivalence using per-principal maps. The safest approach is exact ordered ACL reproduction. If representation tolerance is required, only accept transformations proven not to alter cross-principal ordering, or use a platform-native effective-access check that accounts for user and group SIDs. Ambiguous transformations must fail closed.

### Verification summary

- The previous same-principal order failure is fixed.
- The previous `INHERIT_ONLY` failure is fixed.
- The environment-variable `admin` collision remains fixed.
- All 60 submitted ACL, POSIX-permission, and backup tests passed.
- One new group-membership adversarial test failed.

Therefore, the ACL fix is still incomplete.

marregui:
Fixed on `8bc507eb8`, and you were right for the third round running.

Group membership was the half of the clause I had implemented the other half of: I had "the list is walked in order" and not "the walk matches against every identity the requester holds at once". Same sentence in both MS-DTYP §2.5.3.2 and RFC 7530 §6.2.1; I took one clause out of it and left the other.

### What changed

The comparison no longer summarises a list per principal at all. That structure *was* the defect — a `Map<UserPrincipal, …>` cannot represent "a group entry settles a user's permission", and `GroupPrincipal` being a subtype of `UserPrincipal` means no compiler was ever going to say so. Patching it a fourth time would have kept it.

`AclComparison` asks the question over **tokens** instead, and gets away with it because of one observation: a witness never needs more than two principals. If some token is granted a permission by the candidate and not by the reference, take `x` from the candidate's deciding entry and `y` from the reference's — every entry ahead of those failed to match the larger token, so it fails to match `{x, y}` too, and both deciding entries still match. Trying every pair of the principals the two lists name therefore **is** quantifying over every token, not an approximation of it. Your Alice/Administrators case is refused by the shape of the rule now, not by a clause that models it.

That also makes it exact rather than conservative, which matters in the other direction: it accepts precisely what a real access check accepts, so it cannot reintroduce v04's finding 2.1 by refusing writes on a store that normalises. I did not take the exact-ordered-reproduction option for that reason, as before.

### Done as extra

Fixing the instance would not have stopped a v07, so most of the work went on closing the class instead.

- **An oracle.** The access check itself, written straight from MS-DTYP §2.5.3.2, taking the requester's identities as an explicit token. `AclComparison` is asserted to agree with it **exactly** — to accept precisely what it accepts, not merely to be safe where it is safe — over exhaustively generated universes: every list up to two entries across three principals, both deciding types, all three permission masks and both states of `INHERIT_ONLY`; every list up to three entries at one bit; every list up to two entries across all four entry types; and 200,000 deterministically drawn pairs up to eight entries deep. About 13.6 million list pairs.
- **A four-identity pass**, with all sixteen subsets of the four presented as tokens. This is the one that actually tests the claim the design rests on — with three identities, "a witness needs at most two" is nearly vacuous, and an implementation that in fact needed three would be caught there and nowhere else.
- **Your v05 and v06 shapes are asserted to lie inside the enumerated universe**, so the search that has to find the next finding is shown to have been wide enough to have found the last two. That felt like the thing missing from the previous rounds: named regression tests prove a fix, they do not prove the net has holes small enough.
- **False refusals over the exhausted universe are asserted to be zero**, as a count rather than inferred, so a future change trading exactness for simplicity has to state what it cost.
- **Algebraic properties**, 100,000 pairs each: renaming every principal through a bijection changes no answer, swapping the permission bits changes no answer, and judging two lists whole is the same as judging each bit's slice separately. Plus transitivity — the write path judges the partial file against the target and then the written file against it again, and leans on transitivity without saying so anywhere.
- **Concurrency**, since this decides whether credentials may be written and is reached from whatever thread a REST request landed on: identical answers under sixteen threads behind a start gate, sixteen concurrent replacements of different files in one directory each landing their own content with no working file left behind, and a file replaced while sixteen readers read it never seen half written.
- **A capability-gated native-store class** that asserts what a store *decided* rather than what shape it returned. It self-skips everywhere we can run it and will pass or fail honestly the day EDG-947 gives us a Windows job.
- **Five guards mutation-checked.** Worth reporting because one of them failed to fail: the `AUDIT`/`ALARM` exclusion broke nothing when reverted, because the generated universe contained no audit entries. An audit entry read as a decision stands in front of the allow behind it and hides it — so a replacement that grants a principal access looks like one that grants none, and the write is **accepted**. A widening waved through, which is the direction that matters. The alphabet now carries all four entry types and there is a named test for both directions.

Verified on this head: 5178 core unit tests green, four skipped (three of them the native-store class), spotless clean from the composite root.

### And then the rest of the class

Since three of your rounds landed on the same file, I audited `ConfigFileReaderWriter` end to end rather than only along the lines this branch changed. Five defects, none of them introduced by this work and none of them the bridge message loss — so they are filed as EDG-949 rather than grown onto this branch. Two are worth knowing about: with configuration hot reload enabled, the watcher dies silently and permanently the first time the file cannot be parsed, so a corrected typo is never picked up; and the rolling backup selects its victim loosely enough to overwrite an unrelated file sitting in the configuration directory. Both predate everything here.

### I would like to stop here

The reported defect was declared fixed at v03 — your words, and still true at this head. Everything since has been on the configuration write-back, which is scope this branch absorbed defensively rather than the ticket's bug, and it has now taken three rounds of its own. The ACL comparison is the place that kept generating them, and it is no longer argued: it is proven against an oracle over an enumerated universe, and the next clause of the specification would be caught here rather than by you. That seems like the right moment to stop growing #1714.

Everything found and not fixed here is on a ticket, so nothing is riding on the diff:

- **EDG-937** — the six remaining v03 findings, including R3-02 and R3-03, which need Martin's call on scope before they need code.
- **EDG-947** — run the configuration-write file-protection tests on Windows in CI. Until that exists, what NTFS actually returns is unexecuted; the comparison no longer depends on the answer, which was the point.
- **EDG-948** — survey which configuration elements hold credentials against the five-suffix name rule.
- **EDG-949** — the five defects from the full-class audit above.
- **EDG-885** stays open when this merges: nothing in production releases a sampled topic, so the release path needs an API change.

### Ready for another round

#1714 is up for another look. #431 and #357 still have no submitted review, and #357 cannot merge on its own — the persistence contract spans core and the file-native module, so the three have to go together.

Thank you for these three rounds. The ordering one I had argued myself out of in writing, and this one I had no way to catch; both of those are on me, and the oracle exists because of them.

caoccao (approved):
LGTM

PR #1715, opened 2026-08-17
EDG-899: Emit the @JsonValue form for enums in adapter config schemas

caoccao:
**Motivation**

Resolves [EDG-899](https://linear.app/hivemq/issue/EDG-899/adapter-config-schema-advertises-enum-constant-names-instead-of-the)

`CustomConfigSchemaGenerator` advertised enums by their Java constant names while Jackson reads and writes the `@JsonValue` form. Where the two differ, every value the schema offers is rejected on deserialization, and a declared default can be a value the same schema calls invalid.

`OpcuaConditionType` (EDG-835, 8c25c19b2) is the first adapter config enum to declare `@JsonValue`, so it is the first to hit this. `GET /api/v1/management/protocol-adapters/tag-schemas/opcua` returned:

```
definition.type.enum    = [CONDITION, ACKNOWLEDGEABLE_CONDITION, ALARM_CONDITION, ...]
definition.type.default = "AlarmConditionType"     # not a member of its own enum
```

That left no way to create an OPC UA tag from the UI:

- **Keep the pre-filled default** — AJV rejects it (`'Node type' must be equal to one of the allowed values`) and the tag drawer will not save.
- **Pick any offered value** — `POST .../adapters/{id}/tags` returns 200, then `ProtocolAdapterManager.refresh` throws and the tag is dropped. No error toast, no tag, no monitored item:

```
java.lang.IllegalArgumentException: Cannot construct instance of OpcuaConditionType,
problem: Unknown OPC UA condition type 'ALARM_CONDITION'.
Known types: ConditionType, AcknowledgeableConditionType, AlarmConditionType, ...
(through reference chain: OpcuaTag["definition"]->OpcuaTagDefinition["type"])
    at ProtocolAdapterFactory.convertTagDefinitionObjects(ProtocolAdapterFactory.java:81)
```

Found by the HiveMQ Edge smoke test suite against `2026.13-SNAPSHOT`. Master only — not in any released version.

**Changes**

- Add `JacksonOption.FLATTENED_ENUMS_FROM_JSONVALUE` to the `JacksonSchemaModule` in `CustomConfigSchemaGenerator`, so the schema advertises the form Jackson actually serializes.
- Two regression tests in `CustomConfigSchemaGeneratorTest`.

Enums without a `@JsonValue` accessor keep their constant names, so `OpcuaTagKind` and every other adapter enum are unaffected — `OpcuaConditionType` is currently the only `@JsonValue` enum in the adapter configs. The regenerated schema is self-consistent:

```
type.enum    = [ConditionType, AcknowledgeableConditionType, AlarmConditionType, ...]
type.default = AlarmConditionType     # in enum: true
kind.enum    = [VALUE, CONDITION, REFRESH, EVENT_SUBSCRIPTION]   # unchanged
```

The `defaultValue = "AlarmConditionType"` on `OpcuaTagDefinition.type` was always the correct wire value, so it is left as is — the enum was the broken side.

**Verification**

- `CustomConfigSchemaGeneratorTest` 6/6. The new tests cover both directions, since the risk here is to enums that were already fine: one asserts the wire form is emitted *and* that a declared default is a member of the emitted enum; the other asserts enums without `@JsonValue` still emit constant names.
- Rebuilt the full distribution and re-ran the smoke suite end to end. The tag saves, Edge logs `Added monitored items: [ns=1;s=Temperature]`, the northbound mapping publishes to MQTT, and adapter deletion cleans up — green over consecutive runs.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/api/json/CustomConfigSchemaGenerator.java
+++ b/hivemq-edge/src/main/java/com/hivemq/api/json/CustomConfigSchemaGenerator.java
@@ -63,7 +63,14 @@ public class CustomConfigSchemaGenerator {
                 .with(new JacksonSchemaModule(
                         JacksonOption.RESPECT_JSONPROPERTY_REQUIRED,
                         JacksonOption.INCLUDE_ONLY_JSONPROPERTY_ANNOTATED_METHODS,
-                        JacksonOption.RESPECT_JSONPROPERTY_ORDER))
+                        JacksonOption.RESPECT_JSONPROPERTY_ORDER,
+                        // Without this, an enum is advertised by its Java constant names while Jackson
+                        // reads and writes the @JsonValue form, so every value the schema offers is
+                        // rejected on deserialization. OpcuaConditionType is the case in point: the
+                        // schema listed ALARM_CONDITION, the API only accepts AlarmConditionType, and
+                        // a tag saved from the UI was dropped when the adapter reloaded.
+                        // Enums without a @JsonValue accessor keep using their constant names.
+                        JacksonOption.FLATTENED_ENUMS_FROM_JSONVALUE))
                 .with(new ModuleConfigSchemaGeneratorModule());
         withEnumDisplayNameProvider(configBuilder);
         withMutuallyExclusiveFieldsMarker(configBuilder);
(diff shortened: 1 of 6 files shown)


marregui (approved):
lgtm

PR #1716, opened 2026-08-17
EDG-835: Accept an explicitly null comment in a condition command

caoccao:
Closes QA finding **P4.3** on [EDG-835](https://linear.app/hivemq/issue/EDG-835/implement-opc-ua-alarms-and-conditions-support-in-edge), raised by @estefaniaamundaray after #1707 merged.

## The defect

OPC 10000-9 §5.7.3 gives a condition method's `Comment` argument three meanings, and null is one of them:

| Payload | Intent |
|---|---|
| `{"method": "ACKNOWLEDGE", "eventId": "…"}` | leave the existing comment unchanged |
| `… "comment": ""` | **erase** it |
| `… "comment": "Erwin has seen this"` | set it |

`ConditionUpdate.fromJson` has always implemented all three — an explicit `"comment": null` is read as absent, with the reasoning in the code — and the user documentation promises it. But `ConditionSchemas.writeSchema()` declared the field a bare `string`, and JSON `string` does not admit `null`. So the payload was refused at the southbound schema gate before the parser that understood it was reached:

```
$.value.comment: null found, string expected
```

Nothing unusual is needed to hit it. Jackson, `JSON.stringify` and `json.dumps` all emit `"field": null` for a field that was never assigned — omitting the key is what takes configuration — so a service that builds an acknowledgement as a struct and serialises it had every command rejected over a field it deliberately said nothing about. And `comment` is the operator's audit note: §5.5.2 fires a fresh event on any comment change, so the null-vs-empty distinction is the difference between "don't touch the note" and telling every other client the note is gone.

## The fix

`comment` is now nullable — `type: ["string", "null"]`. The semantics behind it were already correct.

## What deliberately stays strict

`eventId`, `duration` and `selectedResponse` are **not** nullable. They are OPC UA method arguments, required by the specification wherever they apply — `Acknowledge`'s `EventId` (§5.7.2), `TimedShelve`'s `ShelvingTime` (§5.8.10.4), `Respond`'s `SelectedResponse` (§5.6.3) — and there is no null value of any of them that means anything. A caller with nothing to say about such a field omits the key.

The consequence is real and is stated in the javadoc rather than left to be discovered: **a client that serialises the whole command object must suppress its null keys** (`@JsonInclude(NON_NULL)`, a `JSON.stringify` replacer, `model_dump(exclude_none=True)`). Sending `"duration": null` on an `ACKNOWLEDGE` is refused, and that refusal is the intended contract.

I first relaxed all four and was corrected: the serialiser argument is not a reason to advertise a value the protocol cannot carry. The javadoc now records both halves so the asymmetry reads as a decision rather than an inconsistency.

## Tests

`ConditionSchemaNullableFieldsTest` — 10 tests. It drives **the gate itself**: the same networknt validator Edge puts in front of a southbound write, against the composed document a client fetches from `GET .../writing-schema?direction=SOUTHBOUND`. That is deliberate — the defect was invisible to the tests already in this package, because the schema was right on its own and the parser was right on its own, and what was wrong was only visible between them. Asserting acceptance rather than a rendered type also means the error paths are the `$.value.comment` an operator sees.

It pins the strictness of the three argument fields as well as the fix, driven off a list so a new argument field has to declare which side it is on, and carries the serialiser case both ways: the struct with nulls intact is refused on those fields, the same command with nulls suppressed is accepted.

`libs.json.schema.validator` is added as a **test-only** dependency. hivemq-edge has it as `implementation`, so it reached this module's test runtime but not its compile classpath.

## Verification

- `./gradlew :hivemq-edge-build:hivemq-edge-module-opcua:check --rerun-tasks` — **BUILD SUCCESSFUL**, 681 tests, 0 failures, 1 pre-existing skip.
- `spotlessApply` from the composite root.
- **Negative check:** with the one `.nullable()` removed, 4 tests fail with `$.value.comment: null found, string expected` — the finding's assertion verbatim. The 3 tests guarding the argument fields pass either way, correctly, since they are guard-rails rather than regression proofs.
- No OpenAPI impact — tag schemas are generated at runtime, not baked into `ext/hivemq-edge-openapi.yaml`. No integration test pins these type strings.

## One thing not fixed here

`PropertyItem.tsx:52` looks its type icon and label up by a single type name, so a `type: ["string","null"]` misses and falls back to a generic icon. `comment` is now one such field on the southbound side — but the condition **read** schema has marked every field nullable since #1707, through the same `JsonSchemaBrowser` path, so this is pre-existing and cosmetic rather than introduced here. Same shape as the `ConnectionStatusBadge` finding from code review v09; worth its own ticket.

### DIFF
--- a/modules/hivemq-edge-module-opcua/build.gradle.kts
+++ b/modules/hivemq-edge-module-opcua/build.gradle.kts
@@ -63,6 +63,11 @@ dependencies {
     // hundred-megabyte JUnit XML that GitHub's result parser refuses. Nothing in main uses Logback directly;
     // the adapter logs through SLF4J and the binding is the runtime's business.
     testImplementation(libs.logback.classic)
+    // Test-only: the same validator Edge puts in front of a southbound write, so a published schema can be
+    // checked by running a payload through the gate rather than by reading the rendered type. hivemq-edge has
+    // it as an `implementation` dependency, so it reaches this module's test runtime but not its compile
+    // classpath. See ConditionSchemaNullableFieldsTest for the finding that needed it.
+    testImplementation(libs.json.schema.validator)
 }
 
 configurations {
--- a/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/condition/ConditionSchemas.java
+++ b/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/condition/ConditionSchemas.java
@@ -344,6 +344,39 @@ public final class ConditionSchemas {
      * condition as a whole. {@link ConditionUpdate.Method.Arguments} is the list. A static schema cannot
      * express that dependency, so the fields are optional here and the adapter checks them per method before
      * making the call.
+     * <p>
+     * <b>Optional and nullable are different questions, and only {@code comment} answers yes to the second.</b>
+     * Optionality in JSON Schema is about whether a key must be present; nullability is about whether
+     * {@code null} is a value the field can carry. A field can be omitted freely and still not admit a null,
+     * and four of the five are exactly that.
+     * <p>
+     * {@code comment} admits null because the specification gives null a meaning there. OPC 10000-9 §5.7.3
+     * makes the {@code Comment} argument a {@code LocalizedText} that may be NULL and says what NULL does —
+     * leave the condition's existing comment untouched, as against {@code ""} to erase it and text to set it.
+     * {@link ConditionUpdate#fromJson} has always implemented all three, and the user documentation promises
+     * them; declaring the field a bare string was the one thing making the promise unreachable, so the command
+     * an operator sends most often was refused over a field they deliberately said nothing about. Found by QA
+     * as EDG-835 P4.3.
+     * <p>
+     * {@code eventId}, {@code duration} and {@code selectedResponse} are <b>deliberately not nullable</b>, and
+     * the reason is that they are OPC UA method arguments rather than free-form fields of Edge's own: where each
+     * applies the specification makes it <em>required</em> — {@code Acknowledge}'s {@code EventId} (§5.7.2),
+     * {@code TimedShelve}'s {@code ShelvingTime} (§5.8.10.4), {@code Respond}'s {@code SelectedResponse}
+     * (§5.6.3). There is no such thing as a null {@code EventId} that means something; a caller who does not
+     * have one has nothing to say about the field, and the way to say nothing is to omit the key. Declaring them
+     * nullable would advertise a value the protocol cannot carry and no method accepts.
+     * <p>
+     * The consequence is worth stating rather than leaving to be met: a client that builds this command as a
+     * struct and serialises it must <b>suppress its null keys</b>, because Jackson,
+     * {@code JSON.stringify} and {@code json.dumps} all emit {@code "field": null} for a field that was never
+     * assigned. {@code @JsonInclude(NON_NULL)}, a {@code JSON.stringify} replacer, or
+     * {@code model_dump(exclude_none=True)} is what that takes. Sending {@code "duration": null} on an
+     * {@code ACKNOWLEDGE} is refused at the gate, and that refusal is the intended contract — the field is not
+     * a place to put a null.
+     * <p>
+     * {@code method} is not nullable either, for a third reason again: it is the field that decides which of the
+     * others apply, so a command that does not name an action is not an under-specified command but a different
+     * kind of thing. It is rejected as early as it can be, which is here.
      */
     public static @NotNull Schema writeSchema() {
         return new SchemaBuilder()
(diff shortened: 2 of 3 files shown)


marregui (approved):
lgtm

mschoenert-hivemq:
I am not strictly opposed to this, but I do not see the point.

All three options the specification defines were always obtainable:

1. `{"method":"ACKNOWLEDGE", "eventId":"123"}` — leaves the comment field alone
2. `{"method":"ACKNOWLEDGE", "eventId":"123", "comment":""}` — clears the comment field
3. `{"method":"ACKNOWLEDGE", "eventId":"123", "comment":"erwin has seen it"}` — sets the comment field

Your change adds a *second* way to say the first one:

1b. `{"method":"ACKNOWLEDGE", "eventId":"123", "comment":null}` — also leaves the comment field alone

That may well be preferable for clients that do not automatically omit empty fields, and I accept the serialiser argument. **But if we want to be permissive in this regard, should we not then make all fields nullable?**

Which is really a question about a general rule rather than about this PR: **for southbound payloads, should any field that is optional also be nullable — with absent and explicit null always treated as equal?**

I lean towards yes, because a rule is easier to hold than a per-field judgement, and because the asymmetry is otherwise hard to explain to a client author. The current position — `comment` accepts null, `eventId`, `duration` and `selectedResponse` do not — reads as an inconsistency unless you already know that the first is an OPC-UA `LocalizedText` with three meanings and the others are plain required arguments.

So the narrow question for this PR is whether to extend the same treatment to the other three. The wider one, which I would rather settle once: do we adopt "optional implies nullable" as the rule for everything we accept southbound, so this never has to be decided field by field again?

At this point I would keep this PR out of the release. But I leave the final judgement to you.

caoccao:
Good questions. I hold the same concern and discussed with QA offline.
In JavaScript, there are `undefined` and `null` that can differentiate **does not exist** and nullable.
In JSON, my understanding is omitting the property means **does not exist**, and `null` means nullable.

| Code | Existence | Feature |
|--|--|--|
| `{}` | does not exist | optional |
| `{ comment: null }` | must exist | not optional, nullable |
| `{ comment: "" }` | must exist | not optional, not nullable |

I think omitting the property is the right approach.

mschoenert-hivemq:
## On empty-string vs explicit-null vs not-present

Following up on my earlier comment, because the general question matters more than this PR.

JSON gives us three spellings where most formats give two: `"comment": ""`, `"comment": null`, and no `comment` key at all. Three items admit five groupings, so the question is which one we adopt as a general rule.

**Two of the five are nonsense and can be dismissed quickly.** Grouping `{empty, null}` against `{absent}` pairs a value with the absence of one. Grouping `{empty, absent}` against `{null}` is worse: it makes an explicit null the only way to say "I have a value", which inverts what null means in every language that has it.

That leaves three:

1. **All three equivalent** — one meaning, "nothing useful here".
2. **`{empty}` vs `{null, absent}`** — a value is present, or it is not.
3. **All three distinct** — three separate meanings.

**I want option 2, and I want to argue it from what we need to express rather than from what our downstream formats can carry.** At the highest level there are exactly two things worth distinguishing about a field: *a value is present*, and *no value is present*. Option 1 is too coarse for that, option 3 too fine.

**Why option 1 is too coarse.** "I care about this value, and the value is the empty string" is a different statement from "I do not know or do not care about this value". Collapsing them means we can never express **clearing** a field, because clearing and not-mentioning become the same request. That is not hypothetical — it is exactly the OPC-UA condition comment, where the specification gives an empty text and an absent text opposite meanings. It also generalises badly to other types: zero is a legitimate setpoint, and an empty array is a real state distinct from "not specified".

**Why option 3 is too fine.** I cannot think of anything I would want to express by the difference between an explicit null and a missing key. And the distinction survives almost nothing: deserialise a payload into an ordinary object in any language and the key's presence is gone — a variable holds a value or it holds nothing. Protobuf collapses the two as well. Java can technically model it, with a nested `Optional` or Jackson's `JsonNullable`, but nothing idiomatic does, and `@JsonInclude` settles absent-versus-null at serialisation time rather than in the type.

So the burden sits with the other side: **if anyone can name a field where explicit-null and not-present should carry different meanings, that is the case to examine.** Until one appears, two.

**As a rule, then:** for a southbound payload, every optional field is also nullable, and absent and explicit null are always treated as equivalent. The value of a rule is that it holds everywhere. Deciding it field by field is how we arrived at a condition command where `comment` accepts a null and `eventId`, `duration` and `selectedResponse` refuse one — an asymmetry no client author can infer from the outside.

caoccao:
In general I agree with your decision.

In practice, when `comment` is null in Java, JSON serialization can either omit `comment` or generate `comment: null`. That depends on `@JsonInclude(Include.NON_NULL)` or the following.

```java
ObjectMapper mapper = new ObjectMapper();
mapper.setDefaultPropertyInclusion(
  JsonInclude.Value.construct(
    JsonInclude.Include.NON_NULL,
    JsonInclude.Include.NON_NULL
  )
);
```

I don't want us to have a design that depends on some flags in a particular JSON library.

I prefer `{ comment: null }` is an invalid input. That means if comment is provided, it must not be null.

mschoenert-hivemq:
OK. I could also live with forbidding `{ ... "method":null ... }`. 

That means we would still generally support expressing two semantically different things:
1.  I know or care what the value is (and the value may be the empty string)
2. I don't know or don't care for the value
And the second case would be expressed by omitting the value.

In which case we could reserve explicit null for a few special cases.
E.g. when connecting to SQL databases.

PR #1719, opened 2026-08-18
EDG-907: Refuse an unreadable adapter configuration at the API, not on the reload

marregui:
EDG-907

## The defect

`POST` / `PUT` / `createCompleteAdapter` answer the caller before Edge has read the
configuration they submitted. `ProtocolAdapterExtractor` writes the entity and notifies
`ProtocolAdapterManager.refresh`, which converts it on a dedicated single-thread executor —
after `Response.ok()` has already gone out. A configuration that does not convert therefore
gets a 200 and then never appears: every later `GET /adapters/{id}` is a 404, permanently,
and the reason lives only in Edge's log and in the adapter-scoped `CRITICAL` event
`convertConfigs` fires. Anything polling for readiness burns its full timeout and reports
that timeout, which names nothing about the configuration.

Schema validation does not catch it: `CustomConfigSchemaGenerator` emits
`additionalProperties: false` only on the `oneOf` branches it synthesises for
mutually-exclusive field groups, never on the top-level configuration object. `PUT` had no
validation of the submitted configuration at all.

## The fix

Convert the entity through `ProtocolAdapterConfigConverter.fromEntity` — the same converter
and the same mapper the reload uses — before handing it to the extractor, and return 400
when it throws. The error carries the Jackson reference chain as the `parameter` and the
innermost message as the `detail`, so the caller gets the path through their own payload
plus the sentence the adapter's own configuration class wrote. Same split
`ProtocolAdapterSchemaManager.convertMessage` already uses.

Wired into all three endpoints that accept an operator-supplied config.

## Why not `additionalProperties: false`

It would make the API stricter than the configuration file.
`ProtocolAdapterUtils.ReportUnknownPropertyHandler` is the application-wide posture — warn
and carry on without the setting — so that a configuration written for another version
still starts. Only `OpcUaSpecificAdapterConfig` opts out and refuses. A blanket schema
restriction would reject payloads Edge loads happily from a file, and would state each
adapter's posture in a second place where the two could drift. Converting borrows the
posture rather than duplicating it, and covers every conversion failure — rejected enum
value, uncoercible value, unreadable tag definition — not only an undeclared setting.

## Verification

- 7 new tests in `ProtocolAdaptersResourceImplTest`: all three paths refuse and write
  nothing; all three still accept a good config; nested reference-chain reporting; the
  no-path fallback. Uses a purpose-built refusing config class — core cannot reference the
  OPC UA module.
- Full `:hivemq-edge:test` suite green.
- `spotlessApply` from the composite root; `spotlessJavaCheck` verified on `hivemq-edge`
  and `hivemq-edge-module-opcua`.
- Companion PR in `hivemq-edge-test` proves it end to end against a real Edge over real
  HTTP, and cleans a fixture carrying two settings OPC UA has never had.

GET → PUT round trips are unaffected: `unconvertConfigObject` runs with
`AUTO_DETECT_GETTERS` off, so only `@JsonProperty` members serialise, and those are by
definition settings the class declares.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImpl.java
@@ -27,6 +27,7 @@ import static com.hivemq.protocols.ProtocolAdapterUtils.createProtocolAdapterMap
 import static com.hivemq.util.ErrorResponseUtil.errorResponse;
 
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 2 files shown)


caoccao (approved):
LGTM

PR #1720, opened 2026-08-19
EDG-894: QA findings P6, P7 and P8 from the OPC UA Alarms & Conditions pass

caoccao:
Three QA findings from the [EDG-889](https://linear.app/hivemq/issue/EDG-889) test pass against [EDG-835](https://linear.app/hivemq/issue/EDG-835/implement-opc-ua-alarms-and-conditions-support-in-edge), raised by @estefaniaamundaray on [EDG-894](https://linear.app/hivemq/issue/EDG-894).

**Split out of #1716.** These three previously sat stacked on top of the nullable-`comment` fix. That PR now carries only its own commit, so each fix can be reviewed and land on its own. The commits here are byte-identical to what QA tested as `8a75a6538`, `7197513ae` and `917525e1d`; only the parent chain changed.

**A fourth finding (P4a) was on that stack and is not here.** Advertising per-method required arguments in the southbound schema needed a new member on the adapter SDK's public `ObjectSchema`, and that surface is versioned separately and consumed by third-party adapter authors — not a change to make under release pressure for a Low-severity discoverability gap. hivemq/hivemq-edge-adapter-sdk#142 is closed and [EDG-906](https://linear.app/hivemq/issue/EDG-906) stays open. The behaviour was never wrong: a command missing a method-specific argument is refused, naming the method and the field.

---

## P8 — a blank tag enum drops the tag instead of refusing it

`OpcuaTagDefinition` normalised its three narrowing node-id fields with `blankToNull` and its three enum fields not at all. So `"filterType": "   "` matched no constant, the tag failed to deserialise, and it was **dropped rather than refused** — the write was accepted, nothing was reported, and the tag was simply absent on read-back, because the throw fails conversion of the whole adapter configuration and that leaves the running adapter untouched.

A blank string is what a UI form submits when a box is cleared and what a config generator emits for an unset optional, so this is the ordinary path rather than an odd one.

**The fix is not where the finding puts it.** It proposes `blankToNull(filterType)` beside the other three, which cannot compile — the field is typed as the enum, not as a `String` — and would not run in time if it could, since Jackson constructs the enum before the constructor is entered. Blank is therefore decided in the creators.

A probe found the same defect on two fields the finding does not name:

| Field | `"   "` / `""` before |
|---|---|
| `filterType` | threw (reported) |
| `type` | threw — same creator; absent already meant `AlarmConditionType` |
| `kind` | threw via Jackson's own coercion with no creator at all: `Cannot coerce empty String to OpcuaTagKind value` — a message about Jackson's configuration rather than the operator's mistake |

`kind` is the worst of the three, because it decides what the tag is.

Blank now yields `null` in `OpcuaConditionType.fromConfig` and in a new `OpcuaTagKind.fromConfig`, so each field falls back to the default its own documentation already promises. This is the doctrine `EnumParsing` states one package over for the security config enums: blank or absent means unset, an unrecognised non-blank value is a configuration error and is rejected. **The rejection is kept deliberately** — unset resolves to a usable default, so quietly reading `AlarmConditonType` as unset would publish a different shape than was written, and a mistyped `kind` would subscribe to an alarm node's `Value` attribute and publish nothing.

Both creators trim, because Jackson's default enum handling trimmed before a creator took the decision over; without it, adding one to `kind` would have turned `" CONDITION "` from working into a dropped tag.

Checked rather than assumed: adding `@JsonCreator` to `OpcuaTagKind` does not change how the generated tag config schema renders it — still all four constants with default `VALUE`, and `type` still lists 22. EDG-899's generator keys on `@JsonValue`, which is not added here.

## P7 — `getSchema` could not say why a schema failed

The endpoint answered 500 whose entire body was `Json schema creation for tag failed.` for an ordinary `VALUE` tag. That sentence is the fixed message `TagSchemaCreationOutputImpl.fail(String)` completes its future with; the adapter's own account is recorded separately, on the output, and this layer read only the cause and discarded it. **Every** schema failure in the OPC UA adapter takes that route, so the endpoint could not explain any of them.

The cost was not only cosmetic. QA could not tell whether the fault was in schema generation or in the `direction` parameter that had just arrived with EDG-845, and reported it as a possible compatibility break for every pre-split API client. The adapter had in fact said `Discovery failed: ClientConnection not connected or not initialized`, and nothing carried that to the caller.

The reason is now preferred with the cause kept as fallback: `fail(Throwable, null)` leaves no message on the output, and there the throwable is the only account there is.

**On the rest of P7, which is not changed here.** The finding offers "if NORTHBOUND succeeds, the defect is the default direction". It cannot: an absent direction resolves to `NORTHBOUND`, the same enum value through the same call, and `createTagSchema` runs and fails before the direction is consulted at all. A test pins that the two answers are identical on both the success and the failure path, so **this is not an EDG-845 compatibility break**. What remains is that `conn.client()` is empty until the connection context is set at the end of `start()` while monitored items are created earlier — the known "nothing correct to wait on" gap rather than a new defect.

The status code is left alone. 500 is what this resource does for "the adapter is not usable yet" — `discoverDataPoints` does the same with an even less informative message — so answering 503 with `Retry-After` is an API contract decision rather than a bug fix.

## P6 — a narrowing node under the root notifier was never looked at

`f480b2f94` exempted the Server object from the propagation walk, and that exemption is right: OPC 10000-5 §8.3.2 makes every event of the server accessible at the Server object, which renders `OutsideHierarchy` unreachable there by definition.

But the walk was answering **two** questions. Alongside "does this node's event path reach the selected notifier" it incidentally answered "is this node something the server has", because a node the server does not hold produces a browse failure rather than a clean negative. Short-circuiting dropped both, and §8.3.2 licenses dropping only the first.

So a query tag rooted at the Server object — the natural choice for a plant-wide subscription — narrowed by a typo or a stale id from a migrated configuration was accepted without anything being looked at. It subscribes cleanly, its where clause matches nothing, and it stays silent for the life of the adapter. That is indistinguishable from a quiet plant, which is the outcome everyone hopes for, so nobody investigates.

`presentInAddressSpace` makes the same inverse `HasEventSource` browse the walk's first hop makes, so a server that permits one permits the other. What is read from it differs: not who the parents are, which under the root notifier is settled, but whether the server answered about this node at all.

**The answer is `Unverified`, never `OutsideHierarchy`.** OPC 10000-9 §4.3 permits a server to keep condition instances out of the address space entirely and deliver them through events alone, with §5.7.3 requiring the ConditionId be accepted regardless — so a missing node may be a working configuration. `Unverified` keeps the tag subscribed, which is the whole of review-09 finding 4 and must not be undone, while routing through `reportUnverifiedTag` so the operator is told which tag, which field and which node id.

**Not fixed here, deliberately:** the check is existence only. A narrowing node that exists and is a plain `Variable` — which no event can ever match — still subscribes silently, because `NodeClass` is never read. The caution above is sound for absence and does not extend to a node the server has positively described as a Variable, but that is a separate change, filed as [EDG-908](https://linear.app/hivemq/issue/EDG-908).

---

## Verification

Run on this branch's tree, rebased onto `2de6b49eb`, since these three had only ever been tested stacked on the now-discarded P4a work:

- `./gradlew :hivemq-edge-build:hivemq-edge-module-opcua:check` — **BUILD SUCCESSFUL**
- `./gradlew :hivemq-edge-build:hivemq-edge:test --tests '*ProtocolAdaptersResourceImplTest*'` — **BUILD SUCCESSFUL**
- `./gradlew :hivemq-edge-build:hivemq-edge:spotlessCheck` — clean
- `OpcuaTagDefinitionTest` 15, `EventPropagationVerifierTest` 9, `ProtocolAdaptersResourceImplTest` 34 — **58 tests, 0 failures**
- `hivemq-edge:compileTestJava` compiles without the adapter-SDK `requiredWhen` API, confirming nothing here depends on the discarded SDK change

## For QA

[EDG-908](https://linear.app/hivemq/issue/EDG-908) records its suite state (48 green / 3 red / 13 skipped) against `917525e1d` + adapter-SDK `a5bcd8a`, a stack that no longer exists. The red **count** should hold but not its composition — anything that was green because of P4a will now fail. Worth a re-run against this head and #1716's rather than trusting those numbers.

---

## Also here: a CI build-cache workaround (`a76baf28c`)

Unrelated to the three findings, carried on this branch so the Jenkins `edge-composite` job can go green. `cyclonedxDirectBom` is declared non-cacheable so it re-executes rather than being restored from a cache entry.

It sits in `VersionUpdaterPlugin` — a plugin otherwise only about version bumping — because the task is registered per-project by `com.hivemq.tools.license` across **17 projects in 6 repositories**, and is absent from `hivemq-edge-composite`, so there is no single build script for it. Of the six `com.hivemq.*` convention plugins, `edge-version-updater` is the only one all 17 apply (`repository-convention` and `jacoco-convention` miss four, `errorprone`/`nullaway` three, `spotless` two). The `plugins.withId` guard keeps it inert where the license plugin is absent.

Verified against a populated cache: without the change `cyclonedxDirectBom` returns **FROM-CACHE**; with it the task executes and regenerates `bom.json`/`bom.xml`. Coverage spot-checked on `hivemq-edge-module-databases` and `hivemq-edge-adapter-sdk`, two of the four `repository-convention` would have missed.

**Remove once the underlying caching issue is fixed** — a permanently non-cacheable SBOM task costs time on every CI run.

### DIFF
--- a/edge-plugins/src/main/kotlin/com/hivemq/versionupdater/VersionUpdaterPlugin.kt
+++ b/edge-plugins/src/main/kotlin/com/hivemq/versionupdater/VersionUpdaterPlugin.kt
@@ -8,6 +8,26 @@ class VersionUpdaterPlugin : Plugin<Project> {
 
     override fun apply(project: Project) {
         project.tasks.register<UpdateVersionTask>(UPDATE_VERSION_TASK_NAME)
+
+        // Workaround for a CI build-cache issue: cyclonedxDirectBom must not be served from the cache.
+        //
+        // The task is registered per-project by com.hivemq.tools.license, which applies the CycloneDX plugin
+        // and then configures cyclonedxDirectBom against that project's runtimeClasspath. Seventeen projects
+        // across six repositories apply it, and the composite root does not, so there is no single build
+        // script to put this in.
+        //
+        // It lives here, in a plugin otherwise concerned only with version bumping, because of the six
+        // com.hivemq.* convention plugins this is the only one every affected project applies --
+        // repository-convention and jacoco-convention miss four of them, errorprone and nullaway three,
+        // spotless two. The withId guard means this is inert wherever the license plugin is absent, which
+        // includes the roots that apply this plugin without ever registering the task.
+        //
+        // Remove once the caching issue itself is fixed; nothing else here depends on it.
+        project.plugins.withId("com.hivemq.tools.license") {
+            project.tasks.named("cyclonedxDirectBom") {
+                outputs.cacheIf { false }
+            }
+        }
     }
 }
 
--- a/hivemq-edge/src/main/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImpl.java
@@ -747,11 +747,24 @@ public class ProtocolAdaptersResourceImpl extends AbstractApi implements Protoco
             log.debug("Original exception: ", e);
             return errorResponse(new InternalServerError(null));
         } catch (final @NotNull ExecutionException e) {
+            // Prefer the adapter's own account of what went wrong. TagSchemaCreationOutputImpl.fail(String)
+            // records the adapter's sentence on the output and completes the future with a fixed
+            // "Json schema creation for tag failed.", so reading only the cause discards the only text that
+            // says anything -- and every one of the OPC UA adapter's schema failures takes that route. The
+            // symptom was a 500 whose whole body was that fixed sentence, which QA could not diagnose from and
+            // reported as a possible regression in the direction parameter (EDG-894 P7); the adapter had in
+            // fact said "Discovery failed: ClientConnection not connected or not initialized" and nothing
+            // carried it to the caller.
+            //
+            // Falling back to the cause rather than replacing it: fail(Throwable, null) leaves no message on
+            // the output, and there the throwable is the only account there is.
             final Throwable cause = e.getCause();
+            final String adapterReason = tagSchemaCreationOutput.getMessage();
             final String message;
-            if (cause == null) {
+            if (adapterReason != null && !adapterReason.isBlank()) {
+                message = adapterReason;
+            } else if (cause == null) {
                 message = "unknown error";
-                ;
             } else {
                 message = cause.getMessage();
             }
--- a/hivemq-edge/src/test/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImplTest.java
+++ b/hivemq-edge/src/test/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImplTest.java
@@ -936,4 +936,91 @@ class ProtocolAdaptersResourceImplTest {
         assertThat(response.getLocation().toString())
                 .isEqualTo("/api/v1/management/protocol-adapters/schema/adapter/tag?direction=SOUTHBOUND");
     }
+
+    /**
+     * EDG-894 P7: when schema generation fails, the 500 must say what the adapter said.
+     * <p>
+     * {@code TagSchemaCreationOutput.fail(String)} records the adapter's reason on the output and completes the
+     * future with a fixed {@code "Json schema creation for tag failed."}. This layer read only the cause, so the
+     * entire body of the 500 was that fixed sentence — and every schema failure in the OPC UA adapter takes that
+     * route. QA met it as an unexplained 500 on an ordinary VALUE tag and could not tell from the response
+     * whether the fault was in schema generation or in the {@code direction} parameter that had just been added,
+     * so it was reported as a possible API compatibility break. The adapter had in fact said its connection was
+     * not established, and nothing carried that sentence to the caller.
+     */
+    @Test
+    void getSchema_whenTheAdapterGivesAReason_thenTheResponseCarriesIt() {
+        mockAdapterFailingSchemaWith("Discovery failed: ClientConnection not connected or not initialized");
+
+        final Response response = protocolAdaptersResource.getSchema("adapter", "tag", null);
+
+        assertEquals(500, response.getStatus());
+        assertThat(response.getEntity().toString())
+                .as("the operator has to be told which condition stopped the schema being built")
+                .contains("ClientConnection not connected")
+                .doesNotContain("Json schema creation for tag failed.");
+    }
+
+    @Test
+    void getSchema_whenTheAdapterGivesNoReason_thenTheCauseIsStillReported() {
+        // fail(Throwable, null) leaves no message on the output, and there the throwable is the only account
+        // there is. Preferring the adapter's reason must not mean discarding the cause when there isn't one.
+        final ProtocolAdapterWrapper wrapper = mock();
+        final ProtocolAdapter adapter = mock();
+        when(wrapper.getAdapter()).thenReturn(adapter);
+        doAnswer(invocation -> {
+                    final TagSchemaCreationOutput output = invocation.getArgument(1);
+                    output.fail(new IllegalStateException("the node is not readable"), null);
+                    return null;
+                })
+                .when(adapter)
+                .createTagSchema(any(), any());
+        when(protocolAdapterManager.getProtocolAdapterWrapperByAdapterId("adapter"))
+                .thenReturn(Optional.of(wrapper));
+
+        final Response response = protocolAdaptersResource.getSchema("adapter", "tag", null);
+
+        assertEquals(500, response.getStatus());
+        assertThat(response.getEntity().toString()).contains("the node is not readable");
+    }
+
+    @Test
+    void getSchema_withoutDirection_takesTheSamePathAsNorthbound() {
+        // The other half of P7, and the half that refutes its stated hypothesis. The finding could not reach the
+        // NORTHBOUND call to compare, and offered "if NORTHBOUND succeeds, the defect is the default direction".
+        // It cannot: an absent direction resolves to NORTHBOUND, and createTagSchema runs and fails before the
+        // direction is consulted at all, so the two answers are identical whether the adapter succeeds or fails.
+        mockAdapterFailingSchemaWith("Discovery failed: ClientConnection not connected or not initialized");
+        final Response withoutDirection = protocolAdaptersResource.getSchema("adapter", "tag", null);
+
+        mockAdapterFailingSchemaWith("Discovery failed: ClientConnection not connected or not initialized");
+        final Response northbound = protocolAdaptersResource.getSchema("adapter", "tag", "NORTHBOUND");
+
+        assertEquals(withoutDirection.getStatus(), northbound.getStatus());
+        assertThat(withoutDirection.getEntity().toString())
+                .isEqualTo(northbound.getEntity().toString());
+
+        mockAdapterWithTagSchema("adapter");
+        final Response okWithout = protocolAdaptersResource.getSchema("adapter", "tag", null);
+        mockAdapterWithTagSchema("adapter");
+        final Response okNorthbound = protocolAdaptersResource.getSchema("adapter", "tag", "NORTHBOUND");
+
+        assertEquals(200, okWithout.getStatus());
+        assertThat((JsonNode) okWithout.getEntity()).isEqualTo((JsonNode) okNorthbound.getEntity());
+    }
+
+    private void mockAdapterFailingSchemaWith(final @NotNull String reason) {
+        final ProtocolAdapterWrapper wrapper = mock();
+        final ProtocolAdapter adapter = mock();
+        when(wrapper.getAdapter()).thenReturn(adapter);
+        doAnswer(invocation -> {
+                    final TagSchemaCreationOutput output = invocation.getArgument(1);
+                    output.fail(reason);
+                    return null;
+                })
+                .when(adapter)
+                .createTagSchema(any(), any());
+        when(protocolAdapterManager.getProtocolAdapterWrapperByAdapterId("adapter"))
+                .thenReturn(Optional.of(wrapper));
+    }
 }
--- a/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/condition/EventPropagationVerifier.java
+++ b/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/condition/EventPropagationVerifier.java
@@ -50,9 +50,10 @@ import org.jetbrains.annotations.Nullable;
  * failure, unmappable continuation, or hierarchy deeper than the safety bound is {@link Result.Unverified};
  * callers retain compatibility for those servers but must make that uncertainty visible to the operator.
  * <p>
- * The Server object is the one notifier the walk is not asked about at all: the specification guarantees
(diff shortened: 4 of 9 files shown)


hivemq-edge/src/main/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImpl.java
```diff
@@ -747,11 +747,24 @@ public int getDepth() {
...
             final String message;
-            if (cause == null) {
+            if (adapterReason != null && !adapterReason.isBlank()) {
+                message = adapterReason;
+            } else if (cause == null) {
                 message = "unknown error";
-                ;
             } else {
                 message = cause.getMessage();
             }
```
marregui, line 770: `cause.getMessage()` can be null here, which puts an uninformative 500 back on one path — the same defect class this commit exists to remove. `TagSchemaCreationOutputImpl`'s future is `.orTimeout(30, SECONDS)`; on that path nothing calls `fail(...)`, so `message` stays null *and* the `TimeoutException` is constructed with no message, and the response detail comes out as literal `null`. Same shape if `notSupported()` or `adapterNotStarted()` ever carry a blank message.

The adapter reason one line up is blank-guarded; the cause is not. `Browsing.describeException` already sets the precedent for the fallback — *"falls back to the class name where there is no message, since several of the exceptions reachable here carry none and `failed: null` describes nothing"* — same reasoning, same answer:

```suggestion
            } else if (cause == null) {
                message = "unknown error";
            } else if (cause.getMessage() == null || cause.getMessage().isBlank()) {
                message = cause.toString();
            } else {
                message = cause.getMessage();
            }
```

Pre-existing rather than introduced by you — flagging it because this branch now owns the decision, and it already handles the sibling case one line up.

hivemq-edge/src/test/java/com/hivemq/api/resources/impl/ProtocolAdaptersResourceImplTest.java
```diff
@@ -936,4 +936,91 @@ void getWritingSchema_redirectsWithTheSouthboundDirection() {
         assertThat(response.getLocation().toString())
                 .isEqualTo("/api/v1/management/protocol-adapters/schema/adapter/tag?direction=SOUTHBOUND");
     }
+
+    /**
```
marregui, line 940: Heads-up rather than a finding, and the work is mine.

#1719 appends its new tests at exactly this anchor, immediately after `getWritingSchema_redirectsWithTheSouthboundDirection`, so the two will conflict textually. The sharper problem is in the fixture above: #1719 adds a tenth constructor parameter to `ProtocolAdaptersResourceImpl` (`ProtocolAdapterConfigConverter`), and this file has a single construction site — so whichever PR lands second gets a **compile error**, not a conflict marker anyone will notice in review.

The main-source regions do not overlap (yours is the `getSchema` catch block, mine is `addAdapter`/`updateAdapter`/`createCompleteAdapter`), so there is no semantic interaction — purely ordering.

Land this one first and I will rebase #1719 onto it. Nothing for you to do.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/condition/EventPropagationVerifier.java
```diff
@@ -87,6 +88,59 @@ private static boolean isRootNotifier(final @NotNull NodeId notifier) {
...
+     * <p>
+     * <b>Why anything is asked at all.</b> {@link #isRootNotifier} settles the <em>hierarchy</em> question and
+     * nothing else, but the walk it replaces was answering two questions rather than one. Alongside "does this
+     * node's event path reach the selected notifier" it was incidentally answering "is this node something the
+     * server has", because a node the server does not hold produces a browse failure rather than a clean
+     * negative. Short-circuiting the walk dropped both, and §8.3.2 licenses dropping only the first: it
+     * promises that every event of the server is accessible at the Server object, which says nothing whatever
+     * about whether some node id the operator typed is an event source, or a node, or anything at all.
+     * <p>
+     * The gap that left is the one QA reported as EDG-894 P6. A query tag rooted at the Server object — the
```
marregui, line 102: Scope note on this paragraph, tied to the "P6" label in the description.

EDG-894's P6 is a query tag whose `conditionNode` sits under a **different** notifier. For a non-root notifier `isRootNotifier` is false, the walk runs, returns `OutsideHierarchy`, and `OpcUaSubscriptionLifecycleHandler:2190` drops the tag by name — untouched by this diff, and QA confirmed it green at `3b3e48f28` on 2026-08-14. What this method fixes is the residue that `f480b2f94`'s exemption left behind under the *root* notifier.

So "The gap that left is the one QA reported as EDG-894 P6" is a shade broader than the code: it is the gap left *inside* P6, not P6.

Also unresolved, and worth settling before this claims P6: Estefania re-listed P6 as open on 2026-08-18 with the sibling wording, having confirmed it green on 08-14. Either that re-listing was stale, or something regressed in between — plausibly this exemption, if her fixture's notifier is in fact the Server object. That would make this the right fix for the right reason, but nobody has established which. I have put the question on EDG-894 rather than here.

modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/config/tag/OpcuaTagKind.java
```diff
@@ -65,5 +70,42 @@ public enum OpcuaTagKind {
...
+     * Written as a creator rather than by enabling {@code CoercionConfig} on the mapper: the mapper is shared
+     * by every adapter's configuration, and blank-means-unset is true of this field rather than of every enum
+     * in the product.
+     *
+     * @throws IllegalArgumentException when a non-blank value names no kind. Unlike blank, that is a typo, and
+     *                                 a tag whose kind was mistyped must not silently become a plain value —
+     *                                 it would subscribe to an alarm node's Value attribute and publish
+     *                                 nothing.
+     */
+    @JsonCreator
```
marregui, line 94: Two small things on the creator, neither blocking.

**The one guarantee you hand-checked is the one left unpinned.** The description says: *"Checked rather than assumed: adding `@JsonCreator` to `OpcuaTagKind` does not change how the generated tag config schema renders it — still all four constants with default `VALUE`, and `type` still lists 22."* I went looking for the test that holds that — every `generateJsonSchema` consumer under test in both repos, and enum-constant assertions on the kinds — and there is not one. This is the change's largest blast radius, since the tag schema drives the UI form, and EDG-899 has already shown once this quarter that enum *rendering* and enum *parsing* can disagree with nothing failing. Your own standard elsewhere in this PR is to pin what you checked — you did exactly that for the direction equivalence, one file over.

**Undocumented behaviour change: `{"kind": 0}` used to work.** `FAIL_ON_NUMBERS_FOR_ENUMS` is off by default, so Jackson read a numeric `kind` as an ordinal and `0` resolved to `VALUE`. With a String-delegating creator the `0` is stringified and refused as `Unknown OPC UA tag kind '0'`. Consistent with the doctrine here and probably an improvement, but it is a behaviour change on a config surface with no test and no mention. `type` already had a creator, so it is unaffected. *(Inferred from Jackson's defaults — I did not execute it.)*

edge-plugins/src/main/kotlin/com/hivemq/versionupdater/VersionUpdaterPlugin.kt
```diff
@@ -8,6 +8,26 @@ class VersionUpdaterPlugin : Plugin<Project> {
 
     override fun apply(project: Project) {
         project.tasks.register<UpdateVersionTask>(UPDATE_VERSION_TASK_NAME)
+
+        // Workaround for a CI build-cache issue: cyclonedxDirectBom must not be served from the cache.
```
marregui, line 12: Scope question on `a76baf28c`, not an objection to the change. The commit message is one of the more useful ones I have read this year — the placement argument, the seventeen-projects-six-repositories survey, and verifying against a populated cache rather than reading the API are all exactly right, and the removal note is there.

**But it is unrelated to EDG-894, it is temporary by its own description, and it lands in `edge-plugins`** — build infrastructure for seventeen projects across six repositories — inside a QA-findings PR. Three costs: it cannot be reverted independently of the three OPC UA fixes; it is invisible to anyone triaging CI, because nobody looks for a build-cache workaround in a PR titled "QA findings P6, P7 and P8"; and it can collide with whatever the platform team lands, since they are on the master-red build right now.

**There may be a larger reason to split it out.** `cyclonedxDirectBom` and `updateThirdPartyLicenses` are both registered by `com.hivemq.tools.license`. master has been red since 2026-08-16 on `updateThirdPartyLicenses` — *"No license found for `com.github.stephenc.jcip:jcip-annotations:1.0-1"`* — and locally I see the same pair behaving oddly around the cache: `cyclonedxDirectBom` failing to load its entry with a 403, `updateThirdPartyLicenses` restored `FROM-CACHE` and therefore never re-executing. If a poisoned or stale cache entry in that plugin's tasks is what is breaking master, **this commit may be closer to the root cause than to a workaround** — which is a strong argument for its own PR with its own title, not a weak one. *(Plausible, not established: I have not tied the two tasks together beyond the shared plugin and the coincident timing.)*

Suggest cherry-picking it onto its own branch and dropping it here — and if you agree the link above is worth chasing, saying so to whoever is holding the CI issue, because they are currently looking somewhere else.

marregui (approved):
Reviewed against EDG-894's description **and its comment history** — P7 and P8 exist only in Estefania's 2026-08-17 comment, P9 only in her 2026-08-18 one — rather than against this PR's own account of them.

**Approving.** The three fixes are correct everywhere I could check them, and several of the decisions are right for reasons that would have been easy to get wrong: `message` is `volatile` and assigned *before* `completeExceptionally`, which is the only ordering that works; `Unverified` rather than `OutsideHierarchy` keeps review-09 finding 4 intact; the trimming carried over from Jackson's default handling is what stops `" VALUE "` becoming a dropped tag. Four inline comments below, one of them a one-line `suggestion` you can commit from the browser. None is a blocker.

**One thing to fix in the description before merging, not in the code.** The PR says "P6" flat, but EDG-894's P6 is a narrowing node under a *different* notifier — non-root, so the walk still runs and still returns `OutsideHierarchy`, a path this diff does not touch and QA confirmed green at `3b3e48f28`. What this fixes is the root-notifier residue, which your own Linear comment named exactly ("the remaining P6 blind spot"). The practical cost of the flat label: `OpcuaTagEdgeConfigTest.testWrongNodeForKind_IsRefusedOrTheEmptyFilterIsVisiblyNamed` uses a sibling notifier, so re-running it against this head will show **no change** — which reads as the fix not working. Worth one sentence saying which test demonstrates the change, and ideally a functional case with the notifier set to the Server object.

**A merge-order note that is mine, not yours:** #1719 touches the same two files and adds a tenth constructor parameter to `ProtocolAdaptersResourceImpl`, so whichever lands second gets a compile error rather than a conflict marker. Land this one first — it is the older work and QA is waiting on it — and I will rebase mine onto it. Detail on the inline comment at the test file.

**`a76baf28c` arrived while I was reading and is not part of this.** It is careful work and I have no objection to the change itself — my question is only whether it should ride here; see the inline comment. Everything above was reviewed at `b5a1ac60f`; that commit touches no file the other three do, so nothing above is stale.

P9 and the P6 test-state question are going on EDG-894 rather than here, since neither is code.

PR #1723, opened 2026-08-21
EDG-920: Bump netty to 4.2.17.Final to clear SNYK-JAVA-IONETTY-19005879

caoccao:
Resolves [EDG-920](https://linear.app/hivemq/issue/EDG-920/bump-netty-to-4217final-in-hivemq-edge-to-clear-critical-snyk-finding).

## Problem

The `Run Snyk scan on PRs` workflow fails its `Check delta` step with three new **Critical** findings. All three are the same advisory reached by three different paths:

```
SNYK-JAVA-IONETTY-19005879 — Improper Check for Unusual or Exceptional Conditions
Critical, CVSS 9.1 — io.netty:netty-handler@4.2.16.Final
Fixed in: 4.2.17.Final, 4.1.137.Final

1/3  io.netty:netty-handler@4.2.16.Final                              (direct)
2/3  com.hivemq:hivemq-mqtt-client@1.3.17 => io.netty:netty-handler   (transitive)
3/3  io.netty:netty-codec-http@4.2.16.Final => io.netty:netty-handler (transitive)
```

The two earlier `snyk test` steps also exit 1, but those are `continue-on-error` by design — they exit non-zero whenever any vulnerability exists at all. `Check delta` is the actual gate, and it fails only on issues that are new relative to the baseline project.

This is not specific to any one PR: `master` carries the same `4.2.16.Final`, so the gate fails on every PR that runs the scan. It happened to surface on [#1717](https://github.com/hivemq/hivemq-edge/pull/1717), which is unrelated to netty.

## Root cause

Cross-repo version-catalog drift. Netty was bumped in two of the composite repos but not the third:

| Repo | netty |
| --- | --- |
| `hivemq-edge-commercial-modules` | `4.2.17.Final` |
| `hivemq-edge-test` | `4.2.17.Final` |
| `hivemq-edge` | `4.2.16.Final` ← the laggard |

The other four repos declare no netty version at all. The catalogs live in separate repos, so a bump has to be applied to each one.

## Changes

One line in `gradle/libs.versions.toml`:

```diff
-netty = "4.2.16.Final"
+netty = "4.2.17.Final"
```

All six `netty-*` library aliases share that `version.ref`, so the single bump covers them all. `4.2.17.Final` is the latest 4.2.x and matches what the sibling repos already pin, so this restores alignment rather than introducing a new spread.

## Verification

Snyk runs `snyk test --configuration-matching=^runtimeClasspath$ hivemq-edge/hivemq-edge`, so that exact configuration was resolved:

```
io.netty:netty-buffer:4.2.17.Final       io.netty:netty-handler:4.2.17.Final
io.netty:netty-codec:4.2.17.Final        io.netty:netty-transport:4.2.17.Final
io.netty:netty-codec-http:4.2.17.Final   io.netty:netty-common:4.2.17.Final
```

Both transitive paths are pulled up by Gradle conflict resolution — the mqtt-client path resolves `4.1.136.Final -> 4.2.17.Final`, and `netty-codec-http` moves with the shared ref. No `4.2.16.Final` reference remains in any of the seven repos, and there are no lockfiles or version-pinned license manifests to regenerate.

- `:hivemq-edge:compileJava`, `:hivemq-edge:compileTestJava` and `:hivemq-edge-module-modbus:compileJava` pass.
- `deno task plugins-edge-composite` (cross-repo version alignment) passes: exit 0, 0 errors, 0 `Found version conflict`.
- No Java source touched, so Spotless has nothing to reformat.

## Follow-ups (not in this PR)

- `Check for new issues (hivemq-edge-frontend)` was skipped once the Java step failed, so the frontend half of the scan is still unverified.
- Netty landing in two repos but not the third suggests the update is not grouped across the composite's catalogs — worth checking so this does not drift again.


marregui (approved):
LG8TM

PR #1727, opened 2026-08-25
EDG-931: Stop six core API test classes sharing a hardcoded HTTP port

mschoenert-hivemq:
## Description (the why)

Six test classes each start a real `JaxrsHttpServer` on the **same hardcoded** `TEST_HTTP_PORT = 8088`: `JaxrsSSLTests`, `BasicAuthenticationTests`, `BearerTokenAuthTests`, `ChainedAuthTests`, `EnforceApiAuthTest` and `LdapAuthenticationTests`.

Sequentially on an idle machine they never clash. On a shared CI agent they do — and the failure mode is not a failed test. Whichever class binds second blocks in its setup and never returns.

**Three consecutive CI builds on the EDG-929 branch hung there and had to be aborted by hand:**

| build | last test to complete | silent for |
| -- | -- | -- |
| #1 | `JaxrsSSLTests.testGetResource()` | 31 min |
| #2 | `EnforceApiAuthTest.testGetSecuredResourceWithValidCreds()` | ~22 min |
| #3 | `EnforceApiAuthTest.testGetSecuredResourceWithValidCreds()` | same point |

The signature is distinctive: the last line is a **passing** test, the next class never prints even `STARTED`, and every test that started also finished — so nothing is stuck *inside* a test.

Linear: [EDG-931](https://linear.app/hivemq/issue/EDG-931/stop-six-core-api-test-classes-sharing-a-hardcoded-http-port)

- [ ] No test class binds a port shared with another class
- [ ] The six classes run concurrently without colliding, in any order
- [ ] One port-picking approach across the repo, with no per-site judgement call

## Non-Goals

- A default test timeout for `hivemq-edge` — the module has none, which is why the hangs were silent, but a JUnit default timeout governs test *execution*, not a `@BeforeAll` that never returns. It would not have caught these. Worth doing under its own issue.
- The three database lifecycle tests in `hivemq-edge-test`, which pin a host port chosen at class load — real, different repo, tracked separately.

## Notes for the reviewer

**Each class now takes its port from `util.RandomPortGenerator` in setup.** That helper already exists in this source set, and `JaxrsResourceTests` — the seventh class in this family, and the only one that never had the problem — has always used it. This makes the other six match their own neighbour.

**On the check-then-bind window.** The generator picks a random number, proves it free by opening a socket, closes it, and returns the number, so there is a formal gap before the server binds. It is negligible here: the range is 40000 wide, `Math.random()` is seeded per JVM so concurrent forks do not correlate, the window is microseconds, and a CI agent runs essentially nothing else. It is used at ~150 sites across both repos without producing collisions.

**Why not bind to port 0**, which has no window at all because the OS chooses and binds in one step: it only applies where the test itself creates the socket. Roughly **145 of ~150** port acquisitions across both repos cannot use it — the port must be known *before* anything binds, because it goes into a config XML, a builder, or a fake server's constructor, and an embedded Edge binds it later. Nothing exists to interrogate at the moment the number is needed.

So bind-to-zero could only ever cover a handful of sites. Two tests did use it — `OidcServiceImplRedirectTest` and `HiveMQEdgeHttpServiceImplTest` — and the second commit converts them to the generator too. **Two exceptions cost more in inconsistency than the theoretical window is worth.**

Verified: the whole `com.hivemq.api` package plus both converted tests, green before and after.

### DIFF
--- a/hivemq-edge/src/test/java/com/hivemq/api/JaxrsSSLTests.java
+++ b/hivemq-edge/src/test/java/com/hivemq/api/JaxrsSSLTests.java
@@ -56,10 +56,9 @@ public class JaxrsSSLTests {
 
     protected final Logger logger = LoggerFactory.getLogger(JaxrsSSLTests.class);
 
-    // Picked per class rather than shared (EDG-931). Six classes in this source set used to bind the same
-    // hardcoded 8088; whichever ran second on a busy agent could block forever in setup, with no failure
-    // and no timeout to end it -- three CI builds had to be aborted by hand.
-    private static int testHttpPort;
+    // A random free port, so that tests running in parallel do not conflict. A conflict would surface
+    // as a ProcessingException out of startServer(), logged as "The port ... is already in use".
+    private int testHttpPort;
     static final int CONNECT_TIMEOUT = 1000;
     static final int READ_TIMEOUT = 1000;
     static final String HTTPS = "https";
--- a/hivemq-edge/src/test/java/com/hivemq/api/JaxrsSSLTests.java
+++ b/hivemq-edge/src/test/java/com/hivemq/api/JaxrsSSLTests.java
@@ -69,10 +69,8 @@ public class JaxrsSSLTests {
 
     @BeforeEach
     public void setUp() throws Exception {
-        testHttpPort = RandomPortGenerator.get();
         testKeyStoreGenerator = new TestKeyStoreGenerator();
         JaxrsHttpServerConfiguration config = new JaxrsHttpServerConfiguration();
-        config.setPort(testHttpPort);
         config.setProtocol(HTTPS);
         context = getSslContext("testpassword");
         config.setSslContext(context);
@@ -88,6 +86,10 @@ public class JaxrsSSLTests {
         // -- ensure we supplied our own test mapper as this can effect output
         ObjectMapper mapper = new ObjectMapper();
         config.setObjectMapper(mapper);
+        // Taken last, so that the key generation above happens before the port is claimed rather
+        // than between claiming it and binding it.
+        testHttpPort = RandomPortGenerator.get();
+        config.setPort(testHttpPort);
         server = new JaxrsHttpServer(mock(), List.of(config), null);
         server.startServer();
     }
--- a/hivemq-edge/src/test/java/com/hivemq/api/JaxrsSSLTests.java
+++ b/hivemq-edge/src/test/java/com/hivemq/api/JaxrsSSLTests.java
@@ -46,6 +46,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import util.RandomPortGenerator;
 import util.TestKeyStoreGenerator;
 
 /**
@@ -55,7 +56,10 @@ public class JaxrsSSLTests {
 
     protected final Logger logger = LoggerFactory.getLogger(JaxrsSSLTests.class);
 
-    static final int TEST_HTTP_PORT = 8088;
+    // Picked per class rather than shared (EDG-931). Six classes in this source set used to bind the same
+    // hardcoded 8088; whichever ran second on a busy agent could block forever in setup, with no failure
+    // and no timeout to end it -- three CI builds had to be aborted by hand.
+    private static int testHttpPort;
     static final int CONNECT_TIMEOUT = 1000;
     static final int READ_TIMEOUT = 1000;
     static final String HTTPS = "https";
@@ -66,9 +70,10 @@ public class JaxrsSSLTests {
 
     @BeforeEach
     public void setUp() throws Exception {
+        testHttpPort = RandomPortGenerator.get();
         testKeyStoreGenerator = new TestKeyStoreGenerator();
         JaxrsHttpServerConfiguration config = new JaxrsHttpServerConfiguration();
-        config.setPort(TEST_HTTP_PORT);
+        config.setPort(testHttpPort);
         config.setProtocol(HTTPS);
         context = getSslContext("testpassword");
         config.setSslContext(context);
@@ -142,7 +147,7 @@ public class JaxrsSSLTests {
         HttpsURLConnection.setDefaultHostnameVerifier((string, ssls) -> true);
 
         HttpResponse response = HttpUrlConnectionClient.get(
-                null, getTestServerAddress(HTTPS, TEST_HTTP_PORT, "test/get"), CONNECT_TIMEOUT, READ_TIMEOUT);
+                null, getTestServerAddress(HTTPS, testHttpPort, "test/get"), CONNECT_TIMEOUT, READ_TIMEOUT);
         assertEquals(200, response.getStatusCode(), "Resource should exist");
     }
 
@@ -150,7 +155,7 @@ public class JaxrsSSLTests {
     public void testGetResourceOnWrongProtocol() {
         assertThrows(SocketException.class, () -> {
             HttpUrlConnectionClient.get(
-                    null, getTestServerAddress("http", TEST_HTTP_PORT, "test/get"), CONNECT_TIMEOUT, READ_TIMEOUT);
+                    null, getTestServerAddress("http", testHttpPort, "test/get"), CONNECT_TIMEOUT, READ_TIMEOUT);
         });
     }
 }
--- a/hivemq-edge/src/test/java/com/hivemq/api/auth/BasicAuthenticationTests.java
+++ b/hivemq-edge/src/test/java/com/hivemq/api/auth/BasicAuthenticationTests.java
@@ -44,6 +44,7 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import util.RandomPortGenerator;
 
 /**
  * @author Simon L Johnson
@@ -52,7 +53,10 @@ public class BasicAuthenticationTests {
 
     protected final Logger logger = LoggerFactory.getLogger(BasicAuthenticationTests.class);
 
-    static final int TEST_HTTP_PORT = 8088;
+    // Picked per class rather than shared (EDG-931). Six classes in this source set used to bind the same
+    // hardcoded 8088; whichever ran second on a busy agent could block forever in setup, with no failure
+    // and no timeout to end it -- three CI builds had to be aborted by hand.
+    private static int testHttpPort;
     // See BearerTokenAuthTests: the first request pays the server bootstrap cost and a 1s budget
     // makes whichever test runs first flaky on a loaded CI agent. Matches JaxrsResourceTests.
     static final int CONNECT_TIMEOUT = 5000;
@@ -63,8 +67,9 @@ public class BasicAuthenticationTests {
 
     @BeforeAll
     public static void setUp() throws Exception {
+        testHttpPort = RandomPortGenerator.get();
         final var config = new JaxrsHttpServerConfiguration();
-        config.setPort(TEST_HTTP_PORT);
+        config.setPort(testHttpPort);
         // -- ensure we supplied our own test mapper as this can affect output
         config.setObjectMapper(new ObjectMapper());
 
@@ -101,7 +106,7 @@ public class BasicAuthenticationTests {
         } else {
             headers = null;
         }
-        final var serverAddress = String.format("%s://%s:%s/%s", HTTP, "localhost", TEST_HTTP_PORT, path);
+        final var serverAddress = String.format("%s://%s:%s/%s", HTTP, "localhost", testHttpPort, path);
         return HttpUrlConnectionClient.get(headers, serverAddress, CONNECT_TIMEOUT, READ_TIMEOUT);
     }
 
--- a/hivemq-edge/src/test/java/com/hivemq/api/auth/BasicAuthenticationTests.java
+++ b/hivemq-edge/src/test/java/com/hivemq/api/auth/BasicAuthenticationTests.java
@@ -53,9 +53,8 @@ public class BasicAuthenticationTests {
 
     protected final Logger logger = LoggerFactory.getLogger(BasicAuthenticationTests.class);
 
-    // Picked per class rather than shared (EDG-931). Six classes in this source set used to bind the same
-    // hardcoded 8088; whichever ran second on a busy agent could block forever in setup, with no failure
-    // and no timeout to end it -- three CI builds had to be aborted by hand.
+    // A random free port, so that tests running in parallel do not conflict. A conflict would surface
+    // as a ProcessingException out of startServer(), logged as "The port ... is already in use".
     private static int testHttpPort;
     // See BearerTokenAuthTests: the first request pays the server bootstrap cost and a 1s budget
     // makes whichever test runs first flaky on a loaded CI agent. Matches JaxrsResourceTests.
--- a/hivemq-edge/src/test/java/com/hivemq/api/auth/BearerTokenAuthTests.java
+++ b/hivemq-edge/src/test/java/com/hivemq/api/auth/BearerTokenAuthTests.java
@@ -69,9 +69,8 @@ public class BearerTokenAuthTests {
 
     protected final Logger logger = LoggerFactory.getLogger(BearerTokenAuthTests.class);
 
-    // Picked per class rather than shared (EDG-931). Six classes in this source set used to bind the same
-    // hardcoded 8088; whichever ran second on a busy agent could block forever in setup, with no failure
-    // and no timeout to end it -- three CI builds had to be aborted by hand.
+    // A random free port, so that tests running in parallel do not conflict. A conflict would surface
+    // as a ProcessingException out of startServer(), logged as "The port ... is already in use".
     private static int testHttpPort;
     // The first request against the freshly started server pays the Jersey/Jackson bootstrap cost,
     // which is orders of magnitude slower than every later request. A 1s budget is not enough for
--- a/hivemq-edge/src/test/java/com/hivemq/api/auth/BearerTokenAuthTests.java
+++ b/hivemq-edge/src/test/java/com/hivemq/api/auth/BearerTokenAuthTests.java
@@ -60,6 +60,7 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import util.RandomPortGenerator;
 
 /**
  * @author Simon L Johnson
@@ -68,7 +69,10 @@ public class BearerTokenAuthTests {
 
     protected final Logger logger = LoggerFactory.getLogger(BearerTokenAuthTests.class);
 
-    static final int TEST_HTTP_PORT = 8088;
+    // Picked per class rather than shared (EDG-931). Six classes in this source set used to bind the same
+    // hardcoded 8088; whichever ran second on a busy agent could block forever in setup, with no failure
+    // and no timeout to end it -- three CI builds had to be aborted by hand.
+    private static int testHttpPort;
     // The first request against the freshly started server pays the Jersey/Jackson bootstrap cost,
     // which is orders of magnitude slower than every later request. A 1s budget is not enough for
     // that on a loaded CI agent, so the first test to run would time out at random. Matches
@@ -82,9 +86,10 @@ public class BearerTokenAuthTests {
 
     @BeforeAll
     public static void setUp() throws Exception {
+        testHttpPort = RandomPortGenerator.get();
 
         final JaxrsHttpServerConfiguration config = new JaxrsHttpServerConfiguration();
-        config.setPort(TEST_HTTP_PORT);
+        config.setPort(testHttpPort);
         // -- ensure we supplied our own test mapper as this can effect output
         config.setObjectMapper(objectMapper);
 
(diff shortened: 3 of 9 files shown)


hivemq-edge/src/test/java/com/hivemq/api/JaxrsSSLTests.java
```diff
@@ -55,7 +56,10 @@ public class JaxrsSSLTests {
 
     protected final Logger logger = LoggerFactory.getLogger(JaxrsSSLTests.class);
 
-    static final int TEST_HTTP_PORT = 8088;
+    // Picked per class rather than shared (EDG-931). Six classes in this source set used to bind the same
+    // hardcoded 8088; whichever ran second on a busy agent could block forever in setup, with no failure
+    // and no timeout to end it -- three CI builds had to be aborted by hand.
+    private static int testHttpPort;
```
marregui: Two things here, one edit.

**The comment is wrong twice.** It states a mechanism that doesn't happen (see the review body), and it says *per class* in the one class where that isn't true: this class starts the server in `@BeforeEach` (`:71`) and stops it in `@AfterEach` (`:96`), and has two `@Test` methods (`:144`, `:155`), so the port is drawn twice per run — per test.

**The field should follow that lifecycle.** All four usages (`:73`, `:76`, `:150`, `:158`) are in instance methods and `getTestServerAddress` takes the port as a parameter, so `private int` compiles as-is. A `static` mutated from `@BeforeEach` is a hazard the day anyone turns on JUnit's in-JVM parallelism — which is the direction this ticket is pushing.

```suggestion
    // One port per class (EDG-931): six classes in this source set shared a hardcoded 8088, which
    // blocks forking the core unit tests. Per test here, not per class -- this class starts and
    // stops its server in @BeforeEach/@AfterEach.
    private int testHttpPort;
```

hivemq-edge/src/test/java/com/hivemq/api/JaxrsSSLTests.java
```diff
@@ -66,9 +70,10 @@ public class JaxrsSSLTests {
 
     @BeforeEach
     public void setUp() throws Exception {
+        testHttpPort = RandomPortGenerator.get();
```
marregui: The pick is here; the bind is at `:92`. In between sit `new TestKeyStoreGenerator()` and `getSslContext()`, which generate a BouncyCastle RSA-2048 keypair and a `SHA256withRSA` self-signed cert (`TestKeyStoreGenerator.java:160-163`, `:175-184`). Measured on this machine, idle, Temurin 25: **65–200 ms** over five runs — and BC's software keygen is slower than the JDK provider, so treat that as a floor.

Move this line, carrying `config.setPort(testHttpPort)` from `:76`, down to immediately above `server = new JaxrsHttpServer(...)` at `:92`. `JaxrsHttpServerConfiguration.setPort` is a plain setter, so the config can be built first and the port set last.

Not a `suggestion` block because the paired edit is outside this hunk.

hivemq-edge/src/test/java/com/hivemq/api/auth/BasicAuthenticationTests.java
```diff
@@ -52,7 +53,10 @@ public class BasicAuthenticationTests {
 
     protected final Logger logger = LoggerFactory.getLogger(BasicAuthenticationTests.class);
 
-    static final int TEST_HTTP_PORT = 8088;
+    // Picked per class rather than shared (EDG-931). Six classes in this source set used to bind the same
+    // hardcoded 8088; whichever ran second on a busy agent could block forever in setup, with no failure
+    // and no timeout to end it -- three CI builds had to be aborted by hand.
+    private static int testHttpPort;
```
marregui: Same rewrite as `JaxrsSSLTests` — reasoning in the review body.

```suggestion
    // One port per class (EDG-931): six classes in this source set shared a hardcoded 8088, which
    // blocks forking the core unit tests.
    private static int testHttpPort;
```

hivemq-edge/src/test/java/com/hivemq/api/auth/BearerTokenAuthTests.java
```diff
@@ -68,7 +69,10 @@ public class BearerTokenAuthTests {
 
     protected final Logger logger = LoggerFactory.getLogger(BearerTokenAuthTests.class);
 
-    static final int TEST_HTTP_PORT = 8088;
+    // Picked per class rather than shared (EDG-931). Six classes in this source set used to bind the same
+    // hardcoded 8088; whichever ran second on a busy agent could block forever in setup, with no failure
+    // and no timeout to end it -- three CI builds had to be aborted by hand.
+    private static int testHttpPort;
```
marregui: Same rewrite as `JaxrsSSLTests` — reasoning in the review body.

```suggestion
    // One port per class (EDG-931): six classes in this source set shared a hardcoded 8088, which
    // blocks forking the core unit tests.
    private static int testHttpPort;
```

hivemq-edge/src/test/java/com/hivemq/api/auth/ChainedAuthTests.java
```diff
@@ -67,7 +68,10 @@ public class ChainedAuthTests {
 
     protected final Logger logger = LoggerFactory.getLogger(ChainedAuthTests.class);
 
-    static final int TEST_HTTP_PORT = 8088;
+    // Picked per class rather than shared (EDG-931). Six classes in this source set used to bind the same
+    // hardcoded 8088; whichever ran second on a busy agent could block forever in setup, with no failure
+    // and no timeout to end it -- three CI builds had to be aborted by hand.
+    private static int testHttpPort;
```
marregui: Same rewrite as `JaxrsSSLTests` — reasoning in the review body.

```suggestion
    // One port per class (EDG-931): six classes in this source set shared a hardcoded 8088, which
    // blocks forking the core unit tests.
    private static int testHttpPort;
```

hivemq-edge/src/test/java/com/hivemq/api/auth/EnforceApiAuthTest.java
```diff
@@ -55,17 +56,22 @@ public class EnforceApiAuthTest {
...
 
-    static final int TEST_HTTP_PORT = 8088;
     static final int CONNECT_TIMEOUT = 1000;
     static final int READ_TIMEOUT = 1000;
     static final String HTTP = "http";
 
+    // Picked per class rather than shared (EDG-931). Six classes in this source set used to bind the same
+    // hardcoded 8088; whichever ran second on a busy agent could block forever in this @BeforeAll, with no
+    // failure and no timeout to end it -- three CI builds had to be aborted by hand.
+    private static int testHttpPort;
```
marregui: Same rewrite as `JaxrsSSLTests` — reasoning in the review body.

```suggestion
    // One port per class (EDG-931): six classes in this source set shared a hardcoded 8088, which
    // blocks forking the core unit tests.
    private static int testHttpPort;
```

hivemq-edge/src/test/java/com/hivemq/api/auth/LdapAuthenticationTests.java
```diff
@@ -62,7 +63,10 @@ public class LdapAuthenticationTests {
 
     protected final Logger logger = LoggerFactory.getLogger(LdapAuthenticationTests.class);
 
-    static final int TEST_HTTP_PORT = 8088;
+    // Picked per class rather than shared (EDG-931). Six classes in this source set used to bind the same
+    // hardcoded 8088; whichever ran second on a busy agent could block forever in setup, with no failure
+    // and no timeout to end it -- three CI builds had to be aborted by hand.
+    private static int testHttpPort;
```
marregui: Same rewrite as `JaxrsSSLTests` — reasoning in the review body.

```suggestion
    // One port per class (EDG-931): six classes in this source set shared a hardcoded 8088, which
    // blocks forking the core unit tests.
    private static int testHttpPort;
```

hivemq-edge/src/test/java/com/hivemq/api/auth/LdapAuthenticationTests.java
```diff
@@ -73,6 +77,7 @@ public class LdapAuthenticationTests {
 
     @BeforeAll
     static void setUp() throws Exception {
+        testHttpPort = RandomPortGenerator.get();
```
marregui: The pick is here; the bind is at `:136`. In between sit the LDAP connection setup and `LdapTestConnection.createTestUser(...)` — a real round trip to the LLDAP container, seconds-scale under load. This is the widest window of the six by a long way.

Move this line down to immediately above `final var config = new JaxrsHttpServerConfiguration();` at `:117`. Same one-line change as `JaxrsSSLTests`, same reason.

marregui (approved):
Approving — nothing here blocks the merge. One port per class is right regardless of what wedged those builds, the conversion is complete, and I re-ran the eight affected classes on `081a21a5` (Temurin 25, composite root, `--rerun`): **60 passed, 0 failed**.

The inline suggestions are all one-liners you can take from the browser. Two rules behind them.

**The comment, in six files.** Its load-bearing clause is not what happens. The server is the JDK's `com.sun.net.httpserver` via `JdkHttpServerFactory` (`JaxrsHttpServer.java:31,44`), and `startServer()` catches the `ProcessingException`, logs *"The port '{}' is already in use"* and rethrows — `JaxrsHttpServer.java:140-153` and `:159-172`, with `// still throw the exception to mitigate a silent fail` on the rethrow. Measured on this branch, two servers on one port: the second throws after 37 ms. A PR description is read once; a comment is read for years, so the suggestions state the rule instead of the story. The rest of that argument is on the ticket, because what it changes is what EDG-931 may be closed as, and that isn't a PR question.

**"The window is microseconds" holds at four of the six sites.** In `JaxrsSSLTests` the pick is at `:73` and the bind at `:92`, with a BouncyCastle RSA-2048 keypair and a `SHA256withRSA` self-signed cert in between (`TestKeyStoreGenerator.java:160-163`) — 65–200 ms measured here, idle. In `LdapAuthenticationTests` the pick is at `:80` and the bind at `:136`, with a real LLDAP round trip in between. Exposure is linear in window width, so moving the pick next to the bind — which is exactly what commit 3's title already says — buys about four orders of magnitude for one line each.

**Two ACs aren't met by what's on the branch.** Both are do-it-or-amend-the-wording; neither blocks.

- *"The six classes run concurrently without colliding, in any order."* Nothing on the branch runs them concurrently, and nothing can: `maxParallelForks` appears nowhere in the repo (`hivemq-edge/build.gradle.kts:303` sets only `useJUnitPlatform()`, heap and `jvmArgs`), there is no `junit-platform.properties`, and `org.gradle.parallel` is unset. So the green-before-and-after run is sequential — the same evidence that existed at `bdc10cb5`, which can't distinguish the two states. A temporary `maxParallelForks = 4` inside `tasks.test` would settle it against the 2m53s → 2m04s / 2-failures baseline already in the ticket. Otherwise the honest claim is that the branch removes the blocker rather than proves the outcome.
- *"One port-picking approach across the repo."* `JaxrsResourceTests.java:46` still holds `static final int TEST_HTTP_PORT = RandomPortGenerator.get()`. Not a risk — class-load to `@BeforeAll` is sub-millisecond, nothing like the DB ITs' minutes — but it's the class the ticket holds up as the worked example, and after this PR it's the odd one out of seven.

One note, no action asked: `RandomPortGenerator` draws `[10000, 50000]`, and 17233 of those 40001 values — 43% — sit inside Linux's default ephemeral range (32768–60999), so the kernel can hand a drawn number to an outbound socket before the test binds it. Pre-existing, but it's the seam the whole check-then-bind acceptance rests on, and it's the competing term in the window arithmetic above. If it ever does bite, narrowing to `[10000, 32767]` clears both Linux and macOS.

hivemq-edge/src/test/java/com/hivemq/api/JaxrsSSLTests.java
mschoenert-hivemq: Both taken. `private int` now — a `static` mutated from `@BeforeEach` is a trap, and it compiles unchanged since all four usages are in instance methods.

On the comment I went further than your suggestion, because yours has the same flaw as mine: *"six classes in this source set shared a hardcoded 8088"* is still an account of the bug rather than what the code does. This comment sits in six files and will outlive the incident. It now reads:

```java
// A random free port, so that tests running in parallel do not conflict. A conflict would surface
// as a ProcessingException out of startServer(), logged as "The port ... is already in use".
private int testHttpPort;
```

The second sentence is your finding, kept because it is the string someone hitting this will grep for.

Note that once the comment stops claiming a lifecycle, `private int` states it by itself — so the "per test here, not per class" sentence is not needed. The class count, the port number and the aborted builds are all on the ticket.

hivemq-edge/src/test/java/com/hivemq/api/auth/BasicAuthenticationTests.java
mschoenert-hivemq: Same rewrite as `JaxrsSSLTests:62`. The mechanism you identified was the error; the wording differs from your suggestion for the reason given in that thread.

hivemq-edge/src/test/java/com/hivemq/api/auth/BearerTokenAuthTests.java
mschoenert-hivemq: Same rewrite as `JaxrsSSLTests:62`. The mechanism you identified was the error; the wording differs from your suggestion for the reason given in that thread.

hivemq-edge/src/test/java/com/hivemq/api/auth/ChainedAuthTests.java
mschoenert-hivemq: Same rewrite as `JaxrsSSLTests:62`. The mechanism you identified was the error; the wording differs from your suggestion for the reason given in that thread.

hivemq-edge/src/test/java/com/hivemq/api/auth/EnforceApiAuthTest.java
mschoenert-hivemq: Same rewrite as `JaxrsSSLTests:62`. The mechanism you identified was the error; the wording differs from your suggestion for the reason given in that thread.

hivemq-edge/src/test/java/com/hivemq/api/auth/LdapAuthenticationTests.java
mschoenert-hivemq: Same rewrite as `JaxrsSSLTests:62`. The mechanism you identified was the error; the wording differs from your suggestion for the reason given in that thread.

hivemq-edge/src/test/java/com/hivemq/api/JaxrsSSLTests.java
mschoenert-hivemq: Moved in `54d47c0db`: the pick and `config.setPort(...)` now sit immediately above the `JaxrsHttpServer` construction, so the RSA-2048 keypair and self-signed cert are generated before the port is claimed.

Your 65–200ms is right and my "microseconds" was wrong here. But the risk was small before and is small after — a hundred milliseconds is still a very small window for something else on this machine to take that particular port out of 40001. Worth doing because it costs a line and removes a question, not because it closed a hole.

hivemq-edge/src/test/java/com/hivemq/api/auth/LdapAuthenticationTests.java
mschoenert-hivemq: Moved in the same commit — the pick now sits just above the config construction, after `createTestUser`, so the LLDAP round trip is outside the window.

This is the wider of the two and it is the site that shows why my general claim was wrong: I generalised "microseconds" from the four sites where it holds, and a container round trip is exactly the case that breaks it. Still — seconds is a longer window, not a likely one. Nothing here was failing.

mschoenert-hivemq:
**Thanks — this found a real error, and four of the six points are now fixed.**

**The comment, in six files.** You're right and I was wrong. `JaxrsHttpServer` logs *"The port '{}' is already in use"* and rethrows — `// still throw the exception to mitigate a silent fail`. My comment taught the reader to expect a hang, which is the opposite of what happens. Fixed in `aea77844d`; I did not take your wording verbatim, and the thread on `JaxrsSSLTests:62` says why.

**`JaxrsSSLTests` per-test, not per-class.** Agreed on both halves; the `static` is gone. Your reasoning about in-JVM parallelism is the right one.

**The pick-to-bind window.** Both moved, in `54d47c0db`. Your measurements are correct and my "microseconds" was wrong for those two sites.

**But the conclusion does not follow from the number, and I would rather it not read as though it does.** Whether the window is microseconds, milliseconds or seconds, the practical risk stays very small: something else on a machine doing little but running these tests has to claim that one specific port out of 40001 inside that interval. Going from microseconds to seconds multiplies a very small number; it does not cross into likely. Nothing here was failing, and this change closes no exposure.

It is worth doing because it costs a line and makes the code obviously right rather than arguably right — which matters mainly because it stops the next reader spending an afternoon on the same question. That is the cost being paid, and it is a real one: this gap has been raised repeatedly, and **every time, the cause turned out to be something else** — a hardcoded shared port, a `static final` held for minutes, a leaked thread. Never this gap. [EDG-936](https://linear.app/hivemq/issue/EDG-936) puts that on the record in the source, with the bar for reopening it: a specific failing build, not a plausible argument.

**`JaxrsResourceTests`.** Correct, and it is the class the ticket holds up as the example. Fixed in `54d47c0db` — the `static final` is gone and it picks in `@BeforeAll` like the other six. Seven of seven now.

**The range.** Your 43% is exact: 17233 of 40001 values inside 32768–60999. Deferred deliberately. `RandomPortGenerator` exists in **three** copies across two repos and they have already diverged — only the `hivemq-edge-test` one probes UDP, and the core one leaks its probe socket if `getLocalPort()` throws. Fixing the range now means fixing it three times and hand-syncing them, which is exactly how they drifted. [EDG-936](https://linear.app/hivemq/issue/EDG-936) covers it after the monorepo: one copy, `[10000, 32767]`, a protocol argument, a purpose argument so an issued port can be correlated with a later bind failure, and a loud exhaustion failure.

**The concurrency AC — this one I would push back on.** You are right that nothing *on this branch* runs these classes concurrently, so from the PR alone the claim is unbacked. But it has been demonstrated, on a branch that contains `081a21a5f`: the full core unit suite, all 4796 tests with these seven classes among them, run at 2, 3 and 5 forks with three samples each, plus 4 and 8 — nine concurrent full-suite runs, all green, zero failures. More than the temporary `maxParallelForks = 4` you suggested. That work is [EDG-934](https://linear.app/hivemq/issue/EDG-934), which makes the fork count permanent (94s → 44s).

So: this branch removes the blocker, and the successor demonstrates the outcome. I will note that on the ticket rather than amend the AC.

Every test in the seven affected classes passes.

hivemq-edge/src/test/java/com/hivemq/api/auth/LdapAuthenticationTests.java
marregui: my instruction was wrong, and this is the half that didn't land.

`createTestUser` is outside the window now. `:125` is still inside it:

```java
testHttpPort = RandomPortGenerator.get();   // :120  <- pick
config.setPort(testHttpPort);               // :121
...
new LdapUsernameRolesProvider(ldapConnectionProperties, new SecurityLog())   // :125
...
server.startServer();                       // :137  <- bind
```

That constructor calls `ldapClient.start()` (`LdapUsernameRolesProvider.java:45`), which builds an `LDAPConnectionPool` with `initialConnections = 1` against the container — `LdapClient.java:140-148`. TCP connect plus a simple bind, synchronous. Our own log line at `:156`, *"LDAP client started successfully, connected to {}:{}"*, only prints once that constructor returns.

So there are two container round trips in this method and the pick sits between them. The comment at `:118-119` is true of the one above it and blind to the one below.

Mine to own: I said *"immediately above `final var config = ...` at `:117`"*. You did exactly that. The right anchor was `server = new JaxrsHttpServer(...)` — which is what I wrote for `JaxrsSSLTests` and not for this one.

fix: put it where `JaxrsSSLTests` now has it, immediately above the `JaxrsHttpServer` construction. Then the two sites are the same shape and there's no anchor left to get wrong.

```java
        final var config = new JaxrsHttpServerConfiguration();
        // -- ensure we supplied our own test mapper as this can effect output
        config.setObjectMapper(new ObjectMapper());

        final Set<IAuthenticationHandler> authenticationHandlers = new HashSet<>();
        authenticationHandlers.add(new BasicAuthenticationHandler(
                new LdapUsernameRolesProvider(ldapConnectionProperties, new SecurityLog())));

        final ApiConfigurationService apiConfigurationService = mock(ApiConfigurationService.class);
        when(apiConfigurationService.isEnforceApiAuth()).thenReturn(true);

        final ResourceConfig resourceConfig = new ResourceConfig();
        resourceConfig.register(new ApiAuthenticationFeature(authenticationHandlers, apiConfigurationService));
        resourceConfig.register(TestApiResource.class);
        resourceConfig.register(TestPermitAllApiResource.class);
        resourceConfig.register(TestResourceLevelRolesApiResource.class);

        // Taken last, so that the container round trips above happen before the port is claimed
        // rather than between claiming it and binding it.
        testHttpPort = RandomPortGenerator.get();
        config.setPort(testHttpPort);

        server = new JaxrsHttpServer(mock(), List.of(config), resourceConfig);
        server.startServer();
```

Not a `suggestion` block: the two edits straddle a hunk boundary, and GitHub can only rewrite a contiguous span that's already in the diff — `:136` isn't.

net: not the race argument again. EDG-936's own rule is *"take the port immediately before binding"*; this is the one site on the branch that doesn't yet meet it.

mschoenert-hivemq: Right on both counts, and thanks for catching the half that didn't land.

I verified the second round trip before agreeing: `LdapUsernameRolesProvider`'s constructor calls `ldapClient.start()` (`LdapUsernameRolesProvider.java:44`), which is where the pool gets opened. So the pick was sitting between `createTestUser` and that — the comment named one round trip and was blind to the other.

Fixed in `244066e83`, at the anchor you gave: immediately above the `JaxrsHttpServer` construction, same shape as `JaxrsSSLTests`. The comment now names both round trips rather than one.

Ran it against a real LLDAP container — 4/4 pass.

On owning the anchor: I followed your line literally rather than the rule behind it, which is the more useful thing to take from this. `JaxrsSSLTests` had the right anchor in your own review and I did not notice the two were inconsistent.

And agreed it is not the race argument. The rule is "take the port immediately before binding" — this site simply did not meet it yet.

marregui (approved):
Approving.

The move landed exactly right: `LdapAuthenticationTests:135-136` now sits three lines above the bind, with both container round trips ahead of it. The comment naming both — `createTestUser` *and* the `LdapUsernameRolesProvider` constructor opening the pool — is better than what I asked for; it tells the next reader why the pick is where it is rather than just that it moved.

Checked on `fdce62be2`:

- Pick-to-bind window in all seven classes: config setters, mocks and `ResourceConfig` registration only. No I/O left in any of them.
- Scope still test-only, 9 files. The master merge (`fe00ebae5`) touched none of them.
- Local run, Temurin 25, composite root, `--rerun`: **60 passed, 0 failed**.

One thing that is not code: GitHub Actions has no run for this head — 0 workflow runs 24 minutes after the push, so `check-backend` and the Snyk gate have not reported. The composite Jenkins job is green on `fdce62be2`, which covers the tests and Spotless, so this looks like a missed trigger rather than a problem. Worth a re-trigger before merging if those checks are required.

Issue #100, opened 2023-09-11
Create default logo template and sizing for adapters


simon622:
Currently all logos are sourced and created in different styles. We should standardise the naming and the design of these. We should work with burkhard & team to come up with a common template which we can add to along with defining common dimensions.

sfrehse:
Kicking off with Burkhard.

PR #1738, opened 2026-08-28
EDG-956: Never hand out the same port twice in a row

mschoenert-hivemq:
## Description (the why)

[EDG-956](https://linear.app/hivemq/issue/EDG-956/fix-randomportgenerator-for-good)

**Paired with [hivemq-edge-test#PAIR](PAIR) — the same file lives in both repositories and both must land.**

`RandomPortGenerator` proves a port free by binding it, then **closes the socket** and returns the number. The caller binds it later — often much later, because the number usually goes into a configuration file and is bound during startup.

That check-then-use window is survivable because a collision needs **two** independent things to go wrong:

1. another acquisition must fall inside the window, **and**
2. it must draw the same number out of 40001.

Both are needed, which is why ~150 call sites coexist without colliding.

### The case where the first protection is removed

A caller that takes two ports in a row defeats it outright. `EmbeddedHiveMQExtension.applyConfig()` draws the MQTT listener port and then the API port as consecutive statements, one thread:

```java
port = RandomPortGenerator.get();
apiPort = RandomPortGenerator.get();   // guaranteed inside the first draw's window
```

The second draw is not *likely* to land inside the first one's window — it **is** inside it, with certainty. Only the 1-in-40001 remains.

**Observed on master #100.** One embedded instance bound port 30876 for its MQTT listener and tried to bind the same number for its HTTP API a millisecond later:

```
00:31:35.472 Started MQTT TCP Listener on address 0.0.0.0 and on port 30876
00:31:35.473 Starting WebServer with protocol 'http' on port 30876
00:31:35.531 ERROR ... The port '30876' is already in use.
```

`BasicBridgeIT` failed 1 second after starting, against a 20-second latch — the Edge could not start at all.

**The rate this produces:** 1/40001 per exposed instance, ~814 such instances per build, giving **~2% of builds**. Expected 0.56 affected builds out of 28 captured; exactly 1 observed.

## What changed

**A recently-issued filter.** A number handed out is remembered in a fixed 1024-slot array, and neither it nor its bucket siblings are issued again for the next 32 requests:

```java
final int bucket = randomNumber % LAST_ISSUED_AT.length;
// Consult the array BEFORE opening any socket: a rejected candidate then costs one array read
// rather than a socket open and close.
if (LAST_ISSUED_AT[bucket] + BLOCK_FOR >= counter) {
    continue;
}
if (isFree(randomNumber, protocol)) {
    LAST_ISSUED_AT[bucket] = counter;
    return randomNumber;
}
```

Fixed memory, O(1), entries expire on their own — nothing to reset between tests or classes. An unbounded "already issued" set would grow for the life of the JVM with ports released long ago.

`get()` is now `synchronized`: it is static and the check-and-set is a read-modify-write. JUnit parallel execution is not enabled today, but synchronising costs nothing and enabling it later would otherwise reintroduce a race silently.

**Two signatures.** `get()` probes TCP; `get(Protocol)` takes `TCP`, `UDP` or `BOTH`. TCP and UDP port spaces are independent, so a port free for TCP may already be taken on UDP — a leaked MQTT-SN receiver thread from an earlier test, say.

**Three copies made line-for-line identical.** The file exists in `hivemq-edge` (twice — core and the OPC-UA module) and in `hivemq-edge-test`. They had drifted: only the test-repo copy probed UDP, so the one improvement anybody made reached a third of the call sites. Removing the in-repo duplicate needs a build change (test fixtures) and is deliberately left to a separate change.

## Verification

**The arithmetic simulated directly**, not read off the code:

| check | result |
| -- | -- |
| same number forced twice in a row | second draw **rejected** — the observed failure, closed |
| first-ever call against a zero-filled array | **succeeds** — the off-by-one that would block every bucket is avoided |
| bucket after the window elapses | **reissued** |

Both repositories compile. `BasicBridgeIT`, which exposed the original collision, passes all five methods. Spotless applied; the three files remain byte-identical afterwards.

## Acceptance Criteria

- [x] Cannot return the same port twice within 32 requests
- [x] The array is consulted before any socket is opened
- [x] `get()` is synchronized
- [x] Memory is fixed; no reset needed between tests or classes
- [x] The residual cross-process case is documented in the class comment
- [ ] `BasicBridgeIT` stops appearing in the flaky list — needs time to confirm

## Non-Goals

* **Cross-JVM or cross-process reservation.** The filter cannot see other forks or other processes on the machine. That is the *original* residual risk with both protections intact, and no observed failure has ever matched that shape. The class comment says so, so nobody reads this as a guarantee.
* **Bind-to-zero.** [EDG-931](https://linear.app/hivemq/issue/EDG-931/stop-six-core-api-test-classes-sharing-a-hardcoded-http-port) established why it does not apply: ~145 of ~150 sites need the number *before* anything binds.
* **Removing the in-repo duplicate.** Needs a build change; separate.

### DIFF
--- a/hivemq-edge/src/test/java/util/RandomPortGenerator.java
+++ b/hivemq-edge/src/test/java/util/RandomPortGenerator.java
@@ -25,30 +25,40 @@ import java.net.ServerSocket;
  * <h2>Before you blame this class</h2>
  * <b>This does not cause port collisions, and a test that fails to bind is almost never this class's
  * fault.</b> That claim has been made many times over the years and was false every time but one -- and that
- * one exception is fixed, below. Read the following before repeating it.
+ * one exception is fixed, below.
  * <p>
- * The usual accusation is the window between proving a port free and the caller actually binding it: the
- * socket is closed and the number returned, so in principle someone else could take it in between. In
- * principle. For that to actually happen, <b>two</b> independent things must go wrong:
+ * The accusation is always the same: the window between proving a port free and the caller actually binding
+ * it. The socket is closed and the number returned, so in principle someone else could take it in between.
+ * In principle. For that to actually happen, <b>two</b> independent things must go wrong:
  * <ol>
- *   <li>another acquisition must fall inside that window, <b>and</b>
- *   <li>it must draw the same number out of {@value #RANGE} candidates.
+ *   <li>another acquisition must fall inside that window -- microseconds to a few tens of milliseconds,
+ *       against a suite that runs for minutes, <b>and</b>
+ *   <li>out of {@value #RANGE} candidates it must draw <i>exactly</i> the number we are holding.
  * </ol>
- * Both are needed, and both are unlikely, which is why ~150 call sites coexist without trouble. The range also
- * stays clear of the ports the operating system hands out on its own (see {@link #LOWEST} / {@link #HIGHEST}),
- * so the most plausible outside source is excluded by construction.
+ * Neither is likely on its own and they are independent, so the product is remote. The range is also kept
+ * clear of the ports the kernel hands out for itself (see {@link #LOWEST} / {@link #HIGHEST}), which excludes
+ * the most plausible outside source by construction. That is why ~150 call sites coexist without trouble.
  * <p>
- * <b>So before claiming this class caused a failure, have real evidence: the port number in the log, and what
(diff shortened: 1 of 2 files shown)


mschoenert-hivemq:
**Paired with hivemq/hivemq-edge-test#450** — the same file lives in both repositories and both must land, or the two copies drift apart again.

PR #1741, opened 2026-08-30
EDG-968: Let callers wait until a config change has been applied

mschoenert-hivemq:
## Description (the why)

A configuration change is **acknowledged before it is enacted**. The REST call records the change, queues the work of reconciling the running adapters with it, and returns `200 OK`. Nothing in the API distinguishes "configured" from "in effect".

A caller that changes something and immediately acts on it therefore races that reconciliation, and the race is silent: a write published before the adapter's writing service has rebuilt its schema and data policy is dropped with **nothing logged**. Awaiting `STARTED`/`CONNECTED` does not close the gap — `ProtocolAdapterWrapper.startAdapter` sets `RuntimeStatus.STARTED` well before it calls `startWriting()`.

Today the only remedy available to a caller is to sleep and hope. The EtherNet/IP integration fixture carries two such sleeps (3s after every adapter delete, 2s after every mapping change); removing them is the companion PR, and this is what makes that possible.

Linear: [EDG-968](https://linear.app/hivemq/issue/EDG-968/wait-for-a-configuration-change-to-be-applied-instead-of-sleeping)

## What changes

One new method on `ProtocolAdapterManager`. Nothing existing is modified — this is purely additive.

**The wait is an empty task, not a polled flag.** Refreshes run on a single-threaded executor with exactly one other submitter (the refresh itself), so a task placed on the end of the queue cannot run until every refresh ahead of it has finished. No progress counter to keep in step, and no window in which a caller observes "not busy" before its own work has been picked up.

This is why the existing `isBusy()` is not the right tool for it: that reports a counter, so `false` means either "finished" or "not yet submitted", and a caller that has just made a change cannot tell those apart.

Alongside it, a flag recording whether the last completed refresh reconciled every adapter, so the wait can report success as well as completion. It describes whichever change finished last rather than the caller's specifically — configuration changes arrive only through the API, one at a time, so a caller waiting here is waiting on its own work. That assumption is written into the Javadoc, because a future caller that cannot rely on it would need something finer.

**No plumbing.** `Injector` already exposes `protocolAdapterManager()`, so nothing else needed widening — an earlier draft threaded the manager through `Services` and that turned out to be unnecessary.

## Testing

Exercised by the companion PR in `hivemq-edge-test`, which replaces both fixed sleeps in the EtherNet/IP fixture with this call. That class goes **81.8s → 36.2s** (two runs: 36.2s, 36.0s), all nine tests green.

The write test's own trace is the direct evidence that the wait does what it claims: its first write now round-trips in **104ms with no retries**, where it previously had to retry through the window in which writes were silently dropped.

- [ ] A caller can block until every configuration change requested so far has been enacted
- [ ] The wait reports whether the last completed change succeeded
- [ ] No existing signature or behaviour changes

## Non-Goals

- **Distinguishing one caller's change from another's.** Not needed while changes arrive one at a time through the API; documented as the assumption it is.
- **Replacing `isBusy()`.** It has other callers and a different meaning.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
+++ b/hivemq-edge/src/main/java/com/hivemq/protocols/ProtocolAdapterManager.java
@@ -67,6 +67,7 @@ import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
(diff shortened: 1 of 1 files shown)


mschoenert-hivemq:
Companion test-side PR: hivemq/hivemq-edge-test#455 — replaces the two fixed settle sleeps in the EtherNet/IP fixture with this wait (81.8s → 36.2s). Same branch name, so the composite build pairs them.

Supersedes #1740, which GitHub closed when the branch was renamed. Same commit, no content change.

PR #1752, opened 2026-09-10
EDG-996: Extend InternalTopicFilterSubscriber with async and contextual processors

mschoenert-hivemq:
## Description (the why)

Message handling in the broker and in Edge is sequential, because that is how MQTT's ordering guarantees are met. Sequential without locks and without overhead is achieved with dedicated threads, one performing every operation for a whole bucket of queues. It follows that a component's processor may not block that thread for any extended period, since every queue in the bucket waits behind it.

`InternalTopicFilterSubscriber` gave such work nowhere to go. Its processor returned `void` and said "finished" by returning, so the only way to stay sequential was to do the work first, on that thread -- which a device write, or any call that may block, must not do. Edge consequently grew several near-copies of the queue-draining pipeline, each restating the same subtle discipline.

Linear: [EDG-996](https://linear.app/hivemq/issue/EDG-996/extend-internaltopicfiltersubscriber-async-processor-routed-delivery)

- [x] A processor may answer a **future**; the next message is passed on only when it completes. Handling stays sequential either way -- at most one message in flight per component -- so the asynchronous form does not create that bound, it lets a consumer keep it without occupying the thread meanwhile.
- [x] Push is unaffected: `withProcessor` wraps into the async form, catching so a throw becomes a failed future.
- [x] **Contextual delivery**: one subscriber can serve several topic filters and be told which matched, keyed by MQTT subscription identifier, receiving back its own objects rather than integers.
- [x] **Four processor forms, not three.** How a consumer signals completion and what it is told about a message are independent choices:

  |                     | told only the message | told what it matched    |
  | ------------------- | --------------------- | ----------------------- |
  | **done on return**  | `Processor`           | `ContextProcessor`      |
  | **done on a future**| `AsyncProcessor`      | `AsyncContextProcessor` |

- [x] **Which kind of subscriber follows from the processor**, not from a separate flag -- a flag beside it could disagree with it, the more easily because filters may be set far from where the subscriber is built.
- [x] Both filter families work on the builder **and** on the live subscriber, so a context subscriber is reconfigurable at runtime.
- [x] `withQoS`, per subscriber, defaulting to `AT_LEAST_ONCE`.
- [x] `withQueueLimit` and `withQueueOverflow`, both optional and defaulting to the broker-wide value, replacing hardcoded client-id prefix checks in two other classes.
- [x] Full Edge suite green (4,830 tests). Key behaviours verified by mutation -- the code was broken deliberately and the expected test observed to fail.

## Notes for the reviewer

**Two bugs found and fixed while reviewing this work, both silent:**

1. **A processor could be called with the subscriber's monitor held.** The in-memory single writer -- which Edge uses whenever file persistence is absent -- runs a submitted task *inline on the calling thread*, so a poll submitted inside a synchronized verb ran the component's processor inside the lock. Two consequences: a processor that blocked waiting on a thread wanting that lock would deadlock, and `consume()` set its `consuming` flag *after* submitting the drain, so a `pause()` called from inside a processor hit the "not consuming" guard and was silently discarded. Now the state change happens under the lock and the drain is submitted after releasing it; `start()` is no longer synchronized for the same reason.

2. **The builder's absolute filter verbs silently discarded earlier work.** On the builder the filter set starts empty, so a second `withTopicFilter(...)` could only be throwing away what the caller had just declared. It now throws, naming the filters already there and the `add...` verb the caller meant. On the *live subscriber* the same verb keeps its replacing sense, where a whole new set is an ordinary wish.

**Contract now stated explicitly:** a processor may add or remove topic filters, synchronously or from a future's completion, and such a call will never block or deadlock. *Which* messages the change affects is deliberately undefined.

**Deliberately untested:** the guard that throws when a subscriber holds more than half the subscription-identifier range. Provoking it needs ~134 million filters, so a test would have to fake the condition through internals.

## Non-Goals

- The queueless subscriber variant -- [EDG-997](https://linear.app/hivemq/issue/EDG-997/add-a-queueless-internaltopicfiltersubscriber-variant).
- Timeouts and repeated-timeout escalation.
- Folding `QueueConsumer` onto this class, though it is a near-copy of the same pipeline.
- Deleting the sampler's two prefix special-cases, which this makes possible but which wait until the sampler is itself an `InternalTopicFilterSubscriber`.

Edge Lore page updated to match, in a separate PR.

### DIFF
--- a/hivemq-edge/src/main/java/com/hivemq/mqtt/services/PublishDistributorImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/mqtt/services/PublishDistributorImpl.java
@@ -43,6 +43,8 @@ import com.hivemq.mqtt.message.publish.PUBLISH;
 import com.hivemq.mqtt.message.publish.PUBLISHFactory;
 import com.hivemq.mqtt.topic.SubscriberWithIdentifiers;
 import com.hivemq.persistence.clientqueue.ClientQueuePersistence;
+import com.hivemq.persistence.clientqueue.InternalTopicFilterSubscriber;
+import com.hivemq.persistence.clientqueue.InternalTopicFilterSubscriberFactory;
 import com.hivemq.persistence.clientqueue.QueuePolicy;
 import com.hivemq.persistence.clientsession.ClientSession;
 import com.hivemq.persistence.clientsession.ClientSessionPersistence;
@@ -81,6 +83,11 @@ public class PublishDistributorImpl implements PublishDistributor {
     @NotNull
     private final Lazy<ClientSessionPersistence> clientSessionPersistence;
 
+    // Lazy, like the session persistence beside it: this is only consulted for an internal subscriber, and
+    // resolving the factory eagerly would drag the topic tree and queue persistence into this object's
+    // construction for a path most messages never take.
+    private final Lazy<InternalTopicFilterSubscriberFactory> subscriberFactory;
+
     @NotNull
     private final MqttConfigurationService mqttConfigurationService;
 
@@ -105,11 +112,13 @@ public class PublishDistributorImpl implements PublishDistributor {
     public PublishDistributorImpl(
             final @NotNull ClientQueuePersistence clientQueuePersistence,
             final @NotNull Lazy<ClientSessionPersistence> clientSessionPersistence,
+            final @NotNull Lazy<InternalTopicFilterSubscriberFactory> subscriberFactory,
             final @NotNull ConfigurationService configurationService,
             final @NotNull Lazy<SamplingService> samplingService,
             final @NotNull Lazy<MessageForwarder> messageForwarder) {
         this.clientQueuePersistence = clientQueuePersistence;
         this.clientSessionPersistence = clientSessionPersistence;
+        this.subscriberFactory = subscriberFactory;
         this.mqttConfigurationService = configurationService.mqttConfiguration();
         this.bridgeConfiguration = configurationService.bridgeExtractor();
         this.samplingService = samplingService;
@@ -237,7 +246,16 @@ public class PublishDistributorImpl implements PublishDistributor {
         final boolean qos0Message = Math.min(subscriptionQos, publish.getQoS().getQosNumber()) == 0;
         Long queueLimit = null;
 
-        if (!client.startsWith(INTERNAL_SUBSCRIBER_PREFIX)) {
+        if (client.startsWith(INTERNAL_SUBSCRIBER_PREFIX)) {
+            // An internal subscriber has no client session, so there is no session-scoped limit to read; it
+            // declares its own instead, at the builder. Null here means it declared none, and queuePublish
+            // then falls back to the broker-wide maxQueuedMessages -- the same value it got before this.
+            final InternalTopicFilterSubscriber subscriber =
+                    subscriberFactory.get().getSubscriber(client);
+            if (subscriber != null) {
+                queueLimit = subscriber.queueLimit();
+            }
+        } else {
             final ClientSession clientSession = clientSessionPersistence.get().getSession(client, false);
             final boolean clientConnected = clientSession != null && clientSession.isConnected();
 
--- a/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/ClientQueuePersistenceImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/ClientQueuePersistenceImpl.java
@@ -73,6 +73,10 @@ public class ClientQueuePersistenceImpl extends AbstractPersistence implements C
     private final @NotNull LocalTopicTree topicTree;
     private final @NotNull ConnectionPersistence connectionPersistence;
     private final @NotNull Lazy<PublishPollService> publishPollService;
+
+    // Lazy for the same reason as the poll service above: the factory depends on this persistence, so an
+    // eager reference would be a construction cycle. Only consulted for an internal subscriber's queue.
+    private final @NotNull Lazy<InternalTopicFilterSubscriberFactory> subscriberFactory;
     private final @NotNull MessageForwarder messageForwarder;
     private final @NotNull ShutdownHooks shutdownHooks;
     private final @NotNull Map<String, PublishAvailableCallback> queueidCallbackMap;
@@ -87,8 +91,10 @@ public class ClientQueuePersistenceImpl extends AbstractPersistence implements C
             final @NotNull LocalTopicTree topicTree,
             final @NotNull ConnectionPersistence connectionPersistence,
             final @NotNull Lazy<PublishPollService> publishPollService,
+            final @NotNull Lazy<InternalTopicFilterSubscriberFactory> subscriberFactory,
             final @NotNull MessageForwarder messageForwarder,
             final @NotNull ShutdownHooks shutdownHooks) {
+        this.subscriberFactory = subscriberFactory;
         this.shutdownHooks = shutdownHooks;
         this.localPersistence = localPersistence;
         this.mqttConfigurationService = mqttConfigurationService;
@@ -100,6 +106,41 @@ public class ClientQueuePersistenceImpl extends AbstractPersistence implements C
         singleWriter = singleWriterService.getQueuedMessagesQueue();
         this.messageForwarder = messageForwarder;
         this.queueidCallbackMap = new ConcurrentHashMap<>();
+    }
+
+    /**
+     * What to drop when this queue is full.
+     * <p>
+     * <b>The producer's {@link QueuePolicy} decides first.</b> {@code SAMPLE_RING} is sampling saying it wants
+     * a ring buffer of the most recent messages, and nothing overrides that.
+     * <p>
+     * Within {@code DEFAULT}, an internal subscriber may declare its own strategy at its builder -- which is
+     * where a consumer says everything else about itself. Declaring none leaves the broker-wide setting, which
+     * is what every queue got before this existed.
+     * <p>
+     * <b>The queue id is deliberately not consulted for the policy.</b> Reading it off the id was the EDG-882
+     * F-05 defect: a share name is chosen by a client, so an ordinary subscription to
+     * {@code $share/$SAMPLER::customer/alerts} had its messages discarded under a policy meant for
+     * diagnostics. The id is used here only to look a subscriber up, which is an identity question rather than
+     * a policy one.
+     * <p>
+     * The factory is resolved lazily, and only for a queue whose id says it could be an internal subscriber:
+     * the factory depends on this persistence, so holding it eagerly would be a construction cycle.
+     */
+    private @NotNull MqttConfigurationService.QueuedMessagesStrategy overflowStrategyFor(
+            final @NotNull String queueId, final @NotNull QueuePolicy policy) {
+
+        if (policy == QueuePolicy.SAMPLE_RING) {
+            return MqttConfigurationService.QueuedMessagesStrategy.DISCARD_OLDEST;
+        }
+        if (queueId.startsWith(InternalTopicFilterSubscriber.INTERNAL_SUBSCRIBER_PREFIX)) {
+            final InternalTopicFilterSubscriber subscriber =
+                    subscriberFactory.get().getSubscriber(queueId);
+            if (subscriber != null && subscriber.queueOverflow() != null) {
+                return subscriber.queueOverflow();
+            }
+        }
+        return mqttConfigurationService.getQueuedMessagesStrategy();
     }
 
     @Override
@@ -119,14 +160,10 @@ public class ClientQueuePersistenceImpl extends AbstractPersistence implements C
         }
 
         return singleWriter.submit(queueId, (bucketIndex) -> {
-            // What to do when the queue is full comes from the producer (see QueuePolicy). It used to be
-            // read off the queue ID -- a share name a client chooses -- so an ordinary subscription to
-            // $share/$SAMPLER::customer/alerts had its own messages discarded under a policy meant for
-            // diagnostics (EDG-882 F-05).
-            final MqttConfigurationService.QueuedMessagesStrategy queuedMessagesStrategy =
-                    policy == QueuePolicy.SAMPLE_RING
-                            ? MqttConfigurationService.QueuedMessagesStrategy.DISCARD_OLDEST
-                            : mqttConfigurationService.getQueuedMessagesStrategy();
+            // What to do when the queue is full comes from the producer's QueuePolicy, and within DEFAULT from
+            // whatever an internal subscriber declared -- see overflowStrategyFor. Whether the count bound also
+            // applies to QoS 0 remains the policy's alone: it is sampling's ring-buffer semantics, not
+            // something a consumer picks.
             final boolean applyMaxToQos0 = policy == QueuePolicy.SAMPLE_RING;
 
             localPersistence.add(
@@ -134,7 +171,7 @@ public class ClientQueuePersistenceImpl extends AbstractPersistence implements C
                     shared,
                     publish,
                     queueLimit,
-                    queuedMessagesStrategy,
+                    overflowStrategyFor(queueId, policy),
                     retained,
                     applyMaxToQos0,
                     bucketIndex);
@@ -168,14 +205,8 @@ public class ClientQueuePersistenceImpl extends AbstractPersistence implements C
 
         return singleWriter.submit(queueId, (bucketIndex) -> {
             final boolean queueWasEmpty = localPersistence.size(queueId, shared, bucketIndex) == 0;
-            // What to do when the queue is full comes from the producer (see QueuePolicy). It used to be
-            // read off the queue ID -- a share name a client chooses -- so an ordinary subscription to
-            // $share/$SAMPLER::customer/alerts had its own messages discarded under a policy meant for
-            // diagnostics (EDG-882 F-05).
-            final MqttConfigurationService.QueuedMessagesStrategy queuedMessagesStrategy =
-                    policy == QueuePolicy.SAMPLE_RING
-                            ? MqttConfigurationService.QueuedMessagesStrategy.DISCARD_OLDEST
-                            : mqttConfigurationService.getQueuedMessagesStrategy();
+            // As in the single-message add above: the policy decides, an internal subscriber's declaration
+            // refines it within DEFAULT, and applying the bound to QoS 0 stays the policy's own business.
             final boolean applyMaxToQos0 = policy == QueuePolicy.SAMPLE_RING;
 
             localPersistence.add(
@@ -183,7 +214,7 @@ public class ClientQueuePersistenceImpl extends AbstractPersistence implements C
                     shared,
                     publishes,
                     queueLimit,
-                    queuedMessagesStrategy,
+                    overflowStrategyFor(queueId, policy),
                     retained,
                     applyMaxToQos0,
                     bucketIndex);
--- a/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/ClientQueuePersistenceImpl.java
+++ b/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/ClientQueuePersistenceImpl.java
@@ -118,22 +118,31 @@ public class ClientQueuePersistenceImpl extends AbstractPersistence implements C
      * where a consumer says everything else about itself. Declaring none leaves the broker-wide setting, which
      * is what every queue got before this existed.
      * <p>
-     * <b>The queue id is deliberately not consulted for the policy.</b> Reading it off the id was the EDG-882
-     * F-05 defect: a share name is chosen by a client, so an ordinary subscription to
-     * {@code $share/$SAMPLER::customer/alerts} had its messages discarded under a policy meant for
-     * diagnostics. The id is used here only to look a subscriber up, which is an identity question rather than
-     * a policy one.
+     * <b>ONLY FOR A NON-SHARED QUEUE, and {@code shared} is the only thing that can tell.</b> A shared queue's
+     * id is built from a share name, and a share name is the client's to choose -- so a subscription to
+     * {@code $share/$INTERNAL::<a live subscriber's name>/} yields a queue id that, once the {@code $share/}
+     * prefix is stripped, is indistinguishable from that subscriber's own. Asking the factory does not help:
+     * an internal subscriber's id carries no topic suffix, so nothing recoverable from the string separates
+     * "the subscriber called X" from "a client's share group called X". The flag the caller already holds is
+     * the answer, and without it a client could take an internal subscriber's overflow strategy for its own
+     * queue.
+     * <p>
+     * This is the third time the same shape has bitten: reading a queue's policy off its id gave a legitimate
+     * client a sampler's ten-message ring (EDG-882 F-05) and, a round later, a bridge's queue limit and a
+     * rewritten QoS (EDG-882 QA round 2). Both were fixed by asking the owning service instead. That remedy
+     * does not transfer here -- see above -- so this branch is gated on {@code shared} rather than made
+     * name-independent.
      * <p>
      * The factory is resolved lazily, and only for a queue whose id says it could be an internal subscriber:
      * the factory depends on this persistence, so holding it eagerly would be a construction cycle.
      */
     private @NotNull MqttConfigurationService.QueuedMessagesStrategy overflowStrategyFor(
-            final @NotNull String queueId, final @NotNull QueuePolicy policy) {
+            final @NotNull String queueId, final boolean shared, final @NotNull QueuePolicy policy) {
 
         if (policy == QueuePolicy.SAMPLE_RING) {
             return MqttConfigurationService.QueuedMessagesStrategy.DISCARD_OLDEST;
         }
-        if (queueId.startsWith(InternalTopicFilterSubscriber.INTERNAL_SUBSCRIBER_PREFIX)) {
+        if (!shared && queueId.startsWith(InternalTopicFilterSubscriber.INTERNAL_SUBSCRIBER_PREFIX)) {
             final InternalTopicFilterSubscriber subscriber =
                     subscriberFactory.get().getSubscriber(queueId);
             if (subscriber != null && subscriber.queueOverflow() != null) {
@@ -171,7 +180,7 @@ public class ClientQueuePersistenceImpl extends AbstractPersistence implements C
                     shared,
                     publish,
                     queueLimit,
-                    overflowStrategyFor(queueId, policy),
+                    overflowStrategyFor(queueId, shared, policy),
                     retained,
                     applyMaxToQos0,
                     bucketIndex);
@@ -214,7 +223,7 @@ public class ClientQueuePersistenceImpl extends AbstractPersistence implements C
                     shared,
                     publishes,
                     queueLimit,
-                    overflowStrategyFor(queueId, policy),
+                    overflowStrategyFor(queueId, shared, policy),
                     retained,
                     applyMaxToQos0,
                     bucketIndex);
--- a/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java
+++ b/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java
@@ -102,9 +102,14 @@ public final class InternalTopicFilterSubscriber {
 
     public static final @NotNull String INTERNAL_SUBSCRIBER_PREFIX = "$INTERNAL::";
 
-    // SHARED_IN_FLIGHT_MARKER acts as a boolean inflight flag -- not a real wire packet ID, since
-    // messages never go to an MQTT client. removeShared() uses uniqueId, not the packet ID, so
-    // the value here does not matter.
+    // SHARED_IN_FLIGHT_MARKER acts as a boolean inflight flag -- not a real wire packet ID, since messages
+    // never go to an MQTT client. Reading a message above QoS 0 stamps it with the id supplied here and leaves
+    // it in the queue; a later read skips anything stamped. So all this value has to be is non-zero.
+    //
+    // IT IS ALSO WHAT removeMessage() DELETES BY, so the same value identifies "the stamped message" rather
+    // than a particular one. That is unambiguous only because a subscriber holds exactly one message at a
+    // time -- see inFlight. Two in flight at once would both be stamped with this, and the removal would
+    // delete whichever the queue happened to hold first.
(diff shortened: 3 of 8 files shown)


caoccao (changes requested):
I found **four PR issues and one pre-existing defect**.

1. **[P1] Async processing does not enforce one message in flight.**  
   [[InternalTopicFilterSubscriber.java:1419–1428](https://github.com/hivemq/hivemq-edge/blob/a5bbd3b240949b7284bb2e4839b9e435653a9e9c/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1419-L1428)](https://github.com/hivemq/hivemq-edge/blob/a5bbd3b240949b7284bb2e4839b9e435653a9e9c/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1419-L1428)  
   Completion chaining controls only one source of polls. Reading a QoS 0 message removes it immediately, so the next arrival triggers another poll while the first processor’s future remains pending. Calling `pause().consume()` also bypasses the bound for QoS 1. Both reproductions delivered `b` before `a` completed. All poll entry points need a shared in-flight guard.

2. **[P1] A stopped subscriber can consume its replacement’s messages.**  
   [[InternalTopicFilterSubscriber.java:1438–1445](https://github.com/hivemq/hivemq-edge/blob/a5bbd3b240949b7284bb2e4839b9e435653a9e9c/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1438-L1445)](https://github.com/hivemq/hivemq-edge/blob/a5bbd3b240949b7284bb2e4839b9e435653a9e9c/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1438-L1445)  
   An outstanding future survives `stop()`. Its completion unconditionally submits another poll, even after the subscriber is deallocated and its ID reused. I reproduced the old processor consuming a command queued for its replacement. Completion callbacks and scheduled polls must check that their subscriber still owns the queue.

3. **[P2] Rebuilding a contextual subscriber can route persisted messages to the wrong destination.**  
   [[InternalTopicFilterSubscriber.java:290–294](https://github.com/hivemq/hivemq-edge/blob/a5bbd3b240949b7284bb2e4839b9e435653a9e9c/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L290-L294)](https://github.com/hivemq/hivemq-edge/blob/a5bbd3b240949b7284bb2e4839b9e435653a9e9c/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L290-L294)  
   Identifiers restart at `1` and depend on filter insertion order, while persisted messages retain their previous identifiers. Rebuilding the same filter set in a different order made a queued `commands/a` message resolve to device B. Runtime filter churn creates similar mismatches after restart. The mapping must survive with the queue, or stale identifiers must be prevented from resolving to unrelated filters.

4. **[P2] Shared queues can inherit an unrelated internal subscriber’s overflow policy.**  
   [[ClientQueuePersistenceImpl.java:130–140](https://github.com/hivemq/hivemq-edge/blob/a5bbd3b240949b7284bb2e4839b9e435653a9e9c/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/ClientQueuePersistenceImpl.java#L130-L140)](https://github.com/hivemq/hivemq-edge/blob/a5bbd3b240949b7284bb2e4839b9e435653a9e9c/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/ClientQueuePersistenceImpl.java#L130-L140)  
   `overflowStrategyFor()` ignores whether the queue is shared. An internal instance ID containing `/` can match a legal shared queue’s string ID. I reproduced a shared queue discarding its oldest message under an internal subscriber’s `DISCARD_OLDEST`, despite the broker specifying `DISCARD`. Apply subscriber overrides only when `shared == false`.

5. **[P1, pre-existing] Processed QoS 1/2 messages are never removed from their queue.**  
   [[InternalTopicFilterSubscriber.java:1448–1451](https://github.com/hivemq/hivemq-edge/blob/a5bbd3b240949b7284bb2e4839b9e435653a9e9c/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1448-L1451)](https://github.com/hivemq/hivemq-edge/blob/a5bbd3b240949b7284bb2e4839b9e435653a9e9c/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1448-L1451)  
   Messages are read from non-shared storage, but `removeShared()` searches shared storage. A successfully processed QoS 1 message therefore remains queued and marked in flight. Eventually the queue fills and rejects further messages. I confirmed this code also exists on the base revision.

**Validation:** 91 existing targeted tests passed. Six focused reproducers failed at the expected assertions; the restart case simulates a restored queue.

The [[CI dependency scan](https://github.com/hivemq/hivemq-edge/actions/runs/34487480726)](https://github.com/hivemq/hivemq-edge/actions/runs/34487480726) also reports a FreeMarker 2.3.34 vulnerability; this diff does not change dependencies.

mschoenert-hivemq:
Thank you — I verified all five against the code, and every one is correct. **Finding 1 is fixed in `0c2c4264`;** the rest are outstanding and I have not touched them yet.

## 1 — one message in flight: fixed

You are right that the bound did not hold, and the reproduction matches what I found. One correction to the prescription, though: **the fix is not a guard at each entry point.** All four already funnel through `submitPoll`, and that is the only thing that ever reads the queue — so four guards would be four copies of one rule, waiting to rot when a fifth trigger is added.

The defect is that the funnel did not express the condition it was supposed to enforce. It unconditionally scheduled a read, when what it means is *read the next message when none is in flight*. The waiting lived in the completion chain instead: one path knew the rule and the funnel it flowed through did not.

What is now in place:

- `inFlight`, true from the moment a message is claimed until its future settles.
- `pollUnlessInFlight()` — used by every trigger meaning *a message may be available*: the publish-available callback, `consume()`, the error path. Gives up if a message is in flight; the completion will poll again, so the trigger is redundant rather than lost.
- `pollAfterInFlight()` — used by the one trigger that also means *the message I was working on is finished*. Clears the flag, then polls.
- Both are **submitted as tasks**, so the flag is only ever touched inside the single-writer serialisation. A plain field, no atomic: the completion submits rather than clearing inline, which would put the write on the consumer's thread for no gain.

Two placement details, both proved by mutation:

- **The claim is taken before the read**, because `readNew` submits a task of its own — claiming afterwards leaves a gap in which another trigger polls and hands out a second message.
- **An empty read therefore releases the claim itself.** No message means no completion, so nothing else ever would, and the subscriber goes permanently deaf. Removing that line left the whole suite green, so it now has its own test.

Two new tests, each failing without the fix: one fires the publish-available callback while a message is in flight (your reproduction); one polls an empty queue and then wakes. Neither trigger was reachable from the previous tests, which is why 32 passing tests missed this.

Also renamed `processPublish` to `processMessage`, and corrected the comment claiming the chaining *was* the bound — it only ever covered one of the four paths in.

## 2, 3, 4 — confirmed, not yet addressed

- **2 (stopped subscriber consumes its replacement's messages):** confirmed. Every use-after-shutdown guard sits on the public verbs; none of the pipeline methods checks it.
- **3 (identifiers do not survive a restart):** confirmed. The counter is an in-memory field reset at construction, while queued messages are file-backed and keep their old identifiers. My "not reused until the counter wraps" reasoning never covered a restart, which resets rather than advances it.
- **4 (shared queues inherit an internal subscriber's overflow policy):** confirmed. `overflowStrategyFor` never consults the `shared` flag, and a shared queue's id is a share name a client chooses. This is the same class of defect EDG-882 fixed by removing the queue-id heuristic, which makes it worth taking seriously.

## 5 — pre-existing, and I would like to keep it separate

Confirmed, and confirmed identical on the base revision: messages are read from non-shared storage (`readNew(clientId, false, …)`) but acknowledged against shared storage (`removeShared`). Processed QoS 1/2 messages are never removed, and the queue fills.

It makes the subscriber unusable for QoS 1 in practice, so it deserves its own issue rather than being folded in here — happy to raise it if you agree.

mschoenert-hivemq:
**Finding 2 is now fixed too, in `68ebe649`** — and it turned out to be the same defect as a pause not taking effect, so one clause covers both.

The subscriber already had a `consuming` flag: set when it starts draining, cleared when it pauses. It was written by `consume()`/`pause()` and **read by nothing on the poll path**. So pausing removed the wake-up callback but did not stop a drain already under way, and did not stop an outstanding message's completion from reading the next one. The state recorded intent; nothing enforced it.

`pollUnlessBusy()` now reads it:

```java
if (inFlight || !consuming) {
    return;
}
```

That subsumes your finding rather than needing a check of its own. `stop()` pauses before deallocating, so a deallocated subscriber is by definition one that is no longer consuming — and the completion of its outstanding future therefore declines to poll. No lifetime test, no second concept.

**It also makes `pause()` mean what the documentation always claimed.** I previously had to document that pausing does not stop a drain in progress; that was true only because the flag was unread. It now does, so I inverted the assertion in the test that pinned the old behaviour.

Two tests, both failing without the clause:

- `aCompletionArrivingAfterStopDoesNotReadFromTheQueue` — your case: a message queued under the freed client id, and the dead subscriber's completion arriving afterwards. Asserts it processes nothing *and* does not read at all, since a replacement may already own that queue.
- `aProcessorThatPausesDuringTheFirstDrainActuallyPauses` — now asserts the second message is not delivered.

Full suite green: 5,221 tests, 0 failures.

Still outstanding, both confirmed: **3** (subscription identifiers are an in-memory counter reset at construction, while queued messages are file-backed and keep their old identifiers) and **4** (`overflowStrategyFor` never consults the `shared` flag). And **5** remains the pre-existing read-non-shared/acknowledge-shared mismatch, which I would still rather raise separately than fold in here.

mschoenert-hivemq:
**Finding 5 is fixed too, in `3e3b7cad`** — as its own commit, since it is a pre-existing bug rather than part of this feature. I had proposed splitting it out to a separate issue; on reflection it is two lines and the branch already touches this exact method, so carrying it here is less ceremony than a second pull request for the same file.

The deletion now calls `remove(clientId, packetId)` against the client-queue map instead of `removeShared(clientId, uniqueId)` against the shared-subscription map.

Two things worth flagging that were not in your report:

**The marker constant is now part of the contract.** `remove` keys on the *packet identifier*, and every message here is stamped with the same `SHARED_IN_FLIGHT_MARKER` rather than a real wire id — so the call deletes "the stamped message", not a named one. That is unambiguous only because a subscriber holds exactly one message at a time, which is the `inFlight` bound from finding 1. Before that bound existed this fix would have been unsound: two messages in flight would both carry the marker and the deletion would take whichever the queue held first. The constant's comment said its value "does not matter" — true under `removeShared`, false now — so it has been rewritten to say what it is used for.

**The tests could not have caught this, and still could not after the fix.** The fixture stubbed only `removeShared`, so nothing asserted *which* deletion was called; the entire suite passed both before and after the change, which is how I knew the coverage was absent rather than thin. Two tests now: one pins that a QoS 1 message is removed from the client queue and that the shared removal is never called; one pins that a QoS 0 message triggers no deletion at all, since reading it already removed it. Restoring the old call fails the first and only the first.

Full suite green: 5,223 tests, 0 failures.

---

**Where the review stands.** 1, 2 and 5 fixed. Two remain, both confirmed:

- **3** — subscription identifiers come from an in-memory counter reset at construction, while queued messages are file-backed and keep the identifiers they were given, so a restart can resolve a message to a different destination. No small fix: either the mapping persists with the queue, or stale identifiers must be recognised and dropped. Worth its own decision.
- **4** — I owe you a correction here. I confirmed the mechanism (the overflow lookup ignores the `shared` flag) but not the *consequence*, and the consequence does not follow: shared and non-shared queues are two separate maps, so the same name in each is two unrelated queues. Nothing another caller does can make this subscriber's queue shared, and a message queued with the flag set would land in a queue this subscriber never reads. So it is a missing precondition on a method rather than a reachable defect. Passing the flag through is still worth doing as tidiness — happy to, if you would like it on this branch — but I do not think it is a bug, and I should have established that before agreeing it was one.

mschoenert-hivemq:
**Finding 4 is fixed in `3d03aa74`, and I owe you an apology for the route it took.** I dismissed it twice before establishing you were right.

What I got wrong, in order:

1. I claimed the scenario was unreachable because shared and non-shared queues are separate stores. True, but beside the point — the defect is not that the subscriber's own queue is touched.
2. I then claimed `$INTERNAL::` names are blocked anyway. **They are blocked as CLIENT IDENTIFIERS, at connect time.** There is no equivalent check on a share group name at subscribe time, and the `$share/` prefix is stripped before the queue id is formed — so a subscription to `$share/$INTERNAL::<a live subscriber's name>/` yields a queue id starting with the internal prefix, reaching the test directly from a client-chosen string.
3. I proposed replacing the prefix test with a factory lookup, mirroring how the sampler and forwarder cases were fixed. That does not transfer: an internal subscriber's id carries **no topic suffix**, so the stripped share name can reconstruct it exactly and the factory cannot tell the two apart. The sibling cases work only because their own queue ids are themselves share-derived, so the collision stays inside one namespace.

So the flag the caller already holds is the only thing that can answer the question, and it is now threaded through. One test: the same id added with `shared = true` keeps the broker-wide strategy instead of the subscriber's, and fails without the flag.

Worth noting this is the **third** occurrence of the same shape — the two comments in `PublishDistributorImpl` describe the sampler and forwarder instances. Whatever else comes of this review, that pattern seems worth a rule rather than a third patch.

---

**Where the review stands.** 1, 2, 4 and 5 fixed. One remains:

**3** — subscription identifiers come from an in-memory counter reset at construction, while queued messages are file-backed and keep the identifiers they were given, so a restart can resolve a message to a different destination. There is no small fix: either the mapping persists with the queue, or stale identifiers must be recognised and dropped. I would rather that were its own issue than a hurried commit here, but I am happy either way — your call, and I will not resolve your review myself.

mschoenert-hivemq:
@caoccao — finding **3** is now addressed, in `bb03f99`. It was the one I left open longest, because you were right that it is real and I did not want to paper over it.

### What you found

Subscription identifiers restart at `1` and are handed out in filter insertion order, while a persisted message keeps the identifier it was stamped with. Rebuild the same filter set in a different order and a queued `commands/a` resolves to device B. Your reproduction was correct, and the runtime-churn variant you mention is the same defect reached by a different route.

### Why I did not take either remedy you offered

You suggested making the mapping survive with the queue, or preventing stale identifiers from resolving. Both work. But putting the two side by side made the actual problem visible: **the identifier is a cached answer, computed when a message is enqueued, interpreted by a table that lives in the subscriber.** Persisting the table fixes the symptom by making the cache durable; blocking stale resolution fixes it by detecting when the cache has gone bad. Neither questions why the answer was cached at that moment at all.

It did not need to be. A message carries its concrete topic. A subscriber holds its own filters. That is everything required to compute the match — so it can be computed when the message is *polled*, against the filters that exist then.

### What changed

`asAsync` no longer reads `message.getSubscriptionIdentifiers()`. It walks the subscriber's own `topicFilterContexts`, matches each filter against the message's topic, and unions the contexts behind those that match. Deleted along with the old path:

| Removed | Why it existed |
| --- | --- |
| `subscriptionIdByTopicFilter`, `topicFilterBySubscriptionId` | Both directions were needed — filter→id to register, id→filter to deliver. |
| `nextSubscriptionId` + forward-walking, wrap-around allocation | So a freed id was not handed straight back to a different filter. |
| `MAX_TOPIC_FILTERS` (half the id range) | So the forward search provably terminated rather than merely usually terminating. |
| `subscriptionIdFor` / `freeSubscriptionId` and their six call sites | Allocation had to happen before registration, in the right order, in every path. |
| A precondition on `addToTree` | It *read* an id without allocating one. A caller who forgot got a filter whose messages resolved to no context and were dropped silently. |

Net **−110 lines** in the production class. Your finding 3 is not fixed so much as **rendered unstatable**: there is no number, so nothing can go stale — across a restart, a `detach()`/`attach()`, or runtime churn.

It also closes a gap neither of us had named. Identifiers were per subscriber, so two subscribers both using `1` was always fine; the problem was only ever lifetime, never scope. Poll-time matching removes the lifetime question entirely.

### The matcher

New `private static boolean topicMatchesFilter(String filter, String topic)` — one left-to-right walk over both strings, no allocation, no splitting into levels. Two details carry the wildcards, and both are where hand-written MQTT matchers go wrong:

- **`+` is tested before the character comparison, and the loop advances on the filter alone.** This is what lets `+` match an *empty* level (`a/+` matches `a/`), where there is no character in the topic to compare against. My first draft had the tests the other way round and got this wrong on six cases.
- **A trailing `/#` is stripped up front and remembered**, so `a/b/#` matches its own parent `a/b`.

I deliberately did not reuse `PermissionTopicMatcherUtils`, which compares pre-split segment arrays. Precomputing the filter's split form only removes half of a cost the character walk never pays — the segment approach must also split the *incoming topic*, per message.

### Verification

- Full unit suite for the module green: **559 test classes**, `--rerun-tasks` so nothing came from cache.
- Subscriber tests **34 → 63**. The seven that asserted the numbering scheme are gone, replaced by tests of *delivery* — including one pinning that a message queued under a since-removed filter is **not** delivered to the destination that replaced it, which is the misrouting half of your finding stated as a behaviour.
- **24 fixture pairs** on the matching rules, cross-checked against an independent segment-based implementation written from the spec — exhaustively over every well-formed filter against every well-formed topic four levels deep (318,240 pairs), plus the spec's own published examples and multi-byte topics.
- Fixtures confirmed to bite by mutation: inverting the `+` ordering fails exactly the empty-level case; dropping the parent-match clause fails five tests.

Findings **1, 2, 4 and 5** were fixed earlier in this PR. All five are now addressed — happy to walk through any of it if you would rather a different approach on 3.

caoccao (changes requested):
The original reproducers now pass, but I found **four remaining issues** in commit `bb03f99348`.

1. **[P1] A late acknowledgement can delete a replacement subscriber’s message.**  
   [[InternalTopicFilterSubscriber.java:1494–1497](https://github.com/hivemq/hivemq-edge/blob/bb03f99348eaafe5fd605d2b47c1062f5fceb0f9/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1494-L1497)](https://github.com/hivemq/hivemq-edge/blob/bb03f99348eaafe5fd605d2b47c1062f5fceb0f9/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1494-L1497)  
   The new removal uses the same packet marker for every subscriber lifetime. After stopping subscriber A and starting replacement B with the same ID, completing A’s outstanding future deletes B’s in-flight message. I reproduced the queue size dropping from **1 to 0 while B’s future remained pending**. The acknowledgement needs an ownership/generation check within the serialized queue operation.

2. **[P1] A stop/read race still lets an old subscriber consume replacement messages.**  
   [[InternalTopicFilterSubscriber.java:1392–1405](https://github.com/hivemq/hivemq-edge/blob/bb03f99348eaafe5fd605d2b47c1062f5fceb0f9/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1392-L1405)](https://github.com/hivemq/hivemq-edge/blob/bb03f99348eaafe5fd605d2b47c1062f5fceb0f9/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1392-L1405)  
   The lifecycle check happens before the separately submitted read. If the thread is suspended after that check, `stop()` can clear the queue and release the ID before the read is submitted. A replacement can then enqueue a message that the old read consumes. A controlled interleaving reproduced the **old processor receiving the replacement’s message**. Ownership must be checked before the queued read removes or marks messages; checking only before polling is insufficient.

3. **[P1] A failed queue read permanently stalls the subscriber.**  
   [[InternalTopicFilterSubscriber.java:1393–1407](https://github.com/hivemq/hivemq-edge/blob/bb03f99348eaafe5fd605d2b47c1062f5fceb0f9/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1393-L1407)](https://github.com/hivemq/hivemq-edge/blob/bb03f99348eaafe5fd605d2b47c1062f5fceb0f9/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1393-L1407)  
   `inFlight` becomes `true` before `readNew()`, but `Futures.transform()` only runs the supplied function on success. Persistence exceptions complete the returned future exceptionally, bypassing the surrounding `catch` and leaving `inFlight` set forever. I injected one read failure; even after recovery and `pause().consume()`, queued messages were never processed. Handle exceptional future completion and release the claim.

4. **[P2] The new polling guard reads lifecycle state without safe publication.**  
   [[InternalTopicFilterSubscriber.java:1359–1363](https://github.com/hivemq/hivemq-edge/blob/bb03f99348eaafe5fd605d2b47c1062f5fceb0f9/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1359-L1363)](https://github.com/hivemq/hivemq-edge/blob/bb03f99348eaafe5fd605d2b47c1062f5fceb0f9/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1359-L1363)  
   `pause()` writes `consuming` under the subscriber’s monitor, but this reader neither acquires that monitor nor reads a `volatile` field. Unlike `inFlight`, `consuming` is modified outside the single writer. A completion running on another thread is therefore not guaranteed to observe the pause and may continue draining. Publish the flag safely or serialize lifecycle changes through the writer.

**Validation:** 119 existing targeted tests and all six original reproducers passed. Three new tests reproduced findings 1–3 using the real in-memory queue. Finding 4 follows from the synchronization analysis. The new topic matcher also passed comparison against the broker’s topic tree.

mschoenert-hivemq:
Thank you for the review.

This turned into a fairly significant rewrite of the class, so replying to each comment individually would not say much -- most of them no longer point at code that exists. But every one of your findings was addressed, and we believe fixed.

The one exception is `pause()`. It remains possible for a message to be processed after `pause()` has returned. `pause()` cannot wait for a message that is already in flight: the processing runs on the SingleWriter thread, and waiting for it from a caller that may itself be on that thread risks a deadlock. So `pause()` records the request and returns; the loop honours it at its next decision point. The Edge Lore article now states this explicitly -- what `pause()` does and does not promise, and the deadlock scenario that rules the stronger guarantee out.

I would appreciate another look.

caoccao (changes requested):
2 issues remain. They are in the comment of #1754

mschoenert-hivemq:
Two of the four issues you raised on #1754 were in this branch, and both are fixed here.

**Pending teardown permitted resurrection.** `deallocate()` now records the terminal state itself, before submitting the teardown, rather than leaving it to the teardown's own last line. The subscriber is dead to its caller from that instant, so a `start()` in the window cannot re-attach filters the teardown will not remove. This also replaced four booleans with a single six-valued `SubscriberState`; the four suggested sixteen combinations of which only six exist.

**A concurrent `consume()` defeating `pause()`.** The consume ask is now two commands. `CONSUME_DONT_START` is submitted from inside the monitor and only records the goal -- safe there, because recording cannot run consumer code. `CONSUME` is submitted after the monitor is free and does the acting. A competing `pause()` cannot take the monitor until `consume()` releases it, so its `PAUSE` is strictly later and lowers the goal; the `CONSUME` that follows finds the goal no longer active and does not start.

No test for the second one, deliberately: the fix removes the interleaving rather than handling it, so a test pinned to that sequence would assert that something impossible does not happen. A fuzzing harness over thread timings would be worth having, but not in this PR.

mschoenert-hivemq:
Thanks Sam -- you were right to look again, and the first one is on me.

**1. The CONSUME / CONSUME_DONT_START split was never actually implemented.** Sorry about that. The doc comment described the fix -- that the acting half only starts the loop if the goal has not changed in the meantime -- but the code mapped both halves to `ACTIVE`, so the condition the comment described did not exist. I wrote the explanation and not the code, which is the worst of both, because the comment made it look handled.

It is fixed properly now in `f96f8ec3`. `CONSUME` records *nothing at all*: `CONSUME_DONT_START` is sent from inside the monitor and records the intent, and the `CONSUME` that follows once the monitor is released only acts. So a `pause()` that runs whole in the gap between the two halves lowers the goal, and the acting half finds it lowered and does not start. To express that, `goalStateFor` is merged back into `ppfLoopCtrl` as an if-chain -- the rule is no longer one command to one goal, so a `switch` could not carry it: `CONSUME_DONT_START` records and returns, `CONSUME` records nothing, and `TERMINATED` absorbs without returning so that a deferred teardown is still acted on.

**2. Concurrent `stop()` on the queued class.** Fixed in `f4d1be74`, and you were right that the same defect existed on both sides -- it had been fixed on the queueless variant and not looked for here. `deallocate()` is now `synchronized`, so two threads cannot both pass the liveness check and have the loser walk into `detach()`, which throws on a dead subscriber.

The comment there is longer than the change because the asymmetry needs justifying: `start()` deliberately stays unsynchronized. It calls `consume()`, which must submit its acting command with the monitor free, since that submit can run the consumer's processor inline on the calling thread -- and this class promises never to call a processor with its monitor held. Two concurrent starts are harmless anyway: both verbs are idempotent and neither throws on a live subscriber, so the loser finds the work done rather than an exception. That is exactly what was *not* true of the release path.

**3. Publish recreating a queue after teardown.** This one we are deliberately not fixing here, because it is not something this PR introduces. Closing a connection never removes a queue -- only QoS 0 messages are dropped. The queue goes at *session expiry*, and at that moment an ordinary MQTT client has the identical window: the session is removed from the session persistence first, and the decision to enqueue is taken before the clear is submitted, so an enqueue already decided can land after it. The recreated non-shared queue is then never reclaimed, since the orphan sweep only covers shared queues. So there is no existing protection here to copy for internal subscribers.

Filed as EDG-1025 with a proposed general fix. The short version: a registry of retired queue ids, where writing an entry, reading it, and removing it are all serialised under the queue id in the client-queue single writer -- all three, because whatever removes an entry is itself a writer to the same fact. That surfaced a complication worth knowing about: there is no queue-*creation* call anywhere today, so entry-removal has nothing to attach to, and the obvious place to put it (session establishment) runs on a different single writer from queue destruction and so would not be ordered against it.

Both branches are green and rebased. PTAL.

caoccao (changes requested):
**One new regression affects both updated PRs**, reviewed at #1752 `90e87bf` and #1754 `f9931ec`.

**[P1] Completing work while paused permanently blocks resume and teardown.**  
When an async processor finishes after `pause()`, the new early return discards its `CONTINUE` command before changing `loopState` from `ACTIVE` to `PAUSED`. Later `consume()` and `stop()` encounter the active-loop guard and cannot proceed—even though no processing remains outstanding. Preserve the paused goal, but let the completion perform the state transition. [[InternalTopicFilterSubscriber.java:1223–1227](https://github.com/hivemq/hivemq-edge/blob/90e87bf44420cbdce9ab1681e2dcb96af5b27ad1/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1223-L1227)](https://github.com/hivemq/hivemq-edge/blob/90e87bf44420cbdce9ab1681e2dcb96af5b27ad1/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1223-L1227)

Two reproductions confirm this:

- Pause → complete processor → resume: the next message is never processed.
- Pause → complete processor → stop: the subscriber remains registered indefinitely.

Both tests **pass on the previous revision and fail on both updated heads**. The previous consume/pause race and concurrent-stop findings now pass. Existing targeted suites also pass: 118 tests on #1754 and 91 on #1752.

The known queue-recreation race still reproduces; the author has explicitly deferred it to [[EDG-1025](https://github.com/hivemq/hivemq-edge/pull/1752#issuecomment-5679266079)](https://github.com/hivemq/hivemq-edge/pull/1752#issuecomment-5679266079).

mschoenert-hivemq:
Thank you -- you were right, and this was a regression I introduced in the previous round. Fixed in `56bd32a0`.

**The bug, exactly as you described it.** The recording chain in `ppfLoopCtrl` turned "a progress report must not lift a goal a lifecycle verb has lowered" into an early `return`. That skips the guard and the act as well as the record -- and the report ending an iteration is the only thing that can move `loopState` off `ACTIVE`. So the loop was left for ever claiming an iteration that had finished, after which the one-iteration guard turned away every later `consume()` and `stop()`. Both of your reproductions are now tests in the suite, and both fail without the fix.

**On `ppfLoopCtrl` being an anti-pattern -- you are pointing at something real, but I have not taken the remedy, and I want to say why rather than leave it unanswered.**

The tempting reading is that the method is impure: it both computes state and acts on it, so extract a helper returning `(goalState, loopState)` and the acting moves to one call site. But that would not have prevented this bug. The asymmetry is not in the method, it is in the **commands**. Almost every command means *record this goal and move towards it*; exactly one means *record this goal and deliberately do not move towards it, because a second command is coming to do that*. The consume verb is split in two for a hard reason -- acting can run the consumer's processor, and this class promises never to call a processor with its monitor held -- so that one command must record without acting.

No return type absorbs that. A helper returning `(goalState, skip)` relocates the early return into a field, and the same wrong reasoning -- *"we are paused, so there is nothing to do"* -- fills that field in exactly the same way. Wrapping the acting step in `if (actNow) { ... }` has the identical hazard. Every one of those shapes would have let this regression through, because the mistake was never about purity. It was that **the one genuinely special command did not look special**, so a second early return could be added beside it and nothing read wrong.

So the difference is now carried where it cannot be skipped -- at every call site:

| | |
|---|---|
| `RECORD_TEARDOWN_AND_DOIT` | record the goal, act on it |
| `RECORD_PAUSE_AND_DOIT` | record the goal, act on it |
| `RECORD_CONSUME_AND_DONT_DOIT` | record only -- the one exception, named for it |
| `DOIT_NOW` | names no goal; acts on whatever stands |
| `LOOP_AROUND_WAKE` / `_CONTINUE` / `_RESTART` / `_IDLE` | the loop reporting on **itself** |

The `LOOP_AROUND_` prefix is the one that would have caught this. Reading `LOOP_AROUND_CONTINUE` at a branch that returns makes the question unavoidable: *this is the loop saying its iteration finished -- if I return here, what moves it off ACTIVE?* The bare name `CONTINUE` said nothing about who sent it or what it reported.

The enum is also declared in the order the method tests the commands, and the acting step is now an if-chain like the recording chain above it, so both halves read alike.

Full unit suite green on both branches, run with the cache disabled. PTAL.

caoccao (approved):
LGTM

PR #1754, opened 2026-09-13
EDG-997: Add a queueless InternalTopicFilterSubscriber variant

mschoenert-hivemq:
## Description (the why)

An internal subscriber always routed its matched messages through a client queue -- two SingleWriter hops and a persistence round-trip per message. For a consumer whose processor is cheaper than enqueueing, that is pure overhead. The motivating case is a data combiner that only remembers the last message for a topic: it pays for a durable buffer in order to perform an atomic store.

This adds `InternalTopicFilterSubscriberWithoutQueue`, a **separate class** rather than a mode of the queued one. Nearly everything the queued class does is about the queue: the poll/process/finish loop and its state machine, the declared queue limit and overflow strategy, and a processor that may finish later than it returns. A `withoutQueue()` flag would make all of that conditional and fork the state table. What remains without the queue is small -- register filters in the topic tree, and when a message matches, call the processor.

It is reached through `builderWithoutQueue(...)` on the **existing factory**, because that factory is also the registry resolving a client id back to its subscriber, and three places in Edge look subscribers up there.

Linear: [EDG-997](https://linear.app/hivemq/issue/EDG-997/add-a-queueless-internaltopicfiltersubscriber-variant)

**Stacked on [EDG-996](https://github.com/hivemq/hivemq-edge/pull/1752)** -- based on that branch, so this PR shows only the queueless commit. It retargets to master once EDG-996 merges.

## What the caller gives up, and must choose deliberately

With a queue and without a queue are two different things that happen to share a vocabulary. Every call site has to decide which it wants:

| | queued | queueless |
|---|---|---|
| buffering, backpressure, queue limit | yes | no |
| survives a restart | yes | no |
| sequential delivery | yes | **no** |
| future-returning processor | yes | no |
| per-filter context | yes | no (one subscriber per filter) |
| cost per message | 2 SingleWriter hops + persistence | a direct call |

The processor runs on the distributing thread, concurrently, so it **must not block AND must be thread-safe**. There is no future-returning form: a future would mean a message outliving its delivery, and there is nowhere to hold it.

- [x] A matched message reaches the processor on the distributing thread, with no queue and no SingleWriter involvement (asserted directly, not inferred)
- [x] The lifecycle collapses to two verbs, `attachConsume()` / `pauseDetach()`, since without a queue there is no buffer between collecting and processing
- [x] `pauseDetach()` stops delivery but **keeps the identity**; only `deallocate()` releases the client id, and it refuses a still-delivering subscriber
- [x] A throwing processor is reported as a failed delivery and does not poison the subscriber
- [x] Delivery takes no monitor, so a processor may call a lifecycle verb re-entrantly and two threads may deliver concurrently
- [x] The two kinds share one client-id namespace and cannot collide; an id freed by one becomes available to the other
- [x] Ingress-exclusion (generalized No Local) applies to both kinds
- [x] No new client-id prefix: `$INTERNAL::` is kept, so the connect-time reserved-id check needs no edit
- [x] 21 tests, each verified to bite by mutating the code it pins

## Notes for review

**No new prefix, and that is the point.** `ConnectHandler` rejects an external MQTT client connecting with a reserved id, and there it has only the string -- no subscriber object to ask. A prefix of this class's own would have to be added to that check, and forgetting it would let a client take an internal identity. Callers that do hold the object tell the kinds apart by which of the factory's two lookups answers.

**Two typed maps in the factory, no shared supertype.** An interface over the two classes would carry three methods, one of which exists only to ask which kind this really is -- a type conceding that its callers need the distinction. Of the four places that look a subscriber up, three can only mean the queued kind. `register`/`deregister` are synchronized because one client id spans both maps, so checking them and writing one has to be atomic; the lookups stay lock-free.

**One claim is deliberately not covered by a test**, and says so in the code: `attachConsume()` sets the delivery flag before registering filters, because the tree starts distributing the moment a filter is registered. The window needs a real concurrent distributor, so swapping the two statements leaves the suite green. The order is kept because it is free and the alternative is a silent drop.

## Non-Goals

- Migrating any consumer. The class is ready for one; the data combiner is the obvious first candidate, in its own change.
- Reconciling the overlap with MessageFabric, where a connection between handlers is queueless by default and a queue is an explicit inserted element. A discussion for another day.
- Any form of buffering, backpressure or restart survival for this variant. Those are what the queue is.
- Folding the two classes together later.


caoccao (changes requested):
**four reproducible issues** at `133daa7`. Two are new in #1754; two remain in its #1752 base.

1. **[P2] Stopped queueless subscribers can acquire orphan queues.** If the topic tree matches a subscriber before `stop()`, but distribution happens afterward, the registry lookup returns `null` and falls through to queueing. The message then has no consumer or cleanup path. The reproduction leaves one queued message after shutdown. Reject unregistered internal recipients before queueing. [[PublishDistributorImpl.java:254–270](https://github.com/hivemq/hivemq-edge/blob/133daa7c7d663992b111658d3f9ae46c1af003ac/hivemq-edge/src/main/java/com/hivemq/mqtt/services/PublishDistributorImpl.java#L254-L270)](https://github.com/hivemq/hivemq-edge/blob/133daa7c7d663992b111658d3f9ae46c1af003ac/hivemq-edge/src/main/java/com/hivemq/mqtt/services/PublishDistributorImpl.java#L254-L270)

2. **[P2] Concurrent queueless `stop()` calls can throw.** Both calls can pass the initial `deallocated` check; one completes shutdown before the other enters `pauseDetach()`, which then throws `IllegalStateException`. This violates the documented idempotence and can interrupt shutdown cleanup. Make the complete stop transition atomic. [[InternalTopicFilterSubscriberWithoutQueue.java:569–574](https://github.com/hivemq/hivemq-edge/blob/133daa7c7d663992b111658d3f9ae46c1af003ac/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriberWithoutQueue.java#L569-L574)](https://github.com/hivemq/hivemq-edge/blob/133daa7c7d663992b111658d3f9ae46c1af003ac/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriberWithoutQueue.java#L569-L574)

3. **[P2, inherited] A concurrent `consume()` can defeat `pause()`.** `consume()` releases the monitor before submitting `CONSUME`. A concurrent `pause()` can submit `PAUSE` first, leaving the loop active but `consuming=false`. Subsequent `pause()` calls then do nothing. The reproduction processes another message even after a later, nonconcurrent pause. [[InternalTopicFilterSubscriber.java:1589–1603](https://github.com/hivemq/hivemq-edge/blob/133daa7c7d663992b111658d3f9ae46c1af003ac/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1589-L1603)](https://github.com/hivemq/hivemq-edge/blob/133daa7c7d663992b111658d3f9ae46c1af003ac/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1589-L1603)

4. **[P2, inherited] Pending teardown permits subscription resurrection.** With an unfinished processor future, `stop()` returns while lifecycle methods remain enabled. Calling `start()` reattaches filters; completing the future then deregisters the subscriber without removing those filters or its newly registered callback. This leaves subscriptions pointing to a dead consumer. Set a safely published terminal-request guard before returning from `deallocate()`. [[InternalTopicFilterSubscriber.java:1666–1678](https://github.com/hivemq/hivemq-edge/blob/133daa7c7d663992b111658d3f9ae46c1af003ac/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1666-L1678)](https://github.com/hivemq/hivemq-edge/blob/133daa7c7d663992b111658d3f9ae46c1af003ac/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1666-L1678)

**Validation:** 114 existing targeted tests passed. Four added regression tests reproduced these failures. The previous failed-read regression now passes.

mschoenert-hivemq:
Thank you -- all four are fixed, and all four reproduced before they were fixed.

Two were in this PR:

1. **Orphan queues.** The distributor no longer falls through to queueing when a client id carries the internal prefix but no subscriber owns it. It returns `NOT_CONNECTED`, which is what the ordinary-client branch a few lines below already returns for a session that has gone. Worth noting this was not new here: the same fall-through predates the queueless work, so an internal subscriber missing from the registry has always ended up with a queue nobody drains. This PR made it visible by adding a second lookup beside the first.

2. **Concurrent `stop()`.** Both compositions, `start()` and `stop()`, are now synchronized like the verbs they compose. Safe here because none of these verbs runs consumer code -- they touch only the topic tree and the factory -- and `deliver()` deliberately takes no monitor, so a processor calling a verb waits rather than deadlocks. The queued class never had this defect: its `stop()` delegates to a single verb, so there is no gap between two calls.

The other two were in the base branch and are fixed in #1752, which this is stacked on.

Both branches are green.

caoccao (changes requested):
**Three issues remain**, reproduced on both current heads: #1754 `e0212c6` and #1752 `93cbb6e`.

1. **[P2] The consume/pause race is still unfixed.** Both `CONSUME_DONT_START` and `CONSUME` unconditionally set the goal to `ACTIVE`. If `PAUSE` runs between them, the second command overrides it—the conditional behavior described in the comments is missing. Processing resumes, and subsequent `pause()` calls do nothing because the lifecycle state already says paused. [[InternalTopicFilterSubscriber.java:1210–1219](https://github.com/hivemq/hivemq-edge/blob/93cbb6ed53f73a033a6f745aa031c0b536f202f7/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1210-L1219)](https://github.com/hivemq/hivemq-edge/blob/93cbb6ed53f73a033a6f745aa031c0b536f202f7/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1210-L1219)

2. **[P2] Concurrent queued `stop()` calls can still throw.** The queueless version is fixed, but queued `deallocate()` still checks the terminal state separately from its synchronized cleanup calls. Two callers can pass that check; after one completes teardown, the other enters `detach()` and throws `IllegalStateException`, potentially interrupting shutdown cleanup. [[InternalTopicFilterSubscriber.java:1761–1771](https://github.com/hivemq/hivemq-edge/blob/93cbb6ed53f73a033a6f745aa031c0b536f202f7/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1761-L1771)](https://github.com/hivemq/hivemq-edge/blob/93cbb6ed53f73a033a6f745aa031c0b536f202f7/hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriber.java#L1761-L1771)

3. **[P2] A publish can recreate a queued subscriber’s queue after teardown.** The new null check only covers subscribers stopped before lookup. When lookup succeeds, shutdown can clear the queue and deregister the subscriber before the later enqueue. The reproduction leaves an orphan message with no consumer or cleanup path. Ownership needs validation within the serialized enqueue operation. [[Registry check:266–280](https://github.com/hivemq/hivemq-edge/blob/e0212c62123f826055ea709a40538e6c81850b6c/hivemq-edge/src/main/java/com/hivemq/mqtt/services/PublishDistributorImpl.java#L266-L280)](https://github.com/hivemq/hivemq-edge/blob/e0212c62123f826055ea709a40538e6c81850b6c/hivemq-edge/src/main/java/com/hivemq/mqtt/services/PublishDistributorImpl.java#L266-L280), [[enqueue:297–298](https://github.com/hivemq/hivemq-edge/blob/e0212c62123f826055ea709a40538e6c81850b6c/hivemq-edge/src/main/java/com/hivemq/mqtt/services/PublishDistributorImpl.java#L297-L298)](https://github.com/hivemq/hivemq-edge/blob/e0212c62123f826055ea709a40538e6c81850b6c/hivemq-edge/src/main/java/com/hivemq/mqtt/services/PublishDistributorImpl.java#L297-L298).

The two previous queueless findings and restarting after pending teardown are fixed. **117 existing targeted tests passed on #1754; 90 passed on #1752.** The three failures above reproduced on both.

Separately, #1752’s [[dependency security scan](https://github.com/hivemq/hivemq-edge/actions/runs/34859081854/job/104026220713)](https://github.com/hivemq/hivemq-edge/actions/runs/34859081854/job/104026220713) is failing for reported Netty and Jackson vulnerabilities.

mschoenert-hivemq:
Thanks Sam. Replies to all three are on #1752, since the first two are fixes in the queued class that this branch inherits by being stacked on it. Briefly, for this PR:

**1.** The CONSUME / CONSUME_DONT_START split really was missing -- the comment described the fix and the code did not implement it. Sorry. Fixed in `f96f8ec3` on the 996 branch.

**2.** The concurrent-`stop()` fix you asked for here is already on this branch; the same defect on the queued class is fixed in `f4d1be74`.

**3.** Queue recreation after teardown is a pre-existing broker property rather than something these PRs introduce -- an ordinary MQTT client has the identical window when its session expires, and the resulting orphaned queue is never reclaimed because the orphan sweep only covers shared queues. Filed as EDG-1025 with a proposed general mechanism.

Rebased onto the updated 996 branch; green. PTAL.

hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriberFactory.java
```diff
@@ -68,27 +74,52 @@ public class InternalTopicFilterSubscriberFactory {
...
+    private void throwIfClientIdInUse(final @NotNull String clientId) {
+        if (registry.containsKey(clientId) || registryWithoutQueue.containsKey(clientId)) {
+            throw new IllegalStateException("InternalTopicFilterSubscriber '" + clientId
                     + "' is already in use; componentPrefix::instanceId must be unique within Edge"
                     + " (the existing subscriber must deallocate() before this identity can be reused)");
         }
     }
 
-    void deregister(final @NotNull InternalTopicFilterSubscriber subscriber) {
+    synchronized void deregister(final @NotNull InternalTopicFilterSubscriber subscriber) {
```
caoccao: This keyword `synchronized` doesn't make much sense.

hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriberFactory.java
```diff
@@ -68,27 +74,52 @@ public class InternalTopicFilterSubscriberFactory {
...
         registry.remove(subscriber.clientId(), subscriber);
     }
 
-    /**
-     * The allocated subscriber with this reserved-prefix clientId, or null if none is allocated. Used by
-     * the PublishDistributor to ask isExcludedIngressClientId(...) before queueing a publish.
-     */
+    synchronized void deregister(final @NotNull InternalTopicFilterSubscriberWithoutQueue subscriber) {
```
caoccao: This keyword `synchronized` doesn't make much sense.

caoccao (approved):
There are 2 small issues. Approved.

mschoenert-hivemq:
Fixed in `56bd32a0` on the EDG-996 branch, which this one is stacked on and has been rebased onto.

You were right, and it was a regression from the previous round: a progress report arriving while paused returned early, which skipped the act as well as the record -- and that report is the only thing that can mark the iteration finished, so the loop stayed wedged and later `consume()` / `stop()` calls were turned away. Both of your reproductions are now tests.

On the `ppfLoopCtrl` shape, the reasoning is on #1752 rather than repeated here. Short version: the commands are renamed so the one that records *without* acting is visibly different from the rest -- a pure helper returning a skip flag would have let the same mistake through in a different disguise.

Full suite green. PTAL.

hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriberFactory.java
mschoenert-hivemq: You are right -- removed in `a5429352`.

Each `deregister` is one atomic operation on a `ConcurrentHashMap`, so there is no check-then-act gap for the monitor to close. `register` keeps its own and needs it: it tests BOTH registries for a clashing client id and then writes to one, and that sequence must not interleave.

I did check the one case where the monitor might still have earned its keep -- a replacement claiming the same client id while the old subscriber tears down -- and it does not. The removal is `remove(key, value)`, the value-conditional form, so it removes the entry only while it still holds THIS subscriber; a replacement that registered first leaves a different object under that key and the removal correctly does nothing rather than evicting the live one. Both callers are also the last step of their own teardown and neither reads the registry afterwards, so nothing depended on the lock being held.

The asymmetry with `register` is now stated where the keyword used to be, so the absence reads as a decision rather than an omission.

hivemq-edge/src/main/java/com/hivemq/persistence/clientqueue/InternalTopicFilterSubscriberFactory.java
mschoenert-hivemq: Same here, and removed in the same commit -- the two `deregister` overloads are identical in this respect. Reasoning is on the sibling comment above rather than repeated.

mschoenert-hivemq:
Rebased onto master now that #1752 has merged, and your two `synchronized` remarks are fixed in `a5429352` (replies on the comments themselves).

**Why this is asking for your approval again, through no fault of yours.** #1752 merged as a single squashed commit, while this branch still carried that work as 16 separate commits -- the same changes with two different histories, which GitHub reported as a conflict. I rebased only this PR's own three commits onto master and dropped the rest. GitHub had also auto-retargeted the base from the #1752 branch to master when that branch merged, and that base change dismissed the approval you gave yesterday.

The rebase is content-preserving: comparing the rebased tree against what the branch held before, the only difference is EDG-1017 (the clean-up-job parallelism setting), which landed on master separately. Nothing of the queueless work changed.

So the PR is now four commits on master, six files, and mergeable. Full unit suite green on a fresh run. PTAL.

mschoenert-hivemq:
Index of your seven findings and where each was addressed, so nobody has to go hunting.

**Review of 2026-09-14**

1. Stopped queueless subscriber acquires an orphan queue -- fixed in `a4ca7e33`: the distributor refuses rather than falling through to queueing when the prefix matches and no subscriber is registered.
2. Concurrent queueless `stop()` throws -- fixed in `2c3cfc49`: `start()` and `stop()` synchronized, so the loser finds the work done instead of entering a verb that refuses a dead subscriber.
3. Concurrent `consume()` defeats `pause()` (inherited) -- see review 2, item 1 below.
4. Pending teardown permits resurrection (inherited) -- fixed in `7954381b`: four booleans became one six-valued state, recorded in `deallocate()` before the teardown is submitted.

**Review of 2026-09-15**

1. The consume/pause race still unfixed -- you were right; the first attempt was a comment describing a fix that was never written. Fixed in `f96f8ec3`: the acting half of the split records nothing at all, so a pause landing between the two halves is not overridden. Follow-up regression you found the same day fixed in `56bd32a0`.
2. Concurrent queued `stop()` throws -- fixed in `f4d1be74`: `deallocate()` synchronized.
3. Publish recreates a queue after teardown -- **deliberately not fixed here.** Pre-existing broker behaviour, not introduced by these PRs; an ordinary MQTT client has the same window at session expiry. Filed as EDG-1025 with a proposed mechanism.

**Approval of 2026-09-15** -- both `synchronized` remarks fixed in `a5429352`, replies on the comments themselves.

Items 3 and 4 of the first review and items 1 and 2 of the second landed on the #1752 branch and are now in master as part of the squashed `d4e3d272`.

caoccao (approved):

PR #1755, opened 2026-09-13
EDG-967: Add a Reload button that reconciles the workspace with the server

mschoenert-hivemq:
## Description (the why)

The workspace shows objects that no longer exist: a mapper still drawn after it was deleted, and an error when the user tries to delete it again. Reported in [EDG-967](https://linear.app/hivemq/issue/EDG-967/stop-caching-objects-from-the-edge-workspace).

**It is not the REST response cache.** That cache is in-memory only — no persistence plugin is installed — so it is already discarded on every page reload. The stale objects come from the **workspace graph store**, which is persisted to `localStorage` under `edge.workspace` with no version stamp and no expiry, and holds a whole server entity inside each node's `data`. Two behaviours make it drift and stay drifted:

- `onAddNodes` filters out any incoming node whose id is already present, so a node rebuilt from fresh data is discarded rather than replacing the stale one. Renames and config changes never reach the canvas.
- Nothing diffs the stored nodes against the server response. Removal happens only through the few imperative call sites that delete something, so an entity deleted by another user, by REST, by a config reload, or in another tab stays on the canvas permanently — and on disk, across reloads.

The full remedy is to derive the graph from the query data and persist only the user's layout. That touches 31 non-test files and is a multi-week rework. This PR is the contained alternative: it removes the reported symptom without the rework.

## Changes

**A prominent `Reload workspace` control**, beside the zoom/fit/lock/options group in the bottom-left of the canvas — labelled rather than an icon, because discoverability is half the bug. The `Reset Visualisation` control that exists today is two levels down behind a generic gear, and its on-screen text already describes this exact use case, but nobody finds it. Clicking opens a dialog with two choices:

| choice | what it does | what it costs |
|---|---|---|
| **Reload** | clears cached responses, refetches everything, reconciles the canvas | nothing — groups, layout and panel preferences are kept |
| **Reset everything** | clears all browser storage, reloads the app from scratch | groups, layout and preferences are lost, and the user is signed out |

**The reconcile** (`reconcile-utils.ts`) is a pure function — old graph plus freshly built graph in, new graph out — so the decisive logic is unit-testable without React, a browser or a mocked API. Server truth (`data`) comes from the rebuilt node; user truth (`position`, `parentId`, `expandParent`, selection) is carried over from the node already on the canvas. Both already live on the same node object in separate fields.

Three things the reconcile has to respect, each covered by a test:

- **One entity is several things on the canvas.** An adapter is an adapter node *plus* a device satellite *plus* a connector edge; a bridge is a bridge node *plus* a host satellite *plus* a connector. Rather than teach the reconcile every cluster shape, nodes go first and edges follow their endpoints — the same answer for every shape, including ones added later.
- **Groups are the user's, not the server's.** A `CLUSTER_NODE` has no server counterpart, so "the server did not return it" says nothing about whether it should exist. Only server-owned node types are ever removed.
- **Node ids are not uniform.** Adapters, bridges, listeners and devices carry a type prefix; a combiner node uses the bare entity id. The id shapes are collected in one place so the reconcile and its tests describe identity once.

**All-or-nothing fetching** (`useReloadWorkspace.ts`). Each of the six lists the graph is built from succeeds or fails independently, and the existing builder tracks only a combined `isLoading` with no combined error state. A reconcile on a partial result could not tell "the server has no combiners" from "the combiner call failed", and would delete every combiner from the user's workspace during a momentary blip. So the reload fetches all six, retries the whole set, and reconciles nothing unless every one resolved.

Each list carries its own fetcher rather than relying on `refetchQueries`, which only refreshes queries the cache already holds — a list nothing had mounted would be silently skipped and the reload would report success having asked the server nothing.

**Status polling is suspended** for the duration. A blocking overlay stops the *user*, not the *application*: three status queries poll on a timer and write into node `data`, mutating node objects in place, so a status write landing mid-reconcile would interleave with it. In-flight requests are cancelled at the start and polling is restored in a `finally`, so a failed reload cannot leave status frozen.

**An open configuration panel blocks the reload.** Every workspace detail panel is a child route of `/workspace/`, so the current path says both that a panel is open and which entity it is bound to. Dirty-form tracking barely exists in this module, so any open panel is treated as potentially unsaved rather than reloaded underneath.

### Fixed along the way

`onDeleteNode` removed an adapter and its device satellite but **never touched edges**, leaving orphaned connectors dangling between ids that no longer existed. Its spec asserted that behaviour; it now asserts the edges go too.

## Testing

- **Unit** — 1079 tests across the Workspace module, including 12 for the reconcile and 6 for the route-derived panel detection. Both decisive behaviours were verified to bite by mutating the source: removing edge pruning fails exactly the three dangling-edge tests; treating groups as server-owned fails exactly the group-protection test.
- **Component** — the full suite, 1500 tests, passing. The button's own spec covers the reported bug end to end: a ghost node disappears while a dragged position survives, an open panel blocks the reload, and a failing list leaves the workspace untouched.
- **End-to-end** — the full suite, 38 spec files. Two fail (`adapters/modbus`, `adapters/bacnetip`); both were confirmed to fail identically on an untouched checkout of `master`.
- **Types** — the error count is unchanged from `master` (41 before, 41 after), verified against a pristine worktree.

- [ ] A user who suspects the workspace is wrong can find the control without being told where it is
- [ ] Reloading removes entities the server no longer has, and refreshes those it changed
- [ ] Reloading keeps the user's node positions, groups and layout
- [ ] A failed or partial fetch changes nothing at all
- [ ] Removing a node takes its connector edges with it

## Non-Goals

- **The rework.** Deriving the graph from query data rather than storing a copy is the real fix and is not attempted here.
- **Groups in the reconcile.** A group's membership is stored twice — a child-id list on the group and a back-reference plus group-relative position on each child — so pruning a deleted member from a group needs care. Groups are never removed here, but a group holding a deleted member is not yet tidied.
- **Changing what any existing test asserts**, other than the `onDeleteNode` edge assertion described above, which encoded the defect.


aslamshah-hivemq (approved):
Looks good. 

There are some UX improvements which can be handled if needed in future.

aslamshah-hivemq:
The `cla-signed` verification failed, i have signed it now.

PR #1758, opened 2026-09-16
EDG-997: Make deallocate() require detached and paused

mschoenert-hivemq:
## Description (the why)

The two internal topic-filter subscribers disagreed about what releasing an identity means.

`InternalTopicFilterSubscriberWithoutQueue.deallocate()` refuses a subscriber that is still delivering. `InternalTopicFilterSubscriber.deallocate()` did the opposite: it detached and paused on the caller's behalf, and documented itself as having "no precondition".

Miguel raised the divergence reviewing the Lore page ([hivemq-edge-lore#16](https://github.com/hivemq/hivemq-edge-lore/pull/16)), noting that the reason the queueless class gives for refusing -- that deallocating would leave filters in the tree under an id the next owner may take -- argues against deallocating *without* detaching, not against detaching first. This settles it in favour of the refusing side: a terminal verb guessing that the caller also meant to stop the flow hides the caller's mistake instead of reporting it.

Linear: [EDG-997](https://linear.app/hivemq/issue/EDG-997/add-a-queueless-internaltopicfiltersubscriber-variant)

- [x] `deallocate()` throws when the subscriber is still attached or still consuming, naming both ways out
- [x] `deallocate()` still returns quietly on an already-dead subscriber
- [x] `stop()` orders `detach()`, `pause()`, `deallocate()` and remains safe to call twice, and from two threads at once
- [x] The two classes now state the same rule

## Non-Goals

- Any change to the queueless sibling, which already behaved this way
- Any change to the Lore page, which already describes this end state

## Notes

Two things followed from moving the ordering into `stop()` that the previous one-line version did not need:

- **It is now `synchronized`.** Composing three individually-synchronized verbs leaves gaps between them: two threads could interleave so that one completes the release while the other sits between `pause()` and `deallocate()`. Safe to hold the monitor across all three, since none of them runs consumer code -- unlike `start()`, which must NOT be synchronized for exactly that reason.
- **It tests for a dead subscriber first.** `detach()` and `pause()` refuse a dead subscriber rather than returning quietly, so composing them unconditionally would make a second `stop()` throw at the first verb. Only `deallocate()` returns quietly, which is what the previous version relied on.

Nothing in production called `deallocate()` directly; only `stop()` did. Two tests asserted the old behaviour -- one pinned the composing itself and is now its opposite, the other pauses first.

Both guards verified by mutation: removing the dead-subscriber test in `stop()` fails the concurrent-stop test from Sam's earlier review, which is the regression it exists to prevent. Full unit suite green on a fresh run with the cache disabled.


marregui (approved):
lgtm

PR #1759, opened 2026-09-16
EDG-1034 Houston: size OPC UA reads/browses to the server's operation limits, browse Variable children

marregui:
## Summary

Two Houston browse defects found by running the merged #1498 code against the PLCs in the Cologne hardware lab, both in `OpcUaNodeBrowser`, both pre-dating #1498:

1. **HTTP 500 on every browse of a WAGO PFC200** (Codesys OPC UA stack). Phase 2 reads 100 variables × 3 attributes = up to 300 `ReadValueId`s per `Read`; the server advertises `MaxNodesPerRead = 100` and enforces it on the `ReadValueId` count (verified: 99 ids OK, 102 → `Bad_TooManyOperations`). Any browse with more than 33 variables failed.
2. **Variable children were never browsed** (since #1444). `handleBrowseResult` marked a Variable visited when it recorded it, then `browseRecursive` refused to browse it. Struct members, array elements and properties are Variables under a Variable: on the lab S7-1500, 393 of the 671 stable PLC variables were missing (`"MB_CONFIG"."connect".InterfaceId`, `"MB_PTR"."Static_1"[0]` …); Prosys 1230 of 2116; WAGO 24 of 85.

Ticket: [EDG-1034](https://linear.app/hivemq/issue/EDG-1034), which has the device measurements behind both.

## What changed

**Phase 2 — attribute reads sized to the server.** The browser reads `MaxNodesPerRead`, `MaxNodesPerBrowse` and `MaxBrowseContinuationPoints` in one `Read` fired before Phase 1 (0 / missing / failed read → defaults). The attribute batch is `min(100, MaxNodesPerRead / 3)` variables — 33 on the WAGO, unchanged 100 on the S7-1500 (1000) and Prosys (10 000). A `Bad_TooManyOperations` fault halves the rejected slice and re-reads it from the same offset; only a rejected single-variable read fails the stream.

**Phase 1 — level-wise batched traversal instead of one `Browse` per node.** Fixing defect 2 alone would have tripled the number of serialised `Browse` round trips (≈50 s on the S7-1500 over the lab's 95 ms link, against a 120 s timeout), so Phase 1 is now breadth-first:

- every node of a level goes out in `Browse` requests of `min(100, MaxNodesPerBrowse)` nodes;
- all continuation points of a chunk are drained through one `BrowseNext` per page while the browse permit is held, before the next level is requested — the EDG-465 rule for the S7-1500's short-lived continuation points, kept;
- `Bad_TooManyOperations` on a chunk halves it and retries;
- a result answering `Bad_NoContinuationPoints` (the S7-1500 advertises **5** continuation points — a UInt16, unlike the two UInt32 operation limits) is re-browsed together with the other refused nodes of that chunk, in chunks of the advertised capacity, or half the previous chunk when none is advertised; a single node still refused fails the browse with its path;
- variables are collected in a NodeId-keyed map (first, i.e. shallowest, path wins — deterministic now, where the async recursion's arrival order was not) and `visited` only guards the traversal, so a Variable's children are browsed.

The semaphore contract of EDG-576 is unchanged: one permit per adapter, held for a chunk and its continuation pages. The wait for the permit is bounded by the browse deadline and interruptible (`tryAcquire(remaining, NANOSECONDS)`), so a browse queued behind another on the same adapter cannot outlive the 120 s timeout (review finding 1).

**Failure hygiene inside a chunk (review finding 2 and one more).** Every continuation point a `Browse`/`BrowseNext` response hands out is tracked before any status is judged; whatever fails afterwards — a `Bad_NodeIdUnknown` sibling, a bad continuation page, a service fault, a timeout, an interrupt — the cursors still open on the server are released with `BrowseNext(releaseContinuationPoints = true)` instead of being left for the PLC to time out (the S7-1500 has five per session). A response with fewer results than descriptions now fails the browse (`returned no result`) instead of being read as "no children".

**Hardening from the aggressive edge-case pass (third commit).** Found by the new `hivemq-edge-test` ITs against a Milo server that enforces S7-style limits: Milo hands out more continuation points per `Browse` than it accepts per `BrowseNext` (six vs. five), the six-point `BrowseNext` was refused with `Bad_TooManyOperations`, the six-point release was refused too, and the leaked cursors starved every later browse. `BrowseNext` and the release now go out in batches of `MaxBrowseContinuationPoints` (one per request when none is advertised) and a refused `BrowseNext` halves its batch. Also: the request-too-large family (`Bad_ResponseTooLarge`, `Bad_RequestTooLarge`, `Bad_EncodingLimitsExceeded`, `Bad_TcpMessageTooLarge`) is treated like `Bad_TooManyOperations` in both phases, since a chunk of 100 folders can outgrow a PLC's negotiated message size; and a node refused with `Bad_NoContinuationPoints` even when browsed alone (another client holding the session pool) is retried three times with a pause before the browse fails. `maxReferencesPerNode`, `maxDepth` (Variable levels count as before) and the non-Good-status failure messages are unchanged.

## Test plan

- [x] `OpcUaNodeBrowserTest`: 38 new cases (90 total) on a fake Browse/BrowseNext/Read server that enforces operation limits, pages results, caps continuation points and records every request's size and offset. **Every behaviour mutation-checked** — each of these sixteen mutants fails at least one test: no read halving; advertised read limit ignored; no offset rewind on retry; the original visited bug re-introduced; `MaxNodesPerBrowse` ignored; continuation pages not drained; rejected browse chunk not retried; `Bad_NoContinuationPoints` treated as fatal; advertised continuation-point capacity ignored; `acquireUninterruptibly()` instead of the deadline-bounded permit wait; no release of sibling cursors on failure; `BrowseNext` not batched to the capacity; refused `BrowseNext` not halved; release not batched; too-large faults not retried; single-node continuation refusal not retried.
- [x] `hivemq-edge-module-opcua` suite green on Temurin 25; `spotlessJavaCheck` clean.
- [x] `hivemq-edge-test` `DeviceTagBrowseIT`, `DeviceTagImportIT`, `DeviceTagEndToEndIT`, `OpcUaNodeBrowserConcurrencyIT` — 56 green on JDK 21, `--no-build-cache`.
- [x] New in hivemq/hivemq-edge-test#472: `OpcUaNodeBrowserServerLimitsIT` (11 — Codesys and S7 limits enforced by the Milo server, misadvertised limits, nested variables, dual paths, concurrency) and `OpcUaNodeBrowserNetworkIT` (3 — Toxiproxy lag, reset mid-browse with recovery, silent link). Whole `com.hivemq.edge.modules.opcua.*` IT package: 40 classes, 131 tests, 0 failures.
- [x] Hardware lab (Tailscale, ~90 ms RTT), rows before → after, against an independent asyncua walk of each address space (per-session diagnostics subtrees excluded):

  | Server | before | after | walk | browse time before → after |
  |---|---|---|---|---|
  | Siemens S7-1500 | 258 | 750 | 671 = 671, 0 path mismatches | 7.4 s → 11.3 s (Phase 1 6.4 → 5.5 s; the rest is the PLC's ~4 s per 300-attribute `Read`, now on 3× the rows) |
  | WAGO PFC200 G2 | HTTP 500 | 85 | 82 = 82 | — → 0.9 s |
  | CAS-QS-3110-F gateway | 267 | 470 | n/a | 15–17 s → 3.6 s |
  | Prosys Simulation Server | 886 | 1996 | 1385 = 1385 | ~350 ms → 150 ms |

  4 concurrent browses per adapter: all 200, byte-identical. Chaos on four Prosys adapters (75 s, 8 workers, adapters stopped/started ×3, MERGE_SAFE + OVERWRITE imports, delete + recreate): 2148 requests, 1731×200, 0 truncated documents; the single 500 is the pre-existing case of an adapter stopped during the first attribute read (`Bad_SessionClosed` before the first flush), same as master.
- [ ] Jenkins on the same-named composite branch.
- [x] `hivemq-testsuite` against the full distribution built from this branch (BACnet module spliced in as CI does; JDK 25 in the node containers): `hivemq-edge-functional-test` **594 tests, 1 failure, 62 skipped** — the failure is `BridgeHotReloadErrorsTest [2]`, an NPE in its `@BeforeEach` when the first auth-token fetch returned null (bridge hot-reload setup, no OPC UA involved; cases [1], [3]–[5] of the same parameterised test passed); all 60 Houston browse/import tests (`DeviceTagBrowseImportFunctionalTest`, `DeviceTagBrowseStressTest`, `DeviceTagDefaultNameCollisionTest`…) green. `hivemq-edge-non-functional-test` **47 tests, 0 failures, 3 skipped**, including the Toxiproxy browse-resilience tests. `DeviceTagBrowseImportHardwareTest` against the real S7-1500: **8 of 8 green** (1 skipped by design) — its `TEMP-RELAX` lower bounds on the 258-node baseline hold with the 750 rows now returned.


caoccao (changes requested):
**two confirmed regressions** in `881dba37`, both P2:

1. **The browse timeout no longer covers waiting for the semaphore.**  
   [[OpcUaNodeBrowser.java, lines 406–409](https://github.com/hivemq/hivemq-edge/blob/881dba3735e3a10f3471f8ea2121199539b7eb17/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/browse/OpcUaNodeBrowser.java#L406-L409)](https://github.com/hivemq/hivemq-edge/blob/881dba3735e3a10f3471f8ea2121199539b7eb17/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/browse/OpcUaNodeBrowser.java#L406-L409)  
   `acquireUninterruptibly()` now blocks the caller directly. Concurrent requests can exceed the 120-second deadline and ignore cancellation while waiting. Previously, the timed future covered this wait. Use interruptible `tryAcquire(remaining(deadline), NANOSECONDS)` and fail when it expires.

2. **A failed node in a batch leaks successful siblings’ continuation points.**  
   [[OpcUaNodeBrowser.java, lines 433–444](https://github.com/hivemq/hivemq-edge/blob/881dba3735e3a10f3471f8ea2121199539b7eb17/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/browse/OpcUaNodeBrowser.java#L433-L444)](https://github.com/hivemq/hivemq-edge/blob/881dba3735e3a10f3471f8ea2121199539b7eb17/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/browse/OpcUaNodeBrowser.java#L433-L444)  
   A bad result aborts before other results’ cursors are drained or released. I reproduced this with a paginated folder followed by a sibling returning `Bad_NodeIdUnknown`. The PLC retains those resources until it reclaims them. Collect and release abandoned cursors using `BrowseNext(true, …)` on failure.

**Validation:** All 77 existing browser tests pass. Both added regression tests pass on the base revision and fail on the PR head. I used an isolated harness because the full build encountered unrelated adapter-SDK compilation errors.

marregui:
Both confirmed, both fixed in `b6271cd8a` — thanks for reproducing them.

1. **Permit wait vs. deadline.** `browseChunk` now does `tryAcquire(remaining(deadline), NANOSECONDS)` and throws `TimeoutException` when it expires, so a browse queued behind another on the same adapter fails with the usual "Browse operation timed out after N seconds" instead of outliving it, and the wait is interruptible again. Two tests pin it (`browse_waitingForThePermit_countsAgainstTheDeadline`, `…_isInterruptible`); with `acquireUninterruptibly()` put back, both fail.
2. **Leaked sibling cursors.** Every continuation point in a `Browse`/`BrowseNext` response is now collected *before* any status is judged, and any failure inside the chunk — bad result, bad page, service fault, timeout, interrupt — sends `BrowseNext(releaseContinuationPoints = true, …)` for the ones still open (fire-and-forget; the browse is already failing). Two tests pin it: a paginated folder next to a `Bad_NodeIdUnknown` sibling (your case), and a bad continuation page while a sibling still has a page to go; with the release removed, both fail because the fake server still holds the cursor.

One more from re-reading the chunk code with your eyes: a response carrying fewer results than descriptions used to be treated as "no children" for the missing ones; it now fails with `Browse at path '…' returned no result` (test added).

Re-run on the fixed build: 82 browser unit tests (all 11 mutants caught), module suite 1298, `hivemq-edge-test` Houston ITs 56, lab hardware unchanged (S7-1500 750 rows = independent walk 671 stable, WAGO 85, CAS 470, Prosys 1996; 4-way concurrent browses byte-identical) — testsuite OPC UA/Houston functional classes, the non-functional module and `DeviceTagBrowseImportHardwareTest` are running now and I'll add the numbers to the test plan when they land.

caoccao (changes requested):
The original two reproductions are fixed. Reviewing heads `64193de83` and `7f2db77c`, I found **three remaining issues**:

1. **[P2] Late browse responses still leak continuation points — #1759.**  
   [[OpcUaNodeBrowser.java, lines 447–453](https://github.com/hivemq/hivemq-edge/blob/64193de83fbba684a828402c0a481b9c08a147b3/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/browse/OpcUaNodeBrowser.java#L447-L453)](https://github.com/hivemq/hivemq-edge/blob/64193de83fbba684a828402c0a481b9c08a147b3/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/browse/OpcUaNodeBrowser.java#L447-L453)  
   If the timed `get()` expires, execution exits before cursor tracking starts. The request remains in flight; a later response containing continuation points is abandoned. I reproduced one cursor remaining allocated after that response arrived. Attach cleanup to the in-flight future so late responses are handled after timeout or interruption.

2. **[P2] Rejected cursor-release batches are never retried — #1759.**  
   [[OpcUaNodeBrowser.java, lines 583–594](https://github.com/hivemq/hivemq-edge/blob/64193de83fbba684a828402c0a481b9c08a147b3/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/browse/OpcUaNodeBrowser.java#L583-L594)](https://github.com/hivemq/hivemq-edge/blob/64193de83fbba684a828402c0a481b9c08a147b3/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/browse/OpcUaNodeBrowser.java#L583-L594)  
   Normal `BrowseNext` requests halve their batch when rejected, but cleanup trusts the advertised capacity and only logs failures. With six open cursors, advertised capacity five, and enforced capacity two, cleanup leaves **five cursors allocated**. Retry rejected release batches at smaller sizes, or release individually.

3. **[P3] Test helpers assign duplicate NodeIds across separate branches — #472.**  
   [[TestNameSpace.java, lines 251–261](https://github.com/hivemq/hivemq-edge-test/blob/7f2db77c8df2bab6ec1086b71c6385b21f02bf50/src/test/java/com/hivemq/edge/modules/opcua/server/TestNameSpace.java#L251-L261)](https://github.com/hivemq/hivemq-edge-test/blob/7f2db77c8df2bab6ec1086b71c6385b21f02bf50/src/test/java/com/hivemq/edge/modules/opcua/server/TestNameSpace.java#L251-L261)  
   IDs use only the parent’s browse name. Creating `/DB1/Struct/Value` and `/DB2/Struct/Value` gives both children the identifier `Struct/Value`. One overwrites the other, and browsing loses one branch. I reproduced this against the embedded Milo server. Derive IDs from the parent’s full NodeId; `addFolder()` has the same issue.

**Validation:** The 90 existing unit tests and two extracted Milo paging/concurrency cases passed in an isolated harness. All three added reproductions failed. I did not rerun the full broker/Toxiproxy suite.

marregui:
All three confirmed and fixed in `efe1253a0` (branch has since been rebased onto master, so your line links point at old hashes) — thanks for the reproductions again.

1. **Late responses.** The timed wait no longer abandons the in-flight future: on `TimeoutException`/`InterruptedException` a callback is attached that releases whatever continuation points the response carries when it lands, for `Browse` and `BrowseNext` alike. Three tests pin it (`browse_browseResponseArrivesAfterTheTimeout_itsCursorsAreReleasedWhenItLands`, `…browseNextResponseArrivesAfterTheTimeout…`, `…interruptedWhileWaitingForABrowseResponse…`); the fake completes the deferred response after the browse has failed and asserts zero open cursors.
2. **Refused release batches.** A release rejected with `Bad_TooManyOperations` is split in half and both halves re-sent, recursively, down to single cursors; a single-cursor refusal or any other fault is logged and not retried. Your six-cursors / advertised 5 / enforced 2 case is the test (`browse_releaseRefusedAsTooManyOperations_isHalvedUntilItGoesThrough`), plus the two negative cases.
3. **`TestNameSpace` ids** (hivemq-edge-test `03104cec`). `addFolder`/`addVariable` now derive the child identifier from the parent's full NodeId identifier, so `/DB1/Struct/Value` and `/DB2/Struct/Value` are distinct nodes. `sameBrowseNamesInTwoBranches_bothBranchesBrowsedWithDistinctNodeIds` fails on the old helper.

One more found while re-reading with your lens: a `Read` response with fewer results than `ReadValueId`s used to misalign every attribute after the gap; it now fails the browse (`read_returnsFewerResultsThanAttributes_failsInsteadOfMisaligningAttributes`).

**Layout.** The 1 100-line `OpcUaNodeBrowser` was getting hard to review for exactly the cursor rules you were checking, so the same commit splits it into per-concern classes in `…opcua.browse`: `ChunkBrowser` (one chunk's I/O — request, pages, every cursor rule incl. release), `AddressSpaceWalker` (BFS, halving, `Bad_NoContinuationPoints` re-browse), `BatchAttributeSpliterator` (Phase 2 prefetch), `OperationLimits`, `Deadline`, `ServiceFaults`, `AttributeResolver`, `TagDefaults`, and the `DiscoveredVariable`/`PendingNode` records. `OpcUaNodeBrowser` is the ~180-line façade with its public constructors unchanged. Behaviour is equivalent — verified by deploying the old and new jar in turn against Prosys and diffing the browse output by node id (identical outside the per-session diagnostics subtrees). Each class has its own test class over a shared `FakeOpcUaServer`; the package sits at 100 % line and branch coverage (220 tests). `ff6a94dce` is a tidy on top. `OpcUaNodeBrowser.md` was rewritten to match.

Re-run on `ff6a94dce`: module suite 1 435, `hivemq-edge-test` Houston ITs 132, hardware over the lab link — S7-1500 750 rows, WAGO 85, CAS 470, each browsed twice, node-id sets identical to the pre-review baselines outside `/Server/ServerDiagnostics`.

caoccao (changes requested):
The previous three findings are fixed. I found **one remaining P2 issue in #1759**, and no new issues in #472.

**Continuation-point retries bypass request-size recovery.**  
[[AddressSpaceWalker.java, lines 161–165](https://github.com/hivemq/hivemq-edge/blob/ff6a94dcefde9fb81e27b1683ef078e2b71dafee/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/browse/AddressSpaceWalker.java#L161-L165)](https://github.com/hivemq/hivemq-edge/blob/ff6a94dcefde9fb81e27b1683ef078e2b71dafee/modules/hivemq-edge-module-opcua/src/main/java/com/hivemq/edge/adapters/opcua/browse/AddressSpaceWalker.java#L161-L165)

After `Bad_NoContinuationPoints`, `rebrowseExhausted()` calls `chunkBrowser.browse()` without the halving/retry logic used by the main traversal. If this retry returns `Bad_ResponseTooLarge` or `Bad_TooManyOperations`, the entire browse fails.

I reproduced this with four folders: the first request exhausts the cursor pool; retrying the remaining two folders exceeds the response-size limit because their references have longer names. Browsing those folders individually succeeds and returns all 16 variables.

Apply the same request-size backoff inside `rebrowseExhausted()`, preserving the rejected slice’s offset.

**Validation:** 225 focused checks passed, including the previous reproductions and two embedded Milo integration cases. The new combined-failure reproduction fails. The full broker/Toxiproxy suite was not rerun.

marregui:
Confirmed and fixed in `64741d566` — thanks for the four-folder reproduction.

The too-big handling only the level loop had is now one helper, `AddressSpaceWalker.browseOrShrink`, used by both loops: a re-browse of exhausted nodes refused with `Bad_TooManyOperations`/`Bad_ResponseTooLarge`/… halves its size and re-issues the same slice from the same offset, exactly like a level request. Your case is the test (`walk_retryOfExhaustedNodes_rejectedAsTooLarge_isHalvedFromTheSameOffset`: four paged folders, pool of two, the two-node retry refused as `Bad_ResponseTooLarge` → all 16 variables, call sequence asserted); it fails on the previous retry loop. Two more pin the edges: a single-node retry refused propagates the fault, any other fault on a retry propagates without halving.

Re-run: browse package 223, module suite 1 438, package still at 100 % lines/branches.

caoccao (approved):
LGTM

PR #1760, opened 2026-09-17
EDG-974: Reconnect PLC4X adapters and stop leaking on failed attempts

### LINKED ISSUE #1732: S7 Adapter does not retry when host is unreachable during startub
## Expected behavior

HiveMQ Edge should retry the connection continously (with exponential backoff) until the `maxPollingErrorsBeforeRemoval` is reached (or forever if set negative).
This already happens when
- hivemq edge is started, while the S7 server was up and running
- the S7 server then becomes unavailable

here edge does retry forever and once the server is back up, it automatically recovers and shows the protocol adapter as Running in the UI.

## Actual behavior

When starting HiveMQ Edge with a protocol adapter for S7 configured
but the S7 server is currently not running
then the adapter fails, goes to an error state, and does not automatically recover from it, when the S7 server becomes available later.
It requires manually restarting the S7 protocol adapter from the hivemq-edge UI, despite `<maxPollingErrorsBeforeRemoval>-1</maxPollingErrorsBeforeRemoval>` being configured and hivemq edge continously logging errors about the connection state.

S7 Server used https://github.com/AntDeM/S7-DB-Simulator/tree/main for local testing (the issue has also been observed on real Siemens S7 servers)

### Steps

**Bad case:**
- Stop S7 Simulator if running
- Start HiveMQ Edge with attached config
- Watch HiveMQ Edge logs, see expected logs below
- Start S7 Simulator
- Watch HiveMQ Edge continue logging errors, never recovering, adapter is shown as "Error" in UI
- Manually restart adapter from UI
- Watch Error logs stop, adapter is now shown as "Running" in UI

**Good case:**
- Start S7 Simulator
- Load config into Simulator
- Start HiveMQ Edge with attached config
- Stop S7 Simulator
- Watch HiveMQ Edge logs, see expected logs below
- Start S7 Simulator again
- Watch HiveMQ Edge reconnect, logs stop, adapter is shown as Running in UI

Error Logs
```java
2026-08-26 13:10:31,081 [plc4x-s7ha-thread-1] ERROR org.apache.plc4x.java.s7.readwrite.protocol.S7HPlcConnection#doPrimaryTcpConnections - doPrimaryTcpConnections: org.apache.plc4x.java.api.exceptions.PlcConnectionException: Error creating channel.
io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: /192.168.1.51:10102
  Suppressed: java.util.concurrent.CompletionException: Rethrowing promise failure cause
    at io.netty.util.concurrent.DefaultPromise.rethrowIfFailed(DefaultPromise.java:686)
    at io.netty.util.concurrent.DefaultPromise.sync(DefaultPromise.java:420)
    at io.netty.channel.DefaultChannelPromise.sync(DefaultChannelPromise.java:119)
    at io.netty.channel.DefaultChannelPromise.sync(DefaultChannelPromise.java:30)
    at org.apache.plc4x.java.spi.connection.NettyChannelFactory.createChannel(NettyChannelFactory.java:141)
    at org.apache.plc4x.java.s7.readwrite.protocol.S7HPlcConnection.doPrimaryTcpConnections(S7HPlcConnection.java:289)
    at org.apache.plc4x.java.s7.readwrite.protocol.S7HPlcConnection.run(S7HPlcConnection.java:388)
    at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
    at java.base/java.util.concurrent.FutureTask.runAndReset(Unknown Source)
    at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
    at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.net.ConnectException: Connection refused
  at java.base/sun.nio.ch.Net.pollConnect(Native Method)
  at java.base/sun.nio.ch.Net.pollConnectNow(Unknown Source)
  at java.base/sun.nio.ch.SocketChannelImpl.finishConnect(Unknown Source)
  at io.netty.channel.socket.nio.NioSocketChannel.doFinishConnect(NioSocketChannel.java:330)
  at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.finishConnect(AbstractNioChannel.java:384)
  at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.handle(AbstractNioChannel.java:432)
  at io.netty.channel.nio.NioIoHandler$DefaultNioRegistration.handle(NioIoHandler.java:388)
  at io.netty.channel.nio.NioIoHandler.processSelectedKey(NioIoHandler.java:596)
  at io.netty.channel.nio.NioIoHandler.processSelectedKeysPlain(NioIoHandler.java:541)
  at io.netty.channel.nio.NioIoHandler.processSelectedKeys(NioIoHandler.java:514)
  at io.netty.channel.nio.NioIoHandler.run(NioIoHandler.java:484)
  at io.netty.channel.SingleThreadIoEventLoop.runIo(SingleThreadIoEventLoop.java:225)
  at io.netty.channel.SingleThreadIoEventLoop.run(SingleThreadIoEventLoop.java:196)
  at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:1204)
  at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
  at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
  at java.base/java.lang.Thread.run(Unknown Source)
```

### Reproducer code

Configuration used (full config attached)
```xml
<protocol-adapter>
    <adapterId>simulatedS7</adapterId>
    <protocolId>s7</protocolId>
    <configVersion>1</configVersion>
    <config>
        
        <id>simulatedS7</id>
        <host>192.168.1.51</host>
        <port>10102</port>
        <controllerType>S7_300</controllerType>
        <keepAlive>false</keepAlive>

        <s7ToMqtt>
            <publishChangedDataOnly>true</publishChangedDataOnly>

            <pollingIntervalMillis>1000</pollingIntervalMillis>
            <maxPollingErrorsBeforeRemoval>-1</maxPollingErrorsBeforeRemoval>
        </s7ToMqtt>
    </config>
</protocol-adapter>
```

## Details

- Affected HiveMQ Edge version(s): 2026.12
- Used JVM version: 26.0.1

## Investigation so far

I have checked the code to find out why the two cases (host not reachable during startup and host becomes unreachable while edge is running) and found that in `com.hivemq.edge.adapters.plc4x.impl.AbstractPlc4xAdapter` the `start` method will 
- disconnect `tempConnection.disconnect();`
- set connection to null `this.connection = null;` 

in case of an exception (e.g. host not reachable)
```java
try {
   tempConnection.disconnect();
} catch (final Exception ex) {
   log.debug(
           "Tried disconnecting after connection error and caught exception",
           ex);
}
this.connection = null;
```

after that, any call to poll (from retry loop)
will fail immediately because
`if (tempConnection != null && tempConnection.isConnected()) {` is false
and this never recovers as no new `start` call is made.

If however the first `start` succeeds (S7 was running when edge starts) then the connection is no set to `null` on subsequent errors (e.g. S7 now unreachable). `if (tempConnection != null && tempConnection.isConnected()) {` still evaluates to false, because `tempConnection.isConnected()` returns false when the netty channel is not connected, but this does automatically recover once the host becomes reachable again and thus the retried poll methods eventually complete and the adapter is shown as "Running" again.

[hivemq-edge-s7.xml](https://github.com/user-attachments/files/31518762/hivemq-edge-s7.xml)

### LINKED ISSUE #1733: Edge leaks memory when S7 Adapter runs into timeouts and retries
## Expected behavior

When hivemq-edge retries an S7 connection for a protocol-adapter, this has no significant negative effect on the memory usage of edge itself.

## Actual behavior

Once the S7 Server 
HiveMQ Edge uses more and more memory, until it eventually gets killed by the OS (docker in this case)

## To Reproduce

This has been observed with a real Siemens S7 Server and HiveMQ Edge running on a VM in docker.
It has been reproduced locally with a simulated S7 Server from https://github.com/AntDeM/S7-DB-Simulator/tree/main and HiveMQ Edge running in Kubernetes.

### Steps

- Have an active S7 Server (or simulator) running
- Start HiveMQ Edge with a config that connects to the S7 server and pulls some addresses from it
- Stop the S7 Server
- Watch memory used by Edge increase over time as it tries to reconnect to the server

The memory increase is relatively slow (~170MB/h) so reproducing takes a while
See attached images for the memory usage over time

Image
Shows the "default" memory usage of ~1GB. This is stable over the whole week.

Image
Shows the increase by ~2GB in 12 hours (08/21 16:00 to 08/22 04:00) and then further increment until it gets OOM Killed.

### Reproducer code

Configuration used (full config attached)
```xml
<protocol-adapter>
    <adapterId>simulatedS7</adapterId>
    <protocolId>s7</protocolId>
    <configVersion>1</configVersion>
    <config>
        
        <id>simulatedS7</id>
        <host>192.168.1.51</host>
        <port>10102</port>
        <controllerType>S7_300</controllerType>
        <keepAlive>false</keepAlive>

        <s7ToMqtt>
            <publishChangedDataOnly>true</publishChangedDataOnly>

            <pollingIntervalMillis>1000</pollingIntervalMillis>
            <maxPollingErrorsBeforeRemoval>-1</maxPollingErrorsBeforeRemoval>
        </s7ToMqtt>
    </config>
</protocol-adapter>
```

## Details

- Affected HiveMQ Edge version(s): 2026.12
- Used JVM version: 26.0.1

[hivemq-edge-s7.xml](https://github.com/user-attachments/files/31519013/hivemq-edge-s7.xml)

mschoenert-hivemq:
## Description (the why)

Every weekend a customer stops their machines, which takes their S7 PLC offline. Edge then grows until the container out-of-memory kills it, restarts while the PLC is still down, and on Monday never reconnects -- recovery needs manual intervention.

Three faults chain together to produce that. Apache PLC4X 0.13.1 retries the TCP connection roughly once a second while the PLC is unreachable and permanently retains a Netty channel plus event loop group on every attempt, in a map owned by the connection object; measured at ~43 MB/h of live heap for one adapter with two tags. Separately, an adapter whose first connection attempt is refused is left holding no connection, and no code path ever tries to build one again. And a connection whose setup overruns our two-second timeout was simply dropped -- but the driver had already started its retry loop, which can only be cancelled through the object being discarded, leaving it unreachable and leaking until the JVM exits.

The decisive finding from the investigation: Edge *does* recover on its own when the PLC returns, provided it is still alive. The leak is what stops it being alive. Fixing the leak resolves the reported scenario.

Linear: [EDG-974](https://linear.app/hivemq/issue/EDG-974/s7-adapter-unbounded-memory-leak-while-reconnecting-and-no-recovery)
Resolves #1732
Resolves #1733

## Changes

Two changes in the shared PLC4X adapter base, so both the S7 and ADS adapters are covered. Ordering was a hard constraint: the second had to land with the first, since retrying on top of the old abandonment behaviour would mint a new unstoppable leaking loop per cycle.

- `Plc4xConnection#startConnection` keeps the in-flight future and closes whatever it eventually produces instead of walking away on timeout.
- `AbstractPlc4xAdapter` extracts the connect block out of `start()` into a `connect()` guarded by `compareAndSet`, and calls it from the polling failure branch after tearing the old connection down. Tearing down is what stops the driver's retry loop and releases what it retained, so each cycle returns to baseline rather than growing. Pacing is inherited from `PollingTask`'s existing exponential backoff -- no new thread, no new state machine. It removes the permanent null-connection state rather than adding one.
- The polling failure message now distinguishes "no connection was established" from "the connection was lost"; the single wording covered both cases and misdirects anyone reading a support log.

## Acceptance Criteria

- [x] A connection abandoned on timeout is always closed, so no driver retry loop outlives the attempt that started it.
- [x] An adapter that starts while the endpoint is refusing retries on its own and connects once the endpoint returns, with no manual intervention.
- [x] Memory retained by a failed attempt is released when the connection is torn down.
- [x] The polling failure message distinguishes a missing connection from a lost one.
- [ ] Live heap flat over a sustained outage, measured on a running node (see Non-Goals).

## Verification

`AbstractPlc4xAdapterReconnectTest` covers a poll with no connection ever established, a poll with a dead connection, and concurrent polls during an in-flight attempt. The tests were checked to actually bite rather than merely pass: removing the reconnect call fails two of them, and removing the connection teardown fails the leak-guard assertion on its own. Module suite green at 91 tests, module builds clean through errorprone and nullaway.

## Non-Goals

- End-to-end heap measurement on a running node against a stub S7 server. Verified at unit level here; the sustained-outage measurement is a separate exercise.
- Setting the driver's `retryTime` parameter to slow any loop that still escapes from 1 Hz to 0.125 Hz. Independent one-liner, deliberately not bundled.
- Migrating off PLC4X 0.13.1. Upstream deleted the offending classes in a full SPI rewrite headed for 1.0.0, which is a ground-up port for Edge with no release date -- a roadmap item, not this PR.


marregui (changes requested):
Read the diff cold against plc4x 0.13.1 sources and the polling framework before the description. The two mechanisms are right and the driver claims check out (the 1 Hz loop, the `eventLoops` map that never drains, `S7HPlcConnection.close()` cancelling the scheduler). Two things stop me approving: the "connects once the endpoint returns, no manual intervention" criterion does not hold under default config, and the abandoned-attempt fix has a hole that turns the old one-off leak into an unbounded one.

### 1. The adapter is still stopped after ~3.5 min of outage

Every poll that finds no connection calls `pollingOutput.fail(...)` (`AbstractPlc4xAdapter.java:144`). `PollingTask.handleExceptionDuringPolling` (`PollingTask.java:145-169`) stops rescheduling after `maxPollingErrorsBeforeRemoval` consecutive errors (plc4x default **10**, `Plc4xToMqttConfig.java:57`), and `AbstractSubscriptionSampler.onSamplerError` (`:86-91`) then calls `stopAsync(false)` → `ProtocolAdapterWrapper.stop` → STOPPED, with no restart path. Backoff for errors 1..10 sums to Σ2^k·100 ms ≈ 204 s.

- PLC powered on, port refusing: attempts fail in ms → 11 consecutive failures → adapter stopped after ~3.5 min. Deterministic.
- PLC unreachable (the customer's powered-off case): the driver's TCP connect timeout is **1000 ms** (`DefaultTcpTransportConfiguration.java:40`) and the default poll interval is 1000 ms. Whether the next poll lands on the `connecting` branch (success → `resetErrorStats`) or on the failure branch is a coin flip each cycle; once `errorCount ≥ 4` the backoff (1.6 s) exceeds the attempt duration and the run to 11 is deterministic. Expect STOPPED within minutes, sooner with any `pollingIntervalMillis > 1000`.

Pre-existing, not a regression, but it is the Monday symptom in the ticket and the PR ticks the AC. Either the reconnect path must not count as a polling error (publish-empty while an attempt is pending or can be started; fail only when neither), or the PR should state that recovery is bounded by `maxPollingErrorsBeforeRemoval` and the customer needs `-1`. Reproducer: point the adapter at a closed port and watch the runtime status for 4 minutes.

### 2. Regression: a connect that never returns now leaks once per poll cycle instead of once

`closeWhenComplete` (`Plc4xConnection.java:139`) only helps if `getConnection()` eventually returns. In the S7 driver it does not when the S7 handshake never completes: `S7HPlcConnection.connect()` blocks on `sessionSetupCompleteFuture.get()` with no timeout (`S7HPlcConnection.java:214`), the `closeFuture` listener that would complete it exceptionally is commented out (`:170-177`), and the COTP-stage `onTimeout` deliberately does nothing (`S7ProtocolLogic.java:186-190`; `Plc4xNettyWrapper` only deregisters the handler). Trigger: an endpoint that accepts TCP but never answers COTP — wrong rack/slot/TSAP on a 1200/1500, a forwarder in front of a dead PLC, or simply `nc -lk <port>`.

The 1 Hz scheduler is started *before* the blocking `get()` (`:193`), so each hung attempt owns a live `plc4x-s7ha` thread (daemon, MAX_PRIORITY), a Netty event loop group, a socket, and a blocked common-pool worker. ManagedBlocker compensation caps at 256 spares; past that, `connect()`'s own `runAsync` queues behind the hung workers and `connecting` sticks at true. Before this PR: one hung attempt, then dead-end. After: one per ~2-3 s until the cap.

Fix that closes the hole and makes AC 1 true in all cases: obtain the object *before* connecting — `plcDriverManager.getDriverForUrl(url).getConnection(url)` returns it without I/O (`DefaultPlcDriverManager.java:78-80`) — run `connect()` on the async thread, and on timeout `close()` the object and `cancel(true)` the future (`connect()` maps the interrupt to `PlcConnectionException`, `DefaultNettyPlcConnection.java:153`). Cheaper guard if that is out of scope: do not start a new attempt while a previous `pending` is still incomplete. ADS has the same exposure (`DefaultNettyPlcConnection.connect()` has the same untimed `get()`; its close listener is active, so only a silent endpoint hangs it).

### 3. `poll()` can tear down a connection that just succeeded

`poll()` reads `connection` and `isConnected()` first, `connecting` second (`AbstractPlc4xAdapter.java:114-128`); the async path sets `plcConnection`, then `connecting=false` (`:207-213`). Sequence: poll reads `temp` (in flight) → `isConnected()` false → async completes and clears `connecting` → poll reads `connecting` false → else-branch nulls and `disconnect()`s the good connection, starts another, reports "the connection was lost". Self-heals next cycle; costs one PLC session and an ERROR flicker. Fix: read `connecting` first — its false→true transitions happen only on the poll thread and in `start()`, so the reorder closes the window without a lock.

### Nits

- `connect()` resets `connecting` only for `Plc4xException` from `createConnection()`; a `RuntimeException` there, or `RejectedExecutionException` from `runAsync`, leaves `connecting=true` forever and the adapter silently publishes empty. Theoretical today (the constructor only throws `Plc4xException`), but a `try/finally`-shaped reset is one line.
- Test 1 exercises `createConnection()` throwing synchronously, which is the `failStart` path, not "PLC down at startup" (`startedSuccessfully` + async `startConnection()` failure nulling `connection`). That path has no test; a `HangingConnection` variant that throws immediately covers it in ~10 lines. Nothing tests the `PollingTask` interaction from point 1.
- Hand-rolled `awaitUntil`: `libs.awaitility` is in the catalog and used by the OPC UA and databases modules. `AtomicReference<Boolean>` → `AtomicBoolean`.

### Checked and fine

S7 `isConnected()` flips false on TCP loss (`S7HMuxImpl.channelUnregistered`), so the "lost" branch is reachable. A refused/unreachable connect throws before `scheduleAtFixedRate` (`S7HPlcConnection.java:186-193`), so fast-failing attempts do not spawn the loop; teardown on the first non-connected poll is the right lever. `connecting` is reset in `whenComplete` for unchecked failures and the next poll recovers the half-set `connection`. `stop()` vs an in-flight attempt is serialised by `Plc4xConnection.lock`.

Not counted, pre-existing: the small `stop()` window before the async runnable takes `lock`; `catch (Throwable)` swallowing interrupts; ADS teardown blocking up to ~2 s on the polling thread; the 2 s timeout value.

I have not run the suite or the reproducers; 1 and 2 are traced in source and each reproduces in minutes (closed port; `nc -lk`, or the lab S7-1500 with a wrong slot). Happy to run either if useful.

